PR #22, opened 2020-11-24
Fix addSource allowing custom credentials

VictorVelarde:
This fixes issues when dealing with sources for OAuth authenticated contents, where apiKey must be overriden by token (custom `credentials` in the payload).


VictorVelarde:
I have tested this linked locally and it fixes the issue on UserDatasets, so I'm moving on to release a beta.4 patch

PR #26, opened 2020-11-25
Fix CategoryWidgetUI loading and no data

VictorVelarde:
Issue reported by QA: "When I first access, I see the "loading skeleton" properly, but there is a blink and a "no data" message shows before showing the final content"

image


neokore (approved):
LGTM

PR #27, opened 2020-11-25
fix: keep the order when blocking categories in CategoryWidgetUI

VictorVelarde:
Keep the order of categories, even when 'blocking' some of them:

This is the issue:
image


neokore:
It affects to `order` behavior, behaves as `ranking` even if `fixed` is applied.

neokore:
I've created a different fix here: https://github.com/CartoDB/carto-react-lib/pull/32.
@VictorVelarde take a look and close this PR if the other works for you.

VictorVelarde:
Fix on 32 is working nicely

PR #32, opened 2020-11-26
Fix blocked categories not sorted as original data [ch121772]

neokore:
Related to point 5 in https://app.clubhouse.io/cartoteam/story/121772/qa-carto-for-react.

I propose this fix over the one in https://github.com/CartoDB/carto-react-lib/pull/27. @VictorVelarde take a look and tell me if it makes sense for you.


VictorVelarde (approved):
Works nicely!

PR #35, opened 2020-11-27
Add firebase config files for deployment

VictorVelarde:
Specific for carto project.


neokore:
Already done :bow:

VictorVelarde:
This was a previous test I did, so deleting

PR #37, opened 2020-12-01
Viewport filter working with antimeridian

alasarr:
Fix the viewport filter with the antimeridian.

The approach is to clip the bbox in two geometries when it crosses the antimeridian


neokore (approved):
LGTM

PR #44, opened 2020-12-09
Remove OAuthLogin from lib

VictorVelarde:
It is very coupled to final app, and quite small, so moving out to templates.

That way it will be also easier to customize


docs/api-reference/redux.md
```diff
@@ -55,6 +55,14 @@ Action to add a Layer to the store
 | id | <code>string</code> | unique id for the layer |
 | source | <code>string</code> | id of the source of the layer |
 
+### updateLayer
+Action to update a Layer in the store
+
+| Param | Type | Description |
+| --- | --- | --- |
+| id | <code>string</code> | unique id for the layer |
+| layerAttributes | <code>object</code> | layer attributes to update (source, or other custom attributes) |
```
neokore, line 64: Thanks!

neokore (approved):
LGTM

PR #45, opened 2020-12-09
feat: add layerAttributes to addLayer and refactor updateLayer

VictorVelarde:
This way they are both flexible to pass extra params to layers, and they both use the same destructuring approach


neokore (approved):
:+1:

PR #52, opened 2020-12-15
Fix addSource in cartoSlice spreading layerAttributes

juandjara:


src/redux/cartoSlice.js
```diff
@@ -130,7 +130,7 @@ export const removeSource = (sourceId) => ({ type: 'carto/removeSource', payload
  * @param {string} source - id of the source of the layer
  * @param {object} layerAttributes - custom attributes to pass to the layer 
  */
-export const addLayer = ({id, source, layerAttributes = {}}) => ({ type: 'carto/addLayer', payload: {id, source, layerAttributes} });
```
VictorVelarde, line 133: I have some small doubts on this.

First, about "what it should be", then about current implementation and clients.

1. What it should be ™️ 

The original idea was to have flexibility to pass 'a bunch of extra params' to the layer, both for `addLayer` and `updateLayer`. We wanted that, but without 'mixing' at the same level the mandatory params (id and source), required for redux, with these 'custom params' (more related to the layer itself and its internal properties))... and my question to you @juandjara is: does it feel right?

Once we agree on that, we can talk about 'the fixes', to allow several usecases, like the 'change layer visibility'.

2. About current implementation

If we apply this change here, we will need to test the current uses of `layerAttributes` at the template project:
https://github.com/CartoDB/carto-react-template/search?q=layerAttributes

and it looks like:
- UserDataset.js will need a small change here: https://github.com/CartoDB/carto-react-template/blob/bbe40a3362fa71fe02a54dae4b232213a1d10713/template-sample-app/template/src/components/views/datasets/UserDatasets.js#L113
- StoreDetail, using 'selectedStore' would still work nicely (as it doesn't use the current addLayer, just updateLayer)

and this one will be still working nicely: https://github.com/CartoDB/carto-react-template/search?q=selectedStore

VictorVelarde: 🔝 @neokore for awareness on this, as we already talked for a while on `layerAttributes`

juandjara: Well, I've proposed this change but I could have also proposed to remove the destructuring from `updateLayer`, as that would also work and solve the issue.  I did it like in order to try to avoid writting `layerAttributes` in my layer files 😛 , because I find it too verbose. I think if you really want to separate the initial `id` and `source` from this extra data I would call this data something like `state` or `layerState`, because "state" is the generic term used by redux docs when referring to arbitrary pieces of data. Anyway, that naming is not up to me.

VictorVelarde: I'm ok with your current approach, just let me prepare the other PR to merge them as a block

VictorVelarde (approved):

src/redux/cartoSlice.js
VictorVelarde: Done, both are merged now, thx for the fix @juandjara !! 🚀

PR #67, opened 2021-01-27
Add Husky and lint-staged

AdriSolid:
Story details: https://app.clubhouse.io/cartoteam/story/127479

It needed to integrate the default CRA ESLint config. ESLint aligned on lib and template.


.editorconfig
```diff
@@ -0,0 +1,22 @@
...
+indent_style = space
+insert_final_newline = true
+max_line_length = 80
+trim_trailing_whitespace = true
+
+[*.md]
+max_line_length = 0
+trim_trailing_whitespace = false
+
+[*.ejs*]
```
neokore, line 18: I think that this is not needed

AdriSolid: We should remove it from template repo also, as it has same editorconfig file

.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "printWidth": 90,
+  "arrowParens": "always",
+  "semi": true,
+  "tabWidth": 2,
+  "singleQuote": true,
+  "jsxSingleQuote": true
```
neokore, line 7: Lots of :heart: here!
Just a question, can we add `trailingCommas: none`? What are your feelings about trailing commas?

.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "printWidth": 90,
```
neokore, line 2: For me maybe 90 chars is a bit tight, but if it is our standard I'm ok with this.

CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # CHANGELOG
 
 ## Not released
+
+- Client Side Widgets [#62](https://github.com/CartoDB/carto-react-lib/pull/62)
```
neokore, line 5: Why is this line added in this PR? I don't see anything related to CSW here.

.prettierrc
AdriSolid: I prefer to set `trailingCommas` to `none` as well (more like a personal preference)

package.json
```diff
@@ -41,13 +42,29 @@
     "@turf/bbox-polygon": "^6.0.1",
     "@turf/boolean-contains": "^6.3.0",
     "@turf/boolean-intersects": "^6.0.2",
+    "@typescript-eslint/eslint-plugin": "^4.14.1",
+    "@typescript-eslint/parser": "^4.14.1",
```
neokore, line 46: I would like to remove this `eslint-config-react-app` dependencies, but @AdriSolid already tried and it fails in case of not adding them.
What about the plans of moving to typescript :stuck_out_tongue: ?

CHANGELOG.md
AdriSolid: It was missed in last PR.

package.json
```diff
@@ -58,5 +75,20 @@
...
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.+(js|jsx)": [
+      "yarn lint:fix",
+      "git add"
```
neokore, line 87: Just to confirm, we want this to be automatic, right? In other projects I usually get notified, but it doesn't fix and add changes automatically; but I understand that this "more comfortable" for devs.

package.json
AdriSolid: The reason is mainly to have the ESLint config aligned with the template. So `eslint-config-react-app` needs TS packages..
I agree moving to TS! :laughing:

package.json
AdriSolid: Indeed, we can open a discussion for that, but, it's using same config as template.

src/widgets/HistogramWidget.js
```diff
@@ -8,21 +8,21 @@ import { getHistogram } from './models';
...
+ * @param  props
+ * @param  {string} props.id - ID for the widget instance.
+ * @param  {string} props.title - Title to show in the widget header.
+ * @param  {string} props.dataSource - ID of the data source to get the data from.
+ * @param  {string} props.column - Name of the data source's column to get the data from.
+ * @param  {string} props.operation - Operation to apply to the operationColumn. Must be one of those defined in `AggregationTypes` object.
+ * @param  {number[]} props.ticks - Array of thresholds for the X axis.
+ * @param  {formatterCallback} [props.xAxisformatter] - Function to format X axis values.
+ * @param  {formatterCallback} [props.formatter] - Function to format Y axis values.
+ * @param  {boolean} [props.viewportFilter=false] - Defines whether filter by the viewport or not.
```
neokore, line 21: I'm not sure this improves the definition for this parameter. I suggest to keep "globally" or explain a bit the other option.

AdriSolid: 'globally' LGTM

CHANGELOG.md
neokore: I rather having a separate PR for this kind of small mistakes that can be merged quickly to keep everything clear and tidy, and avoid losing this item in changelog in case of this PR is not merged before releasing a new version. But if we are sure this PR is going to be merged soon before releasing, it can stay for this time.

AdriSolid: Not sure if it will be merged soon.. So let's remove it for the moment, and will create a small PR for that :+1:

package.json
neokore: Ok, having this config synced with the one on the template is a priority.

package.json
neokore: Agree, same as above.

neokore (approved):
LGTM, fix the yarn.lock conflict and we are ready to go

PR #72, opened 2021-01-29
Tests for widgets/models

AdriSolid:
Add tests for `widgets/models` (with no filters applied).

### DIFF
--- a/src/tests/widgets/models/FormulaModel.test.js
+++ b/src/tests/widgets/models/FormulaModel.test.js
@@ -1,0 +1,103 @@
+import { minify } from 'pgsql-minify';
+
+import {
+  getFormula,
+  buildSqlQueryToGetFormula,
+  filterViewportFeaturesToGetFormula
+} from 'src/widgets/models/FormulaModel';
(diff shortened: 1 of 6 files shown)


src/tests/widgets/models/FormulaModel.test.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    test(AggregationTypes.COUNT, () => {
+      const args = createArguments(AggregationTypes.COUNT);
+      expect(filterViewportFeaturesToGetFormula(args)).toEqual([{ value: 3 }]);
+    });
+
+    test(AggregationTypes.AVG, () => {
+      const args = createArguments(AggregationTypes.AVG);
+      expect(filterViewportFeaturesToGetFormula(args)).toEqual([{ value: 2 }]);
```
VictorVelarde, line 86: as we said, we should probably rename this function in another PR, to something like getFormulaFromViewportFeatures or calculateFormulaXXXX

VictorVelarde (approved):
LGTM 🚀

PR #75, opened 2021-02-01
fix: widgets loading state when calculating client-side

AdriSolid:
#### Background
Widget loaders were not working in client-side-mode. Widget loaders should be displayed every time viewport changes, ending with the end of the computation.

#### Change List
- Add a new store state with the widget ids.
-  Add `useLoadingStateFromStore` hook and use it in each `/widgets` component.


src/redux/cartoSlice.js
```diff
@@ -276,3 +294,20 @@ export const removeViewportFeatures = (data) => ({
   type: 'carto/removeViewportFeatures',
   payload: data
 });
+
+/**
+ * Action to set the widget loader state
+ * @param {object} widgetId - the id of the widget
```
VictorVelarde, line 300: types should be string & boolean right?

src/widgets/CategoryWidget.js
```diff
@@ -25,47 +26,44 @@ function CategoryWidget(props) {
...
 
   useEffect(() => {
     const abortController = new AbortController();
-    if (
-      data &&
-      credentials &&
-      (!props.viewportFilter ||
-        (props.viewportFilter && viewportFeatures[props.dataSource]))
-    ) {
+    if (data && credentials && widgetIdIsSet) {
```
VictorVelarde, line 41: isLoadingStateSet ? (using widgetIdIsSet feels still not clear to me)

src/widgets/CategoryWidget.js
```diff
@@ -94,12 +92,16 @@ function CategoryWidget(props) {
...
+      title={props.title}
+      loading={widgetLoaders[props.id]}
+      {...props.wrapperProps}
+    >
       <CategoryWidgetUI
         data={categoryData}
         formatter={props.formatter}
         labels={props.labels}
-        loading={loading}
+        loading={widgetLoaders[props.id]}
```
VictorVelarde, line 104: I would use `isLoading` for the prop name

AdriSolid: Indeed, seems better

src/redux/cartoSlice.js
```diff
@@ -130,6 +131,23 @@ export const createCartoSlice = (initialState) => {
         if (state.viewportFeatures[sourceId]) {
           delete state.viewportFeatures[sourceId];
         }
+      },
+      setWidgetLoaders: (state, action) => {
+        const { widgetId, isLoading, displayAllLoaders } = action.payload;
+
+        if (displayAllLoaders) {
+          state.widgetLoaders = Object.fromEntries(
+            Object.keys(state.widgetLoaders).map((key) => [key, true])
```
VictorVelarde, line 140: key --> widgetId, right?

AdriSolid: Indeed! Are we overusing `widgetId` here?

src/widgets/useLoadingStateFromStore.js
```diff
@@ -0,0 +1,43 @@
...
+
+  useEffect(() => {
+    if (!widgetId) {
+      throw new Error('Widget "id" property is required.');
+    }
+
+    setLoading(filterIsByViewport);
+
+    return () => {
+      removeWidgetLoaders();
```
VictorVelarde, line 32: After reading this again, I guess we should remove just loading state for the current widget. If we want to keep the simmetry with the 'add' option, which allows 'all' and also one specific widgetId, we should probably allow passing a widgetId list to removeWidgetLoaders

AdriSolid: Let's modify our action to allow removing specific `widgetId`s, it's true that a single widget component could be un-mounted.

VictorVelarde (approved):
👍🏻

VictorVelarde:
@AdriSolid I made some refactors here: 
- 'loaders' to 'loading' (everywhere)
- actions modification: `setWidgetLoadingState` / `removeWidgetLoadingState` (notice the singular) + a new `setAllWidgetsLoadingState`

Please have a look at the changes, test them to see if I broke anything and, if you are ok with it, go ahead and merge it

AdriSolid:
@VictorVelarde nice! Will need to check in a near-future if can pass to `removeWidgetLoadingState` a `widgetId`s list.

PR #76, opened 2021-02-01
fix: min/max aggregated functions

AdriSolid:
min/max aggregated functions were not correctly built for features calculations.

### DIFF
--- a/src/widgets/operations/aggregation/values.js
+++ b/src/widgets/operations/aggregation/values.js
@@ -1,15 +1,28 @@
 import { AggregationTypes } from '../../AggregationTypes';
 
+const isFeature = (val) => typeof val === 'object';
+
 const sum = (v, column) =>
-  v.reduce(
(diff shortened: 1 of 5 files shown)


src/widgets/operations/aggregation/values.js
```diff
@@ -1,15 +1,27 @@
 import { AggregationTypes } from '../../AggregationTypes';
 
-const sum = (v, column) =>
-  v.reduce(
-    (a, b) => (typeof b === 'object' ? a + (b.properties[column] || 0) : a + b),
-    0
+const isFeature = (value) => typeof value === 'object' && 'properties' in value;
+const isFeatureOrValue = (value, column) =>
```
VictorVelarde, line 4: The previous is nice (`isFeature`) but the result of `isFeatureOrValue` is not boolean and not clear for the caller. After a call, we decided to leave outside of this function the feature management. Conversion from feature to proper value should be done outside

VictorVelarde (approved):
👍🏻

PR #77, opened 2021-02-01
fix: eventual memory leaks on requestAnimationFrame for Category & Formula widgets

VictorVelarde:
Adding a reference, so animation can be properly cleaned up when unmounting the Category and Formula widgets (with cancelAnimationFrame)

It also includes a simple refactor on animations, with more explicit params and a single module for both functions


AdriSolid (approved):
Awesome, UIWidget component looks cleaner also like this!

PR #79, opened 2021-02-02
test: groupby & histogram operations tests

AdriSolid:
This PR includes new tests for `/operations`:
- groupby
- histogram

Also includes a little fix in `operations/histogram` to check if processed value from `aggregatedFunctions` is not finite, so we return `0`. Mainly for `HistogramWidget`'s tooltip.


VictorVelarde:
Tests are nice! I have just done a bit of refactor, mostly on naming

VictorVelarde (approved):
👍🏻

PR #82, opened 2021-02-04
chore: fix linter warnings

VictorVelarde:
Fixing linter warnings.

Please @AdriSolid check specially the changes on userViewportFeatures. Also there is still 1 warning in the Lin 75. useCallback. Please have a look at it locally.

### DIFF
--- a/src/api/useViewportFeatures.js
+++ b/src/api/useViewportFeatures.js
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useCallback, useState } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import { setViewportFeatures, setAllWidgetsLoadingState } from '../redux/cartoSlice';
 import bboxPolygon from '@turf/bbox-polygon';
@@ -36,7 +36,7 @@ function getUniqueFeatures(tiles, uniqueId) {
 
   tiles.forEach(({ data, isCompletelyInViewport }) => {
     if (data) {
-      for (const feature of data) {
+      for (let feature of data) {
         const id = pickPropByUniqueId(feature, uniqueId);
 
         if (!features.has(id)) {
@@ -96,7 +96,7 @@ export default function useViewportFeatures(source, uniqueId, debounceTimeOut = 
       dispatch(setAllWidgetsLoadingState(true));
       computeFeatures(uniqueFeatures, viewport, source.id);
     }
-  }, [uniqueFeatures, viewport, source]);
+  }, [computeFeatures, dispatch, source, uniqueFeatures, viewport, widgetsLoadingState]);
 
   const onViewportLoad = useCallback(
     (visibleTiles) => {
--- a/src/api/useViewportFeatures.js
+++ b/src/api/useViewportFeatures.js
@@ -1,4 +1,4 @@
-import { useEffect, useCallback, useState } from 'react';
+import { useEffect, useCallback, useState, useRef } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import { setViewportFeatures, setAllWidgetsLoadingState } from '../redux/cartoSlice';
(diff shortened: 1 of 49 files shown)


src/api/useViewportFeatures.js
```diff
@@ -71,8 +71,9 @@ export default function useViewportFeatures(source, uniqueId, debounceTimeOut =
   const viewport = useSelector((state) => state.carto.viewport);
   const [uniqueFeatures, setUniqueFeatures] = useState();
   const widgetsLoadingState = useSelector((state) => state.carto.widgetsLoadingState);
+  const referencedWidgetsLoadingState = useRef(widgetsLoadingState);
```
VictorVelarde, line 74: Why using a ref here for this info from store?

AdriSolid: Because adding `widgetsLoadingState` in the dependency array [here](https://github.com/CartoDB/carto-react-lib/blob/27d5215aec0fcc0f86642b41fc1fea83cb3bf187/src/api/useViewportFeatures.js#L99) cause unnecessary/loop re-renders due to dispatching `setAllWidgetsLoadingState`. So, if state this store state is updated, `useEffect` will be called.
Wrapping things with `useRef` allows to persist the value for the entire lifetime of the component.

src/api/useViewportFeatures.js
```diff
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useCallback, useState, useRef } from 'react';
```
AdriSolid, line 1: I usually prefer sorting hooks alphabetically, it's only code style.

src/ui/widgets/CategoryWidgetUI.js
```diff
@@ -339,9 +339,9 @@ function CategoryWidgetUI(props) {
       drawFrame: (val) => setAnimValues(val),
       requestRef
     });
-
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
AdriSolid, line 342: In order to avoid the 'delayed animation' we were talking:
`prevAnimValues` in the dependencies array causes the delay, I recommend to move the animation logic to an `useCallback`, something like:
```javascript
const launchAnimation = useCallback(() => {
   ...
}, [data, prevValue, setValue]);

useEffect(() => {
  launchAnimation();
  // eslint-disable-next-line react-hooks/exhaustive-deps
  return () => cancelAnimationFrame(requestRef.current);
}, [data, setValue, launchAnimation]);
```
We still need to check that disabled line!

src/ui/widgets/FormulaWidgetUI.js
```diff
@@ -62,11 +62,9 @@ function FormulaWidgetUI(props) {
     } else {
       setValue(data);
     }
-
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     return () => cancelAnimationFrame(requestRef.current);
-
-    // eslint-disable-next-line
-  }, [data, setValue]);
+  }, [data, prevValue, setValue]);
```
AdriSolid, line 67: In order to avoid the 'delayed animation' we were talking:
Passing `prevValue` to the dependencies array causes the delay, we could use same technique as my last comment, using `useRef`
```javascript
const referencedPrevValue = useRef(prevValue);
```

VictorVelarde:
I'm gonna split out the ci & test part, while we decide and test the other block

src/ui/widgets/FormulaWidgetUI.js
AdriSolid: Fixed [here](https://github.com/CartoDB/carto-react-lib/commit/75b80e5eb885bc85b650cbae12354641c6e5636f)

VictorVelarde:
Thx for the fix

VictorVelarde:
Please approve it, and I'll merge it

AdriSolid (approved):
:D

PR #87, opened 2021-02-08
test: add api tests

AdriSolid:
Add `/api` tests. Missing `useViewportFeatures` hook, waiting for next improvements.

For: https://app.clubhouse.io/cartoteam/story/133201/test-suites-at-lib-project-ii


src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+  },
+  {
+    column1: value
+  }
+];
+
+describe('Filters', () => {
+  test('should return 1 if no filters present', () => {
+    const params = { filters: {}, type: 'number' };
+    expect(applyFilter(params)()).toBe(1);
```
VictorVelarde, line 23: @AdriSolid shouldn't we use a feature for the test?

On the other hand, I find a bit strange the fact that applying a filter to an undefined feature return 'true'.... I think we should probably consider throwing an error in that case.

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  test('should return true if no filters present', () => {
+    const params = { filters: {}, type: 'boolean' };
+    expect(applyFilter(params)()).toBe(true);
```
VictorVelarde, line 28: same as previous.

src/tests/api/Filter.test.js
AdriSolid: - Was thinking that is not necessary passing a feature here, as long as if `filters` are equal to `{}`, the function firstly returns `1` or `true`: [reference](https://github.com/CartoDB/carto-react-lib/blob/a80ec3efca842b5c7bc683cc763277d316e86d27/src/api/Filter.js#L41)
- If we pass an undefined feature the result should be `0` or `false`, [because the feature doesn't pass the filter](https://github.com/CartoDB/carto-react-lib/blob/a80ec3efca842b5c7bc683cc763277d316e86d27/src/api/Filter.js#L23).

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
+import { applyFilter } from 'src/api';
+import { filters } from './data-mocks/filters';
+
+const makePossibleFeature = (value) => [
```
VictorVelarde, line 4: A bit strange name... 😄  ? how about makeFeatureWithValueInColum? 

Also I find a bit strange to use 'column1' as the default column... is there a reason for this?
If there is no special requirement, we could add a new parameter, and modify a bit the function like:

```
const makeFeatureWithValueInColum = (value, column='column1')
```

What do you think?

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+  test('should return true if no filters present', () => {
+    const params = { filters: {}, type: 'boolean' };
+    expect(applyFilter(params)()).toBe(true);
+  });
+
+  describe('feature passes filter - boolean type', () => {
+    const params = { filters, type: 'boolean' };
+
+    describe('should return false if feature column value is falsy', () => {
+      const testCases = [0, null, undefined, false, ''];
```
VictorVelarde, line 35: if you apply my previous suggestions, this would be 'columnValues' instead of testCases..

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+    expect(applyFilter(params)()).toBe(1);
+  });
+
+  test('should return true if no filters present', () => {
+    const params = { filters: {}, type: 'boolean' };
+    expect(applyFilter(params)()).toBe(true);
+  });
+
+  describe('feature passes filter - boolean type', () => {
+    const params = { filters, type: 'boolean' };
```
VictorVelarde, line 32: I find it a bit difficult to think about the test results, having filters imported from an external file, instead of declaring them inline. If feature building is outside the test and also the filters... it's a bit too abstract to me, as I have to navigate between 3 files to understand what's happening...

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+      type: 'Point',
+      coordinates: [0, 0]
+    },
+    properties: {
+      column1: value
+    }
+  },
+  {
+    column1: value
+  }
```
VictorVelarde, line 17: This 'rawFeature' concept here looks a bit obscure to me, could you please explain what do we want here? when is that kind of 'object' (not really a feature) useful?

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+    const params = { filters, type: 'boolean' };
+
+    describe('should return false if feature column value is falsy', () => {
+      const testCases = [0, null, undefined, false, ''];
+      for (const tc of testCases) {
+        test(String(tc), () => {
+          const [featureWithProperties, rawFeature] = makePossibleFeature(tc);
+          const withProps = applyFilter(params)(featureWithProperties);
+          expect(withProps).toBe(false);
+          const noProps = applyFilter(params)(rawFeature);
```
VictorVelarde, line 41: I try not to mix several **expect** in the same test, as they tend to be more complex... In this case, I have no clear idea about the 'rawFeature'... but I would probably suggest extracting that to another test, even if that requires a bit of duplication

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+      const paramsWithFilterFunctionNotImplemented = {
+        filters: {
+          ...filters,
+          column1: {
+            pow: {}
+          }
+        },
+        type: 'boolean'
+      };
+      const func = applyFilter(paramsWithFilterFunctionNotImplemented);
```
VictorVelarde, line 57: func could probably be called here just 'filter' ?

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+      expect(() => func(makePossibleFeature(1)[0])).toThrow('"pow" not implemented');
```
VictorVelarde, line 58: Using just first feature from this helper look strange to me.

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+    });
+
+    describe('should return true if feature passes filter', () => {
+      const feature = {
+        properties: {
+          column1: 'a',
+          column2: 1.5
+        }
+      };
+      const func = applyFilter(params)(feature);
```
VictorVelarde, line 68: let's try to find here a better name here for the **result** of a **filter** function applied. Something as `featureIsIncluded`... ?

src/tests/api/Filter.test.js
VictorVelarde, line 23: Probably obvious, but why do we get a 'number' here as result (1) instead of true / false?

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+          }
+        },
+        type: 'boolean'
+      };
+      const func = applyFilter(paramsWithFilterFunctionNotImplemented);
+      expect(() => func(makePossibleFeature(1)[0])).toThrow('"pow" not implemented');
+    });
+
+    describe('should return true if feature passes filter', () => {
+      const feature = {
```
VictorVelarde, line 62: Here you use an inline 'feature', instead of the 'helper', and honestly I find it more clear

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+    });
+
+    describe('should return false if feature not passes filter', () => {
+      const feature = {
+        properties: {
+          column1: 'a',
+          column2: 3
+        }
+      };
+      const func = applyFilter(params)(feature);
```
VictorVelarde, line 79: let's try to improve this "func", that are indeed results of func. See my comment 🔝

src/tests/api/Filter.test.js
VictorVelarde: Even if same mocks can be reused in 2 different set of tests... I would probably add them in the same test file for clarity

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,136 @@
...
+      expect(func).toBe(false);
+    });
+  });
+
+  describe('feature passes filter - number type', () => {
+    const params = { filters, type: 'number' };
+
+    describe('should return 0 if feature column value is falsy', () => {
+      const testCases = [0, null, undefined, false, ''];
```
VictorVelarde, line 88: This is a good set of column values, I like how you have included the 'limits' or not-so-common use cases ✨

src/tests/api/Filter.test.js
VictorVelarde: Indeed, after reviewing a bit the code, I would suggest a refactor in `applyFilter` to `buildFeatureFilter` as the function doesn't apply the filter. It creates a function or 'filter' for you to apply later on a single feature. If this applyFilter is not exported in the public bundle, I would go for it

src/tests/api/FilterQueryBuilder.test.js
```diff
@@ -0,0 +1,41 @@
+import { getApplicableFilters, filtersToSQL } from 'src/api/FilterQueryBuilder';
+import { filters } from './data-mocks/filters';
```
VictorVelarde, line 2: I would also suggest to inline filters here

src/tests/api/Filter.test.js
AdriSolid: - It is exported from the public bundle, and, I'm thinking that it is not necessary at all, going to remove it from there.
- Refactoring to `buildFeatureFilter`

src/tests/api/Filter.test.js
AdriSolid: With the new viewport-feature approach, 'data' could be a regular feature (GeoJSON feature), or a 'properties' GeoJSON object, like this 'rawFeature' :roll_eyes:

src/tests/api/Filter.test.js
AdriSolid: That's because at the end, deck.gl [`filterRange` prop is set to `[1, 1]`](https://github.com/CartoDB/carto-react-lib/blob/a80ec3efca842b5c7bc683cc763277d316e86d27/src/api/useCartoLayerFilterProps.js#L15).

VictorVelarde: > * Was thinking that is not necessary passing a feature here, as long as `filters` are equal to `{}`, the function firstly returns `1` or `true`: [reference](https://github.com/CartoDB/carto-react-lib/blob/a80ec3efca842b5c7bc683cc763277d316e86d27/src/api/Filter.js#L41)
> * If we pass an undefined feature the result should be `0` or `false`, [because the feature doesn't pass the filter](https://github.com/CartoDB/carto-react-lib/blob/a80ec3efca842b5c7bc683cc763277d316e86d27/src/api/Filter.js#L23).

My point here is --> if we're testing buildFeatureFilter without filters and also without a feature, how do we know (from the test, without looking into function internals) that the result depends on the missing filter params and not the missing feature? Kind-of like we're having 2 variables involved not set

src/tests/api/Filter.test.js
VictorVelarde: But that is "another thing", so we can't call it 'feature', can we? 

If we really need that, then I propose:
1. to split it into a different helper, with a good name, without 'mixing' it with the pure one, as the logic is different...
2. to remove it and use an 'inline object' in the required tests

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,180 @@
...
+
+  describe('feature passes filter - boolean type', () => {
+    const params = { filters, type: 'boolean' };
+
+    describe('should return false if feature column value is falsy', () => {
+      const columnValues = [0, null, undefined, false, ''];
+      for (const tc of columnValues) {
+        test(`${tc} - with geojson feature`, () => {
+          const [geojsonFeature] = makeFeatureWithValueInColum(tc);
+          const withProps = buildFeatureFilter(params)(geojsonFeature);
```
VictorVelarde, line 53: I would change "withProps" to "isFeatureIncluded" or similar... From my view "withProps" is related to the feature, not the result of the filter

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,180 @@
...
+        test(`${tc} - with geojson feature properties`, () => {
+          const [_, propertiesObject] = makeFeatureWithValueInColum(tc);
+          const noProps = buildFeatureFilter(params)(propertiesObject);
+          expect(noProps).toBe(false);
+        });
+      }
+    });
+
+    describe('should throw if filter function is not implemented', () => {
+      const [geojsonFeature, propertiesObject] = makeFeatureWithValueInColum(1);
```
VictorVelarde, line 68: You could use for example a makeObjectWithValueInColumn or something similar...

src/tests/api/Filter.test.js
```diff
@@ -0,0 +1,180 @@
...
+      expect(featureIsIncluded).toBe(false);
+    });
+  });
+
+  describe('feature passes filter - number type', () => {
+    const params = { filters, type: 'number' };
+
+    describe('should return 0 if feature column value is falsy', () => {
+      const columnValues = [0, null, undefined, false, ''];
+      for (const tc of columnValues) {
```
VictorVelarde, line 118: tc --> cv or value

src/tests/api/SQL.test.js
```diff
@@ -0,0 +1,80 @@
...
+      mockSqlApiRequest({ response, sql, credentials });
+      await expect(executeSQL()).rejects.toThrow(
+        `Failed to connect to ${API} API: TypeError: Cannot read property 'apiKey' of undefined`
+      );
+    });
+
+    test('should throw due status response 401', async () => {
+      mockSqlApiRequest({ response, status: 401, responseIsOk: false, sql, credentials });
+      await expect(executeSQL(credentials, sql)).rejects.toThrow(
+        `Unauthorized access to API ${API}: invalid combination of user ('${credentials.username}') and apiKey ('${credentials.apiKey}')`
```
VictorVelarde, line 55: It's curious that we have --> "API ${API}" in one case and "${API} API"  in the other. I would probably put the same order in both:  ${API} API

src/tests/api/useCartoLayerFilterProps.test.js
```diff
@@ -0,0 +1,49 @@
...
+    expect(Object.keys(result.current)).toEqual([
+      'onViewportLoad',
+      'getFilterValue',
+      'filterRange',
+      'extensions',
+      'updateTriggers'
+    ]);
+  });
+
+  describe('should has correct filter configurations', () => {
```
VictorVelarde, line 21: typo 'should have'

VictorVelarde (approved):
Nice work on tests!. Please consider my suggestions, but I don't want to block the merge, so feel free to merge this

PR #101, opened 2021-02-26
Padawannn/workers

padawannn:
To test this PR:

- Use padawannn/worker-test branch in carto-react-template
- The current deck gl version crash when you use the binary mode and you get an empty tile so you should use the 8.4.6 deck.gl version


src/api/useViewportFeatures.js
```diff
@@ -15,13 +16,23 @@ export default function useViewportFeatures(
...
+
+      await executeTask(sourceId, Methods.VIEWPORT_FEATURES, {
+        tiles: tilesCleaned,
+        viewport,
+        uniqueIdProperty
+      });
 
       dispatch(
-        setViewportFeatures({
+        setViewportFeaturesReady({
```
VictorVelarde, line 33: Doubts on naming for this... Without reading further code... what is "ready"? viewport / viewportFeatures...?

src/api/viewportFeatures.js
```diff
@@ -50,18 +50,18 @@ function addIntersectedFeaturesInTile({ map, tile, viewport, uniqueIdProperty })
 }
 
 function isTileFullVisible(bbox, viewport) {
-  const [minX, maxX, maxY, minY] = viewport;
+  const [minX, minY, maxX, maxY] = viewport;
```
VictorVelarde, line 53: was this a bug?... if we had one, we could probably change it for an object

src/api/viewportFeaturesBinary.js
```diff
@@ -0,0 +1,133 @@
...
+          Array.from(
+            positions.value.subarray(j * positions.size, (j + 1) * positions.size)
+          )
+        );
+      }
+
+      if (intersects(getFeatureByType(ringCoordinates, type), viewportIntersection)) {
+        map.set(featureId, getPropertiesFromTile(data, startIndex));
+      }
+    }
```
VictorVelarde, line 34: Could we extract this into a smaller method with a good name? Or at least the first part, dealing with coords. I think it deserves something like `const ringCoordinates = getRingCoordinatesFor(....whatever)`

src/api/viewportFeaturesBinary.js
```diff
@@ -0,0 +1,133 @@
...
+  }
+}
+
+function getIndices(data) {
+  const indices = data.primitivePolygonIndices || data.pathIndices || data.featureIds;
```
VictorVelarde, line 39: Nice small functions here and nice naming!

src/api/viewportFeaturesBinary.js
```diff
@@ -0,0 +1,133 @@
...
+  return indices.value;
+}
+
+function getFeatureId(data, startIndex) {
+  return data.featureIds.value[startIndex];
+}
+
+function getPropertiesFromTile(data, startIndex) {
```
VictorVelarde, line 47: getClonedPropertiesFromTile ?

src/api/viewportFeaturesBinary.js
```diff
@@ -0,0 +1,133 @@
...
+
+function calculateViewportFeatures({
+  map,
+  fullVisible,
+  viewportIntersection,
+  data,
+  type
+}) {
+  if (fullVisible) {
+    // All the features of the tile are visible
```
VictorVelarde, line 80: I would probably remove the comment and rename the variable the 'fullVisible' to 'allFeaturesAreVisible'

src/api/viewportFeaturesBinary.js
```diff
@@ -0,0 +1,133 @@
...
+  const map = new Map();
+
+  for (const tile of tiles) {
+    // Discard if it's not a visible tile or tile has not data
+    if (!tile.isVisible || !tile.data) {
+      continue;
+    }
+
+    const { bbox } = tile;
+    const fullVisible = isTileFullVisible(bbox, viewport);
```
VictorVelarde, line 106: rename isTileFullVisible for `isTileFullyVisible` ?

src/api/viewportFeaturesBinary.js
```diff
@@ -0,0 +1,133 @@
...
+      data: tile.data.lines,
+      type: GEOMETRY_TYPES['LineString']
+    });
+    calculateViewportFeatures({
+      map,
+      fullVisible,
+      viewportIntersection,
+      data: tile.data.polygons,
+      type: GEOMETRY_TYPES['Polygon']
+    });
```
VictorVelarde, line 129: should we avoid 3 calls with some IF or SWITCH ?

src/workers/viewportFeatures.worker.js
```diff
@@ -0,0 +1,87 @@
...
+
+function getFormula({ filters, operation, column }) {
+  let result = [{ value: null }];
+
+  if (currentViewportFeatures) {
+    const targetOperation = aggregationFunctions[operation];
+
+    const filteredFeatures = !Object.keys(currentViewportFeatures).length
+      ? currentViewportFeatures
+      : currentViewportFeatures.filter(buildFeatureFilter({ filters }));
```
VictorVelarde, line 46: These 3 lines could probably be extracted into a method, as it looks they are repeated on each method

src/api/useViewportFeatures.js
padawannn: ViewportFeaturesReady means that viewportFeatures has been calculated and are ready to be used (by widgets for example). What name do you suggest?

src/api/viewportFeatures.js
padawannn: This is how we are saving the viewport in the store. It's ok if you want to change it by an object but it's a big refactor and it should be done in other PR

src/api/viewportFeaturesBinary.js
padawannn: No, we are getting properties from the tile. Cloned is because we need to mix binary and not binary data. It is something that is part of the logic of the function and should not be part of its name because it does not provide information about what the method does

src/api/viewportFeaturesBinary.js
padawannn: The calculation that has been saved in this variable is if the entire tile is visible with what the name should be `tileIsFullVisible`. That all the features are visible is a consequence of it

src/api/viewportFeaturesBinary.js
padawannn: No. We cannot know the geometry type until we access the data and I don't see the point of spending performance on that. What we can do is put these three calls in a loop to reduce lines of code

src/workers/viewportFeatures.worker.js
padawannn: I copied the functions from models and I did not stop to see the code. I have added the `getFilteredFeatures ` function

src/api/viewportFeaturesBinary.js
padawannn: ok

src/api/viewportFeaturesBinary.js
padawannn: I would like to preserve the comment

src/api/viewportFeaturesBinary.js
padawannn: ...ok

src/api/viewportFeaturesBinary.js
VictorVelarde: MultiPolygon / MultiLineString... is that feasible?

padawannn: This code should be working with Multipoints, MultiLineString and MultiPolygon without changing anything

PR #102, opened 2021-03-07
Fix viewportFeatures mode in PieWidget

VictorVelarde:
Apply calculations from viewportFeatures when using viewportFilter in this widget

### DIFF
--- a/src/ui/utils/chartUtils.js
+++ b/src/ui/utils/chartUtils.js
@@ -1,4 +1,4 @@
-export function dataEqual(optionPrev, optionNext) {
+export function isDataEqual(optionPrev, optionNext) {
   const dataPrev = optionPrev.series[0]?.data;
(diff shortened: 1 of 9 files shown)


aaranadev (approved):

VictorVelarde:
Please @AdriSolid review the PR and add some tests

AdriSolid (commented):
Nice! Just: we need to be sure that there are no lint errors, fixed in my incoming commit.

src/ui/utils/chartUtils.js
```diff
@@ -1,4 +1,4 @@
-export function dataEqual(optionPrev, optionNext) {
+export function isDataEqual(optionPrev, optionNext) {
```
VictorVelarde, line 1: nice renaming 😄  !

PR #111, opened 2021-03-10
Add workers to calculate viewportFeatures

padawannn:
Add the use of workers for calculations of viewportFeatures, using MVT tile in binary mode.

This involves creating a new package: @carto/react-workers


packages/react-api/package.json
```diff
@@ -53,11 +53,14 @@
     "webpack-cli": "^4.5.0"
   },
   "dependencies": {
-    "@babel/runtime": "^7.13.9"
+    "@babel/runtime": "^7.13.9",
+    "@turf/bbox-polygon": "^6.3.0",
```
VictorVelarde, line 57: @padawannn I thought these turf deps were not required now here, are they?

packages/react-widgets/src/models/CategoryModel.js
```diff
@@ -63,59 +55,30 @@ export const buildSqlQueryToGetCategories = ({
...
   operationColumn,
-  filters,
-  alias = 'name'
+  filters
 }) => {
   const query = `
     WITH all_categories as (
       SELECT
-        ${column} as ${alias}
+        ${column} as category
```
VictorVelarde, line 63: I think removing the alias breaks PieWidget. Now 'name' is used for both CategoryW and PieW, but we also have the external param for that.

packages/react-widgets/src/models/CategoryModel.js
```diff
@@ -34,9 +29,7 @@ export const getCategories = async (props) => {
   const operationColumn = props.operationColumn || column;
 
   if (viewportFilter) {
-    return filterViewportFeaturesToGetCategories({
```
VictorVelarde, line 37: I'm not sure if we should force the use of workers here in the models. It feels like we're losing a path to calculate with geojson features in the near-future

packages/react-widgets/src/models/FormulaModel.js
```diff
@@ -16,7 +12,7 @@ export const getFormula = async (props) => {
     filters,
     opts,
     viewportFilter,
-    viewportFeatures,
+    dataSource,
```
VictorVelarde, line 15: Same as in previous. I think we should probably preserve both behaviours, binary with workers and current functions for plain common features. The strategy for the calculation method could be hidden, depending on the params. What do you think?

packages/react-widgets/src/models/HistogramModel.js
```diff
@@ -17,7 +13,7 @@ export const getHistogram = async (props) => {
     filters,
     opts,
     viewportFilter,
-    viewportFeatures,
+    dataSource,
```
VictorVelarde, line 16: same

packages/react-workers/src/workerPool.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+function getWorker(source) {
+  if (!pool[source]) {
+    pool[source] = {
+      worker: new viewportFeaturesWorker(),
+      tasks: []
+    };
+    onmessage(pool[source]);
```
VictorVelarde, line 34: Maybe rename to addOnMessageHandler / addOnErrorHandler ?

Also I would use const worker = pool[source] as the first line, and then substitute other calls, if possible, with just 'worker'

VictorVelarde (changes requested):
Perf is really nice, good job!.

To move on, we still need to apply some changes:
- [x] fix 'category' in CategoryModel, so it keeps compatibility with PieWidget
- [x] bring this branch up to latest master
- [x] broader discussion on feature: always binary vs future options (+ bqtiler specifics)
- [ ] comment failing tests, to go 'green' (we'll need to add more specific ones to vwf)
- [ ] review with @alasarr

packages/react-api/package.json
padawannn: you're right

packages/react-widgets/src/models/CategoryModel.js
padawannn: Done

packages/react-api/package.json
```diff
@@ -53,12 +53,10 @@
     "webpack": "^5.24.2",
     "webpack-cli": "^4.5.0"
   },
-  "dependencies": {
-    "@babel/runtime": "^7.13.9"
```
VictorVelarde, line 57: @padawannn Is there a particular reason to remove here the runtime?

packages/react-workers/src/workerPool.js
padawannn: In this case, I prefer the native names.

`Also I would use const worker = pool[source] as the first line, and then substitute other calls` I think you're losing your mind with the Clean Code course XD. Seriously, that change doesn't add any clarity to the code.

packages/react-api/package.json
padawannn: ups

packages/react-core/src/filters/viewportFeaturesBinary.js
```diff
@@ -0,0 +1,169 @@
...
+  const { properties, numericProps } = getPropertiesFromTile(data, startIndex);
+  return Object.assign({}, properties, numericProps);
+}
+
+function getUniquePropertyValue (data, startIndex, uniqueIdProperty, map) {
+  const { properties, numericProps } = getPropertiesFromTile(data, startIndex);
+  if (uniqueIdProperty) {
+    return numericProps[uniqueIdProperty] || properties[uniqueIdProperty];
+  }
+  return numericProps['cartodb_id'] || numericProps['geoid'] || properties['cartodb_id'] || properties['geoid'] || (map.size + 1);
```
VictorVelarde, line 63: @padawannn What's the meaning here of map.size + 1 ?

packages/react-core/src/filters/viewportFeaturesBinary.js
```diff
@@ -0,0 +1,169 @@
...
+}
+
+function getFeatureByType(coordinates, type) {
+  switch (type) {
```
VictorVelarde, line 67: Will this manage multi-geometry correctly without further TYPEs?

packages/react-core/src/filters/viewportFeaturesBinary.js
```diff
@@ -0,0 +1,169 @@
...
+  viewportIntersection,
+  data,
+  type,
+  uniqueIdProperty
+}) {
+  if (tileIsFullyVisible) {
+    // All the features of the tile are visible
+    const indices = getIndices(data);
+
+    for (let i = 0; i < indices.length - 1; i++) {
```
VictorVelarde, line 103: @padawannn Is this < (length - 1) right? will we miss the last index here?

packages/react-widgets/src/models/CategoryModel.js
VictorVelarde: We already decided on this

packages/react-core/src/filters/viewportFeaturesBinary.js
padawannn: It's a trick. The intersected viewport features are stored in the `map` variable that is a `Set`. The unique key of the set is the `uniqueIdProperty`but with aggregated point we don't have a `uniqueIdProperty ` so as not to have to do a big refactor I return `map.size + 1` that always going to be different.

packages/react-core/src/filters/viewportFeaturesBinary.js
padawannn: Yes because in binary format multi-geometry are stored as single-geometry with the same `featureId`

packages/react-core/src/filters/viewportFeaturesBinary.js
padawannn: Yes, it's right

padawannn: check this line https://github.com/CartoDB/carto-react/blob/25c83daa6803a8573a1242dfe6e15597239ce799/packages/react-core/src/filters/viewportFeaturesBinary.js#L17

VictorVelarde: Oh, I see. Then it looks 👍🏻

VictorVelarde (approved):
👍🏻

packages/react-core/src/filters/viewportFeaturesBinary.js
VictorVelarde: Understood, thx!. With our refactor is now clearer!

PR #113, opened 2021-03-11
fix: histogram with values zero & uniqueIdProperty from cartoLayerProps hook

AdriSolid:
For: https://app.clubhouse.io/cartoteam/story/141415/bug-thinking-machines-histogram
For: https://app.clubhouse.io/cartoteam/story/141411/uniqueidproperty-behaviour-in-hook


packages/react-api/__tests__/hooks/useCartoLayerFilterProps.test.js
```diff
@@ -1,15 +1,16 @@
 import { DataFilterExtension } from '@deck.gl/extensions';
 import { renderHook } from '@testing-library/react-hooks';
 import useCartoLayerProps from '../../src/hooks/useCartoLayerProps';
```
VictorVelarde, line 3: The name of this test file is OLD (LayerFilterProps...) and has to change

packages/react-api/__tests__/hooks/useCartoLayerFilterProps.test.js
```diff
@@ -1,15 +1,16 @@
...
-import { mockClear, mockReduxHooks } from '../mockReduxHooks';
+import { mockReduxHooks, mockClear } from '../mockReduxHooks';
```
VictorVelarde, line 4: why this order change, automatic?

packages/react-api/__tests__/hooks/useCartoLayerFilterProps.test.js
AdriSolid: Good catch!

packages/react-api/reference.md
```diff
@@ -25,6 +25,7 @@ Returns required default props for layers. It manages filtering and viewport cha
 | Param                               | Type                          | Description                                                               |
 | ----------------------------------- | ----------------------------- | ------------------------------------------------------------------------- | ---------------------------------------------------- |
 | props                               | <code>Object</code>           | Required default props for layers                                         |
+| props.uniqueIdProperty              | <code>string</code>           | Unique id property for the layer, default to `cartodb_id`                 |
```
VictorVelarde, line 28: We must also add the uniqueIdProperty as output right?

packages/react-api/__tests__/hooks/useCartoLayerFilterProps.test.js
AdriSolid: Make more sense right? First, initialization, then clearing.

packages/react-api/reference.md
AdriSolid: That block is actually the output

packages/react-core/src/operations/histogram.js
```diff
@@ -17,7 +17,9 @@ export function histogram(features, columnName, ticks, operation) {
   features.forEach((feature) => {
     const featureValue = feature[columnName];
 
-    if (!featureValue) {
+    const isValid = featureValue !== null && featureValue !== undefined;
```
VictorVelarde, line 20: 🤦🏻 ... falsy value bites us again...

packages/react-api/__tests__/hooks/useCartoLayerFilterProps.test.js
VictorVelarde: Ok, I tend to use alphabetic ordering, but it's ok. We should see if we can enforce a criteria in the future with linter

packages/react-api/reference.md
VictorVelarde: sorry man, I didn't notice 👍🏻

VictorVelarde (approved):
👍🏻

PR #121, opened 2021-03-15
Improve googlemap control, considering optional props

VictorVelarde:
To better document and improve its use.

It could even be used externally, with something like: https://codesandbox.io/s/naughty-engelbart-wmrmc?file=/src/App.js

### DIFF
--- a/packages/react-basemaps/src/basemaps/GoogleMap.js
+++ b/packages/react-basemaps/src/basemaps/GoogleMap.js
@@ -6,15 +6,28 @@ import { debounce } from '@carto/react-core';
  * React component for working with Google Maps API and deck.gl
  *
  * @param { Object } props - Properties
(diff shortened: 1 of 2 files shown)


packages/react-basemaps/src/basemaps/GoogleMap.js
```diff
@@ -69,17 +79,18 @@ export function GoogleMap(props) {
 
         if (JSON.stringify(window.cartoViewState) !== JSON.stringify(viewState)) {
           window.cartoViewState = viewState;
-          props.onViewStateChange({ viewState });
+          onViewStateChange && props.onViewStateChange({ viewState });
```
AdriSolid, line 82: Nice!

AdriSolid (approved):
:+1:

PR #123, opened 2021-03-16
Fix histogram removing last spurious bin

VictorVelarde:
This is a use case to detect the issue:

```js
       const features = [
            { field: 1 },
            { field: 2 }, { field: 2 },
            { field: 3 }, { field: 3 }, { field: 3 },
            { field: 4 }, { field: 4 },
            { field: 5 }            
        ];

        const ticks = [2, 4, 6];

        const h = histogram(features, 'field', ticks, AggregationTypes.COUNT);        
        console.log(h); // [1, 5, 3, 0, 0]
```

Results 

Histogram results should be interpreted as:
            >=MIN to <2     --> 1 item
            >=2 to <4       --> 5 items
            >=4 to <6       --> 3 items
            >=6 to MAX      --> 0 items

BUT there is an extra bin, with value 0, that shouldn't be there (unless I'm missing something non obvious)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add Widgets from @carto/react-widgets to StoryBook [#120](https://github.com/CartoDB/carto-react/pull/120)
 - Improve GoogleMap component [#121](https://github.com/CartoDB/carto-react/pull/121)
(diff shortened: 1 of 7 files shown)


VictorVelarde:
We still have to fix tests on viewportFeatures. that I missed

padawannn (approved):

PR #124, opened 2021-03-17
Fix material-ui warnings due to wrong styles in theme

padawannn:
- Warnings violation about `ms` cannot be resolved because it comes from chrome and are shown when an action takes a lot of time.
- SharedArrayBuffer warning comes from react https://github.com/facebook/create-react-app/issues/10474

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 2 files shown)


VictorVelarde (approved):
Nice, please add a clear note to changelog (unreleased), like "Fix material-ui warnings due to wrong styles in theme" and merge it

PR #128, opened 2021-03-21
Fix FormulaWidgetUI render when data is zero

VictorVelarde:
When data was 0, a render was not happening, leaving the widget with the previous value, causing a bug

image

### DIFF
--- a/packages/react-ui/__tests__/widgets/FormulaWidgetUI.test.js
+++ b/packages/react-ui/__tests__/widgets/FormulaWidgetUI.test.js
@@ -3,15 +3,6 @@ import { render, screen } from '@testing-library/react';
 import FormulaWidgetUI from '../../src/widgets/FormulaWidgetUI';
 import { currencyFormatter } from './testUtils';
 
-// Mock animations
-jest.mock('../../src/widgets/utils/animations');
-// eslint-disable-next-line import/first
-import { animateValue } from '../../src/widgets/utils/animations';
-
-animateValue.mockImplementation(({ end, drawFrame }) => {
-  drawFrame(end); // draw final step, no intermediate ones
-});
-
 describe('FormulaWidgetUI', () => {
   test('empty', () => {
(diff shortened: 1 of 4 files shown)


packages/react-ui/__tests__/widgets/FormulaWidgetUI.test.js
```diff
@@ -3,15 +3,6 @@ import { render, screen } from '@testing-library/react';
 import FormulaWidgetUI from '../../src/widgets/FormulaWidgetUI';
 import { currencyFormatter } from './testUtils';
 
-// Mock animations
```
VictorVelarde, line 6: I added this cause I noticed these tests were unstable, when using the real animation steps... I guess that adding the rerender + await removes the problem, but just for you to know

packages/react-ui/__tests__/widgets/FormulaWidgetUI.test.js
```diff
@@ -36,6 +27,12 @@ describe('FormulaWidgetUI', () => {
     expect(await screen.findByText(DATA.value)).toBeInTheDocument();
   });
 
+  test('should render the current value', async () => {
+    const { rerender } = render(<FormulaWidgetUI data={1234} />);
+    rerender(<FormulaWidgetUI data={0} />);
```
VictorVelarde, line 32: @AdriSolid I was thinking about specific tests, in a different file, for animations, not the test on the widget removing the mock, but I guess this is also nice (if we don't need to use a mock anymore)

packages/react-ui/__tests__/widgets/FormulaWidgetUI.test.js
AdriSolid: Yep, [check this out](https://testing-library.com/docs/guide-disappearance/#1-using-findby-queries)

AdriSolid (approved):
:+1:

PR #138, opened 2021-04-08
feat: expose createTheme as createCartoTheme

AdriSolid:
Expose `cartoTheme` as `createCartoTheme`.

Related with `documentation`: https://github.com/CartoDB/documentation/pull/78

### DIFF
--- a/packages/react-ui/src/theme/carto-theme.js
+++ b/packages/react-ui/src/theme/carto-theme.js
@@ -1175,16 +1175,10 @@ export const cartoThemeOptions = {
   }
 };
 
-export function createTheme(options = {}) {
-  const themeOptions = {
-    ...cartoThemeOptions,
-    ...options
-  };
-
-  let theme = createMuiTheme(themeOptions);
-
+export function createTheme() {
+  let theme = createMuiTheme(cartoThemeOptions);
   theme = responsiveFontSizes(theme, {
-    breakpoints: themeOptions.breakpoints.keys,
+    breakpoints: cartoThemeOptions.breakpoints.keys,
     disableAlign: false,
     factor: 2,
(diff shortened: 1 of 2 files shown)


packages/react-ui/src/theme/carto-theme.js
```diff
@@ -1175,16 +1175,10 @@ export const cartoThemeOptions = {
   }
 };
 
-export function createTheme(options = {}) {
```
VictorVelarde, line 1178: Are we sure we can remove default options?

VictorVelarde:
Please add a note to changelog and link here the PR with the doc change

packages/react-ui/src/theme/carto-theme.js
AdriSolid: As described in `documentation` PR, the way to overwrite `cartoThemeOptions` is equal for both implementations in this way.

Also, passing to `createTheme` and object like:
```javascript
const mainColor = { palette: { primary: { main: '#80000' }}}
```
results in an error. Do you have any other idea to handle this?

VictorVelarde: What if we leave the optional param, and just expose the method?

That way we could also allow overriding certain properties, so we would eventually get 3 options for the user:

- call to createTheme without params (one line from the client)
- call to createTheme with certain options (maybe just colors)
- completely custom case, using just cartoThemeOptions (what is currently being used from template)

/cc @borja-munoz

VictorVelarde:
We should talk more about this one with @borja-munoz, before exposing it

VictorVelarde:
Not a clear advantage on the change, so I'm closing this

PR #141, opened 2021-04-26
Update CategoryWidget.js to addres RFC ch150631

juandjara:
RFC link: https://app.clubhouse.io/cartoteam/story/150631/allow-disable-filters-and-allow-having-selected-categories-on-widget-init

### DIFF
--- a/packages/react-widgets/src/widgets/CategoryWidget.js
+++ b/packages/react-widgets/src/widgets/CategoryWidget.js
@@ -80,11 +80,16 @@ function CategoryWidget(props) {
     hasLoadingState,
     onError
   ]);
+  
+  useEffect(() => {
+    const selected = filters[column] && filters[column][FilterTypes.IN] && filters[column][FilterTypes.IN].values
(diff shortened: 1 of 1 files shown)


VictorVelarde:
Hi, @juandjara 2 questions to have a bit more of context for this:

1. Wouldn't this apply the same for other widgets (like PieWidget)? If so, then I think all adaptations should be included in this PR.
2. Do you have or can you prepare a small example (eg codesandbox) where you need & use this piece of code, so we understand better the use case?

VictorVelarde:
Also, an important aspect: have you tried locally this PR with packages linked to **carto-react-template** project? 

That's required to eval compatibility or detect any issue. Right now, linked to `sample app`, I get a `TypeError: Cannot read property 'storetype' of undefined`. Could you please check it?

juandjara:
Hi @VictorVelarde thanks for your response.
Unfortunately, I'm dealing with a difficult bug / implementation right now and cannot check this. I'll try to explain more / test more as soon as I finish

PR #147, opened 2021-05-27
Add carto-cloud-native typings

AdriSolid:
Add new carto-cloud-native typings


packages/react-api/src/types.d.ts
```diff
@@ -1,25 +1,43 @@
...
+import { MAP_TYPES, API_VERSIONS } from '@deck.gl/carto';
 import { FeatureCollection } from 'geojson';
 
+declare enum DEFAULT_MAPS_URLS {
+  V1 = 'https://{user}.carto.com/api/v1/map',
+  V2 = 'https://maps-api-v2.{region}.carto.com/user/{user}',
+  V3 = 'https://maps-{tenant}'
+}
+
 export type Credentials = {
```
alasarr, line 11: Let's do something like this, to separate credentials for v1 and v2.
```javascript
interface CredentialsCarto {
  apiVersion: API_VERSIONS.V1 | API_VERSIONS.V2,
  username: string,
  apiKey: string,
  region?: string,
  mapsUrl?: string,
  serverUrlTemplate?: string
}

interface CredentialsCartoCloudNative {
  apiVersion: API_VERSIONS.V3,
  apiBaseUrl: string,
  accessToken?: string,
}

export type Credentials = CredentialsCarto | CredentialsCartoCloudNative;
```

alasarr (changes requested):
Let's make clearer the difference between carto vs cartoCloudNative credentials.

I've also found we have a type like this:
```
export enum SourceTypes {
  SQL = 'sql',
  BIGQUERY = 'bigquery'
}
```
And it's used in some tests. It should use the [MAP_TYPES defined in @carto/deck.gl](https://github.com/CartoDB/carto-deck.gl/blob/carto-cloud-native/modules/carto/src/api/maps-api-common.js#L4)

alasarr:
References to SourceTypes.SQL should be removed

VictorVelarde (approved):
👍🏻

PR #149, opened 2021-06-08
Add Scatter plot widget & deck.gl update to alpha 10

VictorVelarde:
This is a sample integration within react-template

image


VictorVelarde:
THIS SHOULD BE PUT IN SYNC WITH MASTER

PR #150, opened 2021-06-14
SQL API cloud native

alasarr:
This PR gives support to SQL API for CARTO cloud native (v3)

It also supports CARTO SQL API v1 and v2.

There is a breaking change on the API, since executeSQL parameters are now destructuring.

How to use:

**Create a model**

```javascript
import { executeSQL } from '@carto/react-api';

const CONNECTION = 'bigquery';

export const getTotalArea = async ({ credentials, opts }) => {
  const query = `
  select SUM(do_area) as area from carto-do-public-data.carto.geography_aus_state_2016 
  `;

  return executeSQL({ credentials, connection: CONNECTION, query, opts });
};
```

**Call from a view**

```javascript
import { getTotalArea } from '../../data/models/BigQueryModel';
 const credentials = useSelector((state) => state.carto.credentials);

  useEffect(() => {
    getTotalArea({ credentials }).then( d => {
      // whatever
    })

  }, [ credentials ]);
```

### DIFF
--- a/packages/react-api/src/api/SQL.d.ts
+++ b/packages/react-api/src/api/SQL.d.ts
@@ -1,9 +1,13 @@
 import { Credentials, ExecuteSQL } from '../types';
 
-export function executeSQL(
-  credentials: Credentials,
-  query: string,
-  opts?: {
-    format: '' | 'geojson'
-  }
-): ExecuteSQL;
+export function executeSQL({
+    credentials,
+    query,
+    connection,
+    opts 
+  }: {
+    credentials: Credentials,
+    query: string,
+    connection?: string,
+    opts?: unknown
+  }): ExecuteSQL;
--- a/packages/react-api/src/api/SQL.js
+++ b/packages/react-api/src/api/SQL.js
@@ -18,6 +18,10 @@ import { dealWithApiError, generateApiUrl } from './common';
 export const executeSQL = async ({ credentials, query, connection, opts }) => {
   let response;
 
+  if (!credentials) {
+    throw new Error('No credentials provided');
+  }
+
   try {
     const request = createRequest({ credentials, connection, query, opts });
     response = await fetch(request);
@@ -46,14 +50,14 @@ export const executeSQL = async ({ credentials, query, connection, opts }) => {
  * Create an 'SQL query' request
  * (using GET or POST request, depending on url size)
  */
-function createRequest({ credentials, connection, query, opts }) {
+function createRequest({ credentials, connection, query, opts = {} }) {
   const { abortController, ...otherOptions } = opts;
 
   const { apiVersion = API_VERSIONS.V2 } = credentials;
 
   const rawParams = {
     client: 'carto-react',
-    q: query.trim(),
+    q: query?.trim(),
     ...otherOptions
   };
 
--- a/packages/react-api/src/api/SQL.js
+++ b/packages/react-api/src/api/SQL.js
@@ -1,9 +1,8 @@
 import { encodeParameter, getRequest, postRequest } from '@carto/react-core';
 import { REQUEST_GET_MAX_URL_LENGTH } from '@carto/react-core';
+import { API_VERSIONS } from '@deck.gl/carto';
 
 import { dealWithApiError, generateApiUrl } from './common';
-
-export const API = 'api/v2/sql';
 
 /**
  * Executes a SQL query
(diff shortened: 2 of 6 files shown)


alasarr:
@borja-munoz ☝️

packages/react-api/src/api/SQL.d.ts
```diff
@@ -1,9 +1,13 @@
...
-  credentials: Credentials,
-  query: string,
-  opts?: {
-    format: '' | 'geojson'
-  }
-): ExecuteSQL;
\ No newline at end of file
+export function executeSQL({
+    credentials,
+    query,
```
VictorVelarde, line 5: Once we're considering breaking the interface... should we rename 'query' into 'sql'? IMO "query" suggests that just 'SELECT' is supported...

packages/react-api/src/api/SQL.js
```diff
@@ -13,25 +12,35 @@ export const API = 'api/v2/sql';
  * @param { string } credentials.apiKey - CARTO API Key
  * @param { string } credentials.serverUrlTemplate - CARTO server URL template
  * @param { string } query - SQL query to be executed
+ * @param { string } connection - connection name required for CARTO cloud native
  * @param { Object } opts - Additional options for the HTTP request
- * @param { string } opts.format - Output format (i.e. geojson)
  */
-export const executeSQL = async (credentials, query, opts = { format: '' }) => {
+export const executeSQL = async ({ credentials, query, connection, opts }) => {
```
VictorVelarde, line 18: Isn't the default 'opts' required anymore?

packages/react-api/src/api/SQL.d.ts
alasarr: I used query because it's the name chosen at the apis and to keep coherence

packages/react-api/src/api/SQL.js
```diff
@@ -41,16 +50,23 @@ export const executeSQL = async (credentials, query, opts = { format: '' }) => {
  * Create an 'SQL query' request
  * (using GET or POST request, depending on url size)
  */
-function createRequest({ credentials, query, opts }) {
+function createRequest({ credentials, connection, query, opts = {} }) {
```
VictorVelarde, line 53: ok, I see opts = {} 👍🏻

packages/react-api/src/api/SQL.js
```diff
@@ -41,16 +50,23 @@ export const executeSQL = async (credentials, query, opts = { format: '' }) => {
...
   const { abortController, ...otherOptions } = opts;
 
+  const { apiVersion = API_VERSIONS.V2 } = credentials;
```
VictorVelarde, line 56: nice destructuring with default

packages/react-api/src/api/SQL.js
alasarr: Yes, I didn't know why it was removed :).

We've removed the format because in cloud native opts allow to override all the parameters of each cloud provider. besides format we've timeoutms, maxBytesProcess, maxRows, dryRun, etc...

packages/react-api/src/api/SQL.js
alasarr: I'm getting a master in javascript from the best people :)

packages/react-api/src/api/SQL.js
```diff
@@ -41,16 +50,23 @@ export const executeSQL = async (credentials, query, opts = { format: '' }) => {
...
-    api_key: credentials.apiKey,
-    client: credentials.username,
-    q: query.trim(),
+    client: 'carto-react',
+    q: query?.trim(),
     ...otherOptions
   };
 
+  if (apiVersion === API_VERSIONS.V3) {
+    rawParams.access_token = credentials.accessToken;
```
VictorVelarde, line 65: It looks like the token we're passing here will appear in the url, is that fine? Another option would be trying to inject it in the headers, via requestOpts...

alasarr: Yes, I do that in deck.gl module. https://github.com/visgl/deck.gl/blob/master/modules/carto/src/api/maps-cloud-native-client.js#L21

I didn't do here because the purpose of doing that is to skip the cache, but sql endpoint is not cached in cloud native api

alasarr: Maybe I can do it, but I prefer in a separated PR as an improvement

VictorVelarde: Cool

VictorVelarde (approved):
LGTM. Just please add a note to CHANGELOG with a note like 'Add SQL API support for new v3 api'...

alasarr:
>LGTM. Just please add a note to CHANGELOG with a note like 'Add SQL API support for new v3 api'... and a BREAKING CHANGE note about the API for previous versions

I did but [here](https://github.com/CartoDB/carto-react/pull/151)

PR #154, opened 2021-06-22
Table Widget: core functionality

builtbyedgar:
Screenshot from integration in carto-react-template sample app 3

Capture d’écran de 2021-12-21 18-20-43

This PR introduces `TableWidgetUI` as a new UI component and `TableWidget` as a new widget.

For now, supports sorting, pagination and click on row behavior.
Selection, search and export will be managed in further PRs.

`TableWidgetUI` behavior is fully controlled, so filtering, sorting and pagination has to be done by the parent component. This allows `TableWidget` to use the new `getRawFeatures` API from https://github.com/CartoDB/carto-react/pull/225, to access and sort features in an efficient way.


packages/react-ui/src/widgets/TableWidgetUI.js
```diff
@@ -0,0 +1,387 @@
...
+  });
+  return stabilizedThis.map((el) => el[0]);
+}
+
+const useStyles = makeStyles((theme) => ({
+  tableHead: {
+    backgroundColor: theme.palette.common.white,
+    '& .MuiTableCell-head': {
+      border: 'none'
+    },
```
aaranadev, line 47: The style are without nested, please change the style to Carto4React theme

aaranadev (changes requested):

packages/react-ui/package.json
```diff
@@ -80,6 +80,7 @@
     "echarts": "^4.9.0",
     "echarts-for-react": "^2.0.16",
     "react": "^17.0.1",
-    "react-dom": "^17.0.1"
+    "react-dom": "^17.0.1",
+    "styled-components": "^5.1.0"
```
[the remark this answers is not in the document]
VictorVelarde: I hope it is not... 😄

bbecquet:
Shortcut link: https://app.shortcut.com/cartoteam/story/183138/implement-table-widget-basic-functionality

bbecquet:
Shortcut link: https://app.shortcut.com/cartoteam/story/183138/implement-table-widget-basic-functionality

packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
```diff
@@ -0,0 +1,209 @@
...
+        </Table>
+      </TableContainer>
+      {pagination && (
+        <TablePagination
+          className={classes.pagination}
+          rowsPerPageOptions={rowsPerPageOptions}
+          component='div'
+          count={totalCount}
+          rowsPerPage={rowsPerPage}
+          page={page - 1}
```
bbecquet, line 113: We need to make this adjustment because `TablePagination` from MUI is zero-based whereas our pagination API is 1-based (see https://github.com/CartoDB/carto-react/blob/master/packages/react-workers/src/workers/features.worker.js#L166)

Or maybe we could change our own implementation to start at page 0 too? Would it be possible @Clebal?

padawannn (changes requested):
The PR looks really web but I'm wondering about the behavior when we change the page and after that, the data changes because the viewport has changed or we've applied a filter. In that case, we could be consulting a page that no longer exists and the table would show no data.

@alasarr, @borja-munoz we need to talk about what to do in that case (because is a very common case). Probably, we would need to go back to the first page every time that the data changes.

alasarr:
Looks good to me for a first version. Remember to update the changelog.

bbecquet:
@padawannn about your comment on pagination vs. viewport, I've added this piece of code https://github.com/CartoDB/carto-react/pull/154/commits/17454dd4cdb14306df6c5f104583ab0c0d17dcc8 which resets the page to 1 when the features change.
It uses the fact that the `ready` state of the source goes through the cycle `true` -> `false` -> `true` when the viewport changes.
Do you think it's robust enough?

padawannn:
> @padawannn about your comment on pagination vs. viewport, I've added this piece of code [17454dd](https://github.com/CartoDB/carto-react/commit/17454dd4cdb14306df6c5f104583ab0c0d17dcc8) which resets the page to 1 when the features change. It uses the fact that the `ready` state of the source goes through the cycle `true` -> `false` -> `true` when the viewport changes. Do you think it's robust enough?

@bbecquet we also need to reset the page to 1 if filters change. For example, in your branch template when you click hover a category in the category widget. We also should reset the page to 1  if the datasource change although this case is not common

padawannn (changes requested):
https://github.com/CartoDB/carto-react/pull/154#issuecomment-1003088681

bbecquet:
@padawannn Thanks for the explanation about filters. I've changed the reset callback.

padawannn (changes requested):
@bbecquet you forgot to include the `dataSource` as a reset condition although it is not a common case.

On the other hand, it could be great if we add some tests about page reset

padawannn (approved):

padawannn (approved):

packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
Clebal: @bbecquet totally agree, it's a mistake that page is 1-based. Sorry for my late response, I didn't saw this. Can you open a PR with that change?

PR #159, opened 2021-07-12
Rename Credentials exported types

aaranadev:
Story details: https://app.clubhouse.io/cartoteam/story/167584


VictorVelarde:
Leaving open for a couple of days, in case it's useful, but it looks like there are naming changes, so not applicable as it is

VictorVelarde:
Ok then I wasn't sure they were up-to-date. Let me then check with more detail

packages/react-api/src/types.d.ts
```diff
@@ -11,13 +11,13 @@ interface CredentialsCarto {
   serverUrlTemplate?: string
 }
 
-interface CredentialsCartoCloudNative {
```
VictorVelarde, line 14: After talking to @borja-munoz, it looks like these names should be changed, so commiting a small rename now

VictorVelarde (approved):
👍🏻

PR #167, opened 2021-08-05
Update echarts to v5

Clebal:

PR #168, opened 2021-09-06
Time Series Widget

Clebal:
This PR needs to be merged after #167. echarts v5 is totally necessary for this development.


packages/react-api/src/hooks/useViewportFeatures.js
```diff
@@ -26,6 +26,13 @@ export default function useViewportFeatures(
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const computeFeaturesTileset = useCallback(
     debounce(async ({ tiles, viewport, uniqueIdProperty, sourceId }) => {
+      dispatch(
+        setViewportFeaturesReady({
+          sourceId,
+          ready: false
+        })
+      );
```
VictorVelarde, line 34: Was the lack of this dispatch here causing a bug in calculations?

packages/react-core/src/operations/groupByDate.js
```diff
@@ -0,0 +1,72 @@
...
+      }))
+      .sort((a, b) => a.name - b.name);
+  }
+
+  return [];
+}
+
+// Aux
+
+// TODO: Somewhere else maybe?
```
VictorVelarde, line 64: I agree, as this is duplicated in `TimeSeriesWidgetUI`. I would suggest something like a function inside utils/dateUtils.js, exported from core and consumed from other possible clients

packages/react-api/src/hooks/useViewportFeatures.js
Clebal: Not a bug, but misinformation. I didn't know when the source wasn't ready. Other widgets know when to change using widget loading state, but that cannot be applied in time series because it's applying a filter constantly and then the widget was updating all the time. That's why I need to know when the source is ready or not.

VictorVelarde:
Regarding echarts update, I would say that an upgrade to 5.2.0 in the previous independent PR would be great, so it can be merged and then tested before this one

packages/react-core/src/operations/groupByDate.js
Clebal: Done

Clebal:
I've added new comments for a better code understanding and reviewed some `eslint-disable` to avoid them as much as possible.

I splited a useEffect responsible of stopping the timeline position animation and reporting to the widget about the updates. The stop logic is now located in the useEffect responsible of updating the timeline position state. Thanks to that, right now that useEffect is responsible of whole running logic for timeline.

The reporting logic is in the context like the one for timeframe updates.

packages/react-api/src/hooks/useViewportFeatures.js
padawannn: I don't know if I understand you very well. If the time series it's applying a filter constantly it's like changing any other filter so you don't need setViewportFeaturesReady to false, right? Also, I think this would change the flow in the rest of the widgets

padawannn (changes requested):

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -0,0 +1,303 @@
...
+    }
+    // Only on bucketSize update
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [stepSize]);
+
+  const [isLoading, setIsLoading] = useState(false);
+
+  const filters = useSelector((state) => {
+    const source = selectSourceById(state, dataSource) || {};
+    const filters = { ...source.filters };
```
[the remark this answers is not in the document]
Clebal: Other widgets should do it also. Each time you select a category in CategoryWidget, the useEffect to get the data to feed the widget is being executed again, to obtain the exact same result because the filter that made the useEffect to update isn't being applied to the data. In the CategoryWidget it doesn't matter because you don't have 60 filters per second, but in TimeSeries yes.

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -0,0 +1,303 @@
...
+    delete filters[column];
+    return filters;
+  });
+
+  const isSourceReady = !!viewportFeaturesReady[dataSource];
+
+  useEffect(() => {
+    if (isSourceReady) {
```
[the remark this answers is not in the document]
Clebal: Because data check is always true after the first fetch. As I said in my first comment, others widget has the same bug I want to avoid but the bug is tolerable in them because there's no need to apply a lot of filters in a short amount of time.

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -0,0 +1,303 @@
...
+      setIsLoading(true);
```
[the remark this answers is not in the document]
Clebal: The same reason as in my first comment. If I don't control the loading by myself, the component will be always loading because each time a filter is applied all the widgets loading state are set to true because the loading state is centralised, and each time the source changes, every widget fed by that source is called to load (changing isLoading from false to true). I don't think it's good.

packages/react-api/src/hooks/useViewportFeatures.js
Clebal: Yes, in any other widget, when it applies a filter, the useEffect for fetching data is executed to get the exact same data. It doesn't happen in TimeSeries widget, it only fetches data when: A filter change or source's ready state changes from false to true.

packages/react-ui/src/widgets/time-series-widget-ui/TimeSeriesWidgetUI.js
```diff
@@ -0,0 +1,424 @@
...
+  const [animationStep, setAnimationStep] = useState(0);
+
+  // Calculate animation step by duration
+  useEffect(() => {
+    if (duration && data.length) {
+      setAnimationStep(duration / data.length);
+    }
+  }, [duration, data, setAnimationStep]);
+
+  return (
```
padawannn, line 61: I'm curious why you are using context. Normally, context is used when data needs to be accessible by many components at different nesting levels. Is not your case, right?

packages/react-ui/src/widgets/time-series-widget-ui/hooks/TimeSeriesContext.js
```diff
@@ -0,0 +1,147 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [timelinePosition]);
+
+  useEffect(() => {
+    if (timeframe[0] !== _timeframe[0] && timeframe[1] !== _timeframe[1]) {
+      setTimeframe(timeframe);
+    }
+    // Only when the state out of the context changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [...timeframe]);
```
padawannn, line 124: This useEffect condition is correct? It always returns a new object so what is the sense of using a useEffect?

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -0,0 +1,308 @@
...
+  dataSource,
+  column,
+  operationColumn,
+  operation,
+  stepSizeOptions,
+  onError,
+  wrapperProps,
+  noDataAlertProps,
+  // UI
+  chartType,
```
padawannn, line 33: What about setting a default chart type

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -0,0 +1,308 @@
...
+  id,
+  title,
+  dataSource,
+  column,
+  operationColumn,
+  operation,
+  stepSizeOptions,
+  onError,
+  wrapperProps,
+  noDataAlertProps,
```
padawannn, line 31: This widget is the only one that can set noDataAlertProps, is it ok @borja-munoz

padawannn (changes requested):

packages/react-widgets/src/widgets/TimeSeriesWidget.js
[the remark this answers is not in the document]
VictorVelarde: not applicable now

packages/react-ui/src/index.js
```diff
@@ -15,6 +23,12 @@ export {
   HistogramWidgetUI,
   PieWidgetUI,
   ScatterPlotWidgetUI,
+  TimeSeriesWidgetUI,
+  useTimeSeriesContext,
```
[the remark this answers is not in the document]
Clebal: Export isn't only about sharing between modules. I think it's good to make it available for users to let them play with that pieces.

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -0,0 +1,308 @@
...
+                  name: 'Bucket size',
+                  icon: <StepSizeIcon />,
+                  action: handleOpenBucketSizeMenu
+                }
+              ]
+            : [])
+        ]}
+      >
+        {!timeSeriesData.length ? (
+          <NoDataAlert {...noDataAlertProps} />
```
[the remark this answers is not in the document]
Clebal: NoDataAlert shouldn't be used in UI, I think Widget is the place. But I don't want to modify CategoryWidgetUI in this PR, that's why I just generalise that part of the code in a component, and then exported it. Also, I think it's useful to export NoDataAlert because it can be used for users.

packages/react-ui/src/widgets/time-series-widget-ui/TimeSeriesWidgetUI.js
Clebal: Good question. We need to use context inside TimeSeriesWidgetUI and useTimeSeriesInteractivity. If we don't use a context, we should have to bring every prop through TimeSeriesChart, something that I don't like. But it isn't my unique motivation for creating a context. I like TimeSeriesContext because there is where all the stuff about managing the widget state and it makes that logic transparent to whoever want to use it. Also, because it's concentrate in one place, an usere can create a context that replaces that one and might extends it.

The first approach didn't have that context, but when I saw all that logic inside the TimeSeriesWidgetUI and all that props inside the TimeSeriesChart... it was totally a mess. I preferred the context way, much cleaner.

packages/react-ui/src/widgets/time-series-widget-ui/hooks/TimeSeriesContext.js
Clebal: Yes, it's correct. Array of dependencies is always a new object, but it isn't what useEffect checks, it checks every element inside that array of dependencies. That's why I spread the timeframe array, because I don't want to check new reference by timeframe, I want to check inner values because the timeframe array can change but its inner values no.

I don't want to check this: [[1,1]], I want to check: [1,1].

I think it's a minor improvement that might not be absolutely necessary, if it's annoying for you, I can delete it.

packages/react-widgets/src/widgets/TimeSeriesWidget.js
Clebal: Check defaultProps.

packages/react-widgets/src/widgets/TimeSeriesWidget.js
Clebal: NoDataAlert is a new component, I think we should bring this improvement to other widgets.

packages/react-ui/src/index.js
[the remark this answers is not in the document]
padawannn: As we have spoken, if we offer this functionality we need to maintain and document it so we remove it for now

packages/react-widgets/src/widgets/TimeSeriesWidget.js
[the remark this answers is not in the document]
padawannn: In that case, we should move NoDataAlert to react-widgets module y modify the CategoryWidgetUI and CategoryWidget to keep the coherence of the modules

packages/react-widgets/src/widgets/TimeSeriesWidget.js
padawannn: It's a bit strange that this widget can offer this possibility and the others not. @borja-munoz @Clebal I think that you should decide if add it to other widgets or remove it from this widget for now

packages/react-widgets/src/widgets/TimeSeriesWidget.js
padawannn: You're right

packages/react-widgets/src/widgets/TimeSeriesWidget.js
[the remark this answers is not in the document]
Clebal: Should I open a PR with this? I would bring this feature to every widget.

packages/react-widgets/src/widgets/TimeSeriesWidget.js
Clebal: CategoryWidgetUI had it, others didn't.

borja-munoz: It's a matter of time. If you think you can safely add it to the other widgets and we will still meet the deadline for the time series widget, go ahead. If you are not sure, it is better to remove it from this widget for now.

Clebal: Challenge accepted.

packages/react-widgets/src/widgets/TimeSeriesWidget.js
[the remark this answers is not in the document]
padawannn: Depends on https://github.com/CartoDB/carto-react/pull/168#discussion_r719470231

packages/react-ui/src/widgets/TimeSeriesWidgetUI/hooks/useTimeSeriesInteractivity.js
```diff
@@ -0,0 +1,269 @@
...
+        }
+      },
+    [theme, timeframe]
+  );
+
+  return { timelineOptions, timeframeOptions };
+}
+
+// Aux
+function addEvent(zr, eventKey, event) {
```
VictorVelarde, line 243: addEventWithCleanUp ?

packages/react-widgets/src/widgets/TimeSeriesWidget.js
VictorVelarde: What do you mean with this 🔝  @Clebal ? Will you add noDataAlertProps here to other widgets?

packages/react-ui/src/widgets/TimeSeriesWidgetUI/hooks/useTimeSeriesInteractivity.js
Clebal: Good one!

packages/react-widgets/src/widgets/TimeSeriesWidget.js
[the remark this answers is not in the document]
VictorVelarde: Not applicable anymore

packages/react-ui/src/utils/NoDataAlert.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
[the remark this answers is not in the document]
VictorVelarde: Still a thing to improve @Clebal

packages/react-widgets/src/widgets/TimeSeriesWidget.js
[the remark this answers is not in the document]
VictorVelarde: Yes please, open that PR so it lands **before** this one, and then adapt this one to it

packages/react-core/src/index.d.ts
```diff
@@ -5,6 +5,8 @@ export {
   REQUEST_GET_MAX_URL_LENGTH
 } from './utils/requestsUtils';
 
+export { getMonday } from './utils/dateUtils';
```
VictorVelarde, line 8: if getMonday is public, then we should add a clearer name to its param: so from `export function getMonday(d) {` --> `export function getMonday(date) {`

packages/react-core/src/operations/groupByDate.d.ts
```diff
@@ -0,0 +1,9 @@
+import { GroupByFeature } from '../types'
+import { GroupDateTypes } from './GroupDateTypes';
+
+export function groupValuesByDateColumn(
+  data: [], // TODO: geojson types
```
VictorVelarde, line 5: remove TODO comment

packages/react-ui/src/widgets/CategoryWidgetUI.js
```diff
@@ -510,11 +511,7 @@ function CategoryWidgetUI(props) {
                 />
               ))
             ) : data.length === 0 && !isLoading ? (
-              <Alert severity='warning'>
-                <AlertTitle>No data available</AlertTitle>
-                There are no results for the combination of filters applied to your data.
-                Try tweaking your filters, or zoom and pan the map to adjust the Map View.
-              </Alert>
+              <NoDataAlert />
```
VictorVelarde, line 514: If you create another PR for this mini-feature, then this should be included there

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -0,0 +1,308 @@
...
+    delete filters[column];
+    return filters;
+  });
+
+  const isSourceReady = !!viewportFeaturesReady[dataSource];
+
+  useEffect(() => {
+    if (isSourceReady) {
+      setIsLoading(true);
+      const _filters = getApplicableFilters(filters, id);
```
VictorVelarde, line 79: Ey @Clebal  you should use here new `useSourceFilters` right?

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -0,0 +1,308 @@
...
+      <WrapperWidgetUI
+        title={title}
+        isLoading={isLoading}
+        {...wrapperProps}
+        actions={[
+          ...(wrapperProps?.actions || []),
+          ...(stepSizeOptions?.length
+            ? [
+                {
+                  id: 'a0',
```
VictorVelarde, line 185: why 'a0' here? is it relevant? is it used somehow outside? Thinking if this should be an exported CONST or not

Clebal: Hmm I copied that from WrapperWidgetUI example. What should it be?

```
{
                  id: 'bucketSize',
                  name: 'Bucket size',
                  icon: <StepSizeIcon />,
                  action: handleOpenBucketSizeMenu
}
```

Is it ok?

packages/react-widgets/src/widgets/TimeSeriesWidget.js
Clebal: Yep! Just updated

packages/react-core/src/operations/groupByDate.d.ts
Clebal: image
Everywhere?

packages/react-core/src/index.d.ts
Clebal: Totally agree. Updated.

packages/react-ui/src/widgets/CategoryWidgetUI.js
Clebal: Yes.

packages/react-widgets/src/widgets/TimeSeriesWidget.js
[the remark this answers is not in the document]
VictorVelarde: Done

packages/react-widgets/src/widgets/TimeSeriesWidget.js
VictorVelarde: Done in another PR https://github.com/CartoDB/carto-react/pull/188

packages/react-core/src/operations/groupByDate.d.ts
VictorVelarde: 👍🏻  I did it

VictorVelarde:
I left current const SPEED_FACTORS = [0.5, 1, 2, 3]; 
To be determined if 3 is included or not before merging.

VictorVelarde:
I've reviewed latest changes and they look good to me @Clebal  🚀 . 

It looks like just storybook & reference review are pending @borja-munoz. We can talk on Monday about it

packages/react-core/src/operations/groupByDate.js
```diff
@@ -3,9 +3,27 @@ import { aggregationFunctions } from './aggregation/values';
...
+  // [GroupDateTypes.MINUTES]: (date) =>
+  //   new Date(
+  //     Date.UTC(
+  //       date.getFullYear(),
+  //       date.getMonth(),
+  //       date.getDate(),
+  //       date.getHours(),
+  //       date.getMinutes()
+  //     )
+  //   )
```
VictorVelarde, line 26: Is this for documentation or dead code @Clebal ?

packages/react-ui/src/widgets/TimeSeriesWidgetUI/TimeSeriesWidgetUI.js
```diff
@@ -15,18 +17,26 @@ import { PropTypes } from 'prop-types';
...
+  [GroupDateTypes.MONTHS]: monthsCurrentDateRange,
+  [GroupDateTypes.WEEKS]: daysCurrentDateRange,
+  [GroupDateTypes.DAYS]: daysCurrentDateRange
 };
 
-const TIMEFRAME_STEP_BY_STEP_SIZE = {
-  [GroupDateTypes.DAYS]: 60 * 60 * 24,
-  [GroupDateTypes.WEEKS]: 60 * 60 * 24
+const TIME_WINDOW_STEP_BY_STEP_SIZE = {
+  [GroupDateTypes.YEARS]: 60 * 60 * 24 * 7,
```
VictorVelarde, line 34: YEARS look related to 1 WEEK, @Clebal are these values correct?

packages/react-ui/src/widgets/TimeSeriesWidgetUI/TimeSeriesWidgetUI.js
```diff
@@ -345,6 +370,18 @@ function weeksCurrentDateRange(date) {
   return `Week of ${getMonday(date).toLocaleDateString()}`;
```
VictorVelarde, line 370: It might be interesting to add (Monday) to the copy, as in US they also represent weeks starting from Sunday... what do you think @borja-munoz ?

packages/react-ui/src/widgets/TimeSeriesWidgetUI/TimeSeriesWidgetUI.js
VictorVelarde: In other piece of the code you use: [GroupDateTypes.YEARS]: 60 * 60 * 24 * 7 * 30 * 12 so...

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -21,11 +21,47 @@ import { PropTypes } from 'prop-types';
...
 
-const BUCKET_SIZE_RANGE_MAPPING = {
+const STEP_SIZE_RANGE_MAPPING = {
+  [GroupDateTypes.YEARS]: 60 * 60 * 24 * 7 * 30 * 12,
+  [GroupDateTypes.MONTHS]: 60 * 60 * 24 * 7 * 30,
   [GroupDateTypes.WEEKS]: 60 * 60 * 24 * 7,
-  [GroupDateTypes.DAYS]: 60 * 60 * 24
+  [GroupDateTypes.DAYS]: 60 * 60 * 24,
+  // [GroupDateTypes.HOURS]: 60 * 60,
+  // [GroupDateTypes.MINUTES]: 60
```
VictorVelarde: Please remove commented, dead code

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -264,10 +303,10 @@ TimeSeriesWidget.propTypes = {
   isPaused: PropTypes.bool,
   onPause: PropTypes.func,
   onStop: PropTypes.func,
-  timelinePosition: PropTypes.number,
+  // timelinePosition: PropTypes.number,
```
VictorVelarde, line 311: better remove than comment

packages/react-core/src/operations/GroupDateTypes.js
```diff
@@ -17,11 +17,8 @@ export const GroupDateTypes = Object.freeze({
   DAYS: 'days',
 
   /** Hours */
-  HOURS: 'hours',
+  // HOURS: 'hours',
```
VictorVelarde, line 20: remove commented / unused, please

packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -172,14 +172,19 @@ function TimeSeriesWidget({
...
-      handleTimeWindowUpdate([
-        moment,
-        moment + STEP_SIZE_RANGE_MAPPING[selectedStepSize] * 1000
-      ]);
+      dispatch(
+        addFilter({
+          id: dataSource,
+          column,
+          type: FilterTypes.BETWEEN,
+          values: [moment, moment + STEP_SIZE_RANGE_MAPPING[selectedStepSize] * 1000],
```
VictorVelarde, line 180: Why * 1000 here? is it milliseconds? It's not very evident... maybe you could add a const with a proper name here for the whole expression "const XXX = STEP_SIZE_RANGE_MAPPING[selectedStepSize] * 1000",

PR #171, opened 2021-09-08
Proportion legend middle values are wrong

aaranadev:
Fix LegendProportional when the values are not a number

Story details: https://app.clubhouse.io/cartoteam/story/178008


packages/react-ui/src/widgets/legend/LegendProportion.js
```diff
@@ -43,8 +61,8 @@ function calculateSteps([min, max]) {
 export default function LegendProportion({ legend }) {
   const classes = useStyles();
 
-  const [min, max] = calculateRange(legend);
-  const [step1, step2] = calculateSteps([min, max]);
+  const { min, max, error } = calculateRange(legend);
+  const [step1, step2] = calculateSteps(min, max);
```
borja-munoz, line 65: Not a big deal but we just need to calculateSteps if error is false

PR #172, opened 2021-09-22
Fix Storybook showing args instead of properties

Clebal:
Story details: https://app.shortcut.com/cartoteam/story/147101

Using docs.code.type dynamic and jsx options it's possible to get dynamic source code.

Also, I've made a little clean to avoid undefined values and inconsistencies in stories.


VictorVelarde:
@borja-munoz putting this on-hold in the meantime

PR #184, opened 2021-09-28
GENERAL / Bad management of loading state

Clebal:
Story details: 
- https://app.shortcut.com/cartoteam/story/182598 and
- https://app.shortcut.com/cartoteam/story/182587/general-changes-in-a-source-trigger-un-affected-widget-calculations


packages/react-widgets/src/widgets/CategoryWidget.js
```diff
@@ -38,21 +37,24 @@ function CategoryWidget(props) {
     onError,
     wrapperProps
   } = props;
+  const dispatch = useDispatch();
+
+  const isSourceReady = useSelector(
```
VictorVelarde, line 42: better rename from isSourceReady to `vieportFeaturesForSourceReady`

packages/react-widgets/src/widgets/PieWidget.js
```diff
@@ -39,33 +38,29 @@ function PieWidget({
...
+
   const [categoryData, setCategoryData] = useState([]);
   const [selectedCategories, setSelectedCategories] = useState([]);
+  const [isLoading, setIsLoading] = useState(true);
 
-  const dispatch = useDispatch();
-  const source = useSelector((state) => selectSourceById(state, dataSource) || {});
-  const viewportFeaturesReady = useSelector((state) => state.carto.viewportFeaturesReady);
-
-  const widgetsLoadingState = useSelector((state) => state.carto.widgetsLoadingState);
```
VictorVelarde, line 49: So you're not using widgetsLoadingState anymore right?. In that case we should consider removing it from cartoSlice, the key and related methods

packages/react-widgets/src/widgets/CategoryWidget.js
Clebal: I thought about it, but it's too long. I think it's understandable by the context that when a source is ready (in the context of widget's data fetching) is when viewport features are ready.

packages/react-widgets/src/widgets/PieWidget.js
```diff
@@ -77,28 +72,17 @@ function PieWidget({
           setIsLoading(false);
           if (onError) onError(error);
         });
-    } else {
-      setCategoryData([]);
     }
-
-    return function cleanup() {
-      abortController.abort();
```
VictorVelarde, line 85: ok, yeah, we don't have server-side calcs anymore, 👍🏻

packages/react-widgets/src/widgets/PieWidget.js
Clebal: Yes, I forgot that.

VictorVelarde:
Please, double-check that my merge with master in useViewportFeatures conflict is correct @Clebal

packages/react-widgets/src/widgets/CategoryWidget.js
VictorVelarde: fair enough!

VictorVelarde:
I'll also add some details about the PR scope, as it also includes https://app.shortcut.com/cartoteam/story/182587/general-changes-in-a-source-trigger-un-affected-widget-calculations

packages/react-widgets/src/widgets/CategoryWidget.js
```diff
@@ -113,12 +111,12 @@ function CategoryWidget(props) {
...
 
   return (
-    <WrapperWidgetUI title={title} isLoading={widgetsLoadingState[id]} {...wrapperProps}>
+    <WrapperWidgetUI title={title} isLoading={isLoading} {...wrapperProps}>
       <CategoryWidgetUI
         data={categoryData}
         formatter={formatter}
         labels={labels}
-        isLoading={widgetsLoadingState[id]}
+        isLoading={isLoading}
```
VictorVelarde, line 119: Why is it here the isLoading preserved (vs other widgets)? Does it really have an effect on the widget or could we just leave the generic isLoading for WrapperWidgetUI ?

Clebal: Yes, CategoryWidgetUI uses isLoading prop for its logic. It was there before my changes.

VictorVelarde (commented):
Overall it looks good to me. Simplification makes sense, removing the old '-data' check (probably coming from server-side calculations). 

But changing the 'loadingState' for widgets results in a worse experience. This is a test with current approach (https://sample-app-react.carto.com/) vs the new loading state at widget level:


https://user-images.githubusercontent.com/458196/135323497-58752afd-ed6f-4581-b420-f372e2d7d722.mov



---

https://user-images.githubusercontent.com/458196/135323434-736b8fcc-abb4-4b93-a538-d301b13e6c14.mov


---

As you can see, the current triggers before, and it's more 'progressive' or 'soft', vs current one.
Can you think of a way @Clebal to improve the ux here, with your current approach (that I'm pretty sure is more efficient)?

Clebal:
@VictorVelarde 
It happens because in my code, the dispatch that makes the widgets start loading is inside the debounced function, and in the current code it's inside the useEffect.

I would move the dispatch to set to false the viewportFeaturesReady state inside the useEffect block. We're alerting widgets with a delay that data they are working with isn't ready, I don't think it's ok.

packages/react-widgets/src/widgets/FormulaWidget.js
```diff
@@ -60,19 +61,17 @@ function FormulaWidget(props) {
       setFormulaData(null);
```
padawannn, line 61: I think that we can remove `setFormulaData(null);` if the source is not ready

packages/react-widgets/src/widgets/FormulaWidget.js
```diff
@@ -30,17 +29,19 @@ function FormulaWidget(props) {
     onError,
     wrapperProps
   } = props;
+  const isSourceReady = useSelector(
```
padawannn, line 32: We repeat this code in all widgets. Could we have a redux selector?

packages/react-widgets/src/widgets/ScatterPlotWidget.js
```diff
@@ -66,10 +67,19 @@ function ScatterPlotWidget(props) {
     } else {
       setScatterData([]);
```
padawannn, line 68: I think that we can remove this `else` condition

padawannn (changes requested):
I general LGFM. Some minor changes

packages/react-widgets/src/widgets/FormulaWidget.js
Clebal: What do you think @VictorVelarde? I am in favor of adding a selector in redux, but I saw it isn't common in c4r, then I didn't do it.

padawannn: Just below you have a selector XD `selectSourceById`

Clebal: `selectViewportFeaturesReadyBySourceId` is a good name? Maybe too long

padawannn: Verbose names is a good practice, another more descriptive could be `selectIsViewportFeaturesReadyForSource`

Clebal: Added.

padawannn (approved):
LGFM

packages/react-widgets/src/widgets/FormulaWidget.js
VictorVelarde: I like Javi's naming proposal

VictorVelarde (approved):
LGTM. I've tested it within sample-app and the loading 'experience' has improved with your changes 👍🏻 . Thx for your work here @Clebal

PR #185, opened 2021-09-29
GENERAL / Filters from a widget launches unrequired calculations on itself

Clebal:
Story details: https://app.shortcut.com/cartoteam/story/182594


packages/react-widgets/src/utils/useCustomCompareMemo.js
```diff
@@ -0,0 +1,11 @@
+import { useRef } from 'react';
+
+export default function useCustomCompareMemo(value, isEqual) {
```
padawannn, line 3: According to the rest of the project, hooks should be in a `hooks` folder so you should rename `utils` by `hooks`

padawannn (changes requested):
I like the new hooks, only a minor request

packages/react-widgets/src/utils/useCustomCompareMemo.js
Clebal: Done.

padawannn (approved):

VictorVelarde:
Do you think that it could be useful for custom widgets, maybe?

Clebal:
@VictorVelarde Yes, I thought it would be an useful hook in PS. It's common to create custom widgets.

VictorVelarde (approved):
Nice work! 🚀

PR #188, opened 2021-10-01
Add NoDataAlert widget

Clebal:
Story details: https://app.shortcut.com/cartoteam/story/183641

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
+- Add NoDataAlert widget [#188](https://github.com/CartoDB/carto-react/pull/188)
 - Update echarts to v5 [#167](https://github.com/CartoDB/carto-react/pull/167)
 - Fix unnecessary widget calculations triggered by widget itself [#185](https://github.com/CartoDB/carto-react/pull/185)
 - Add new `useSourceFilters` hook for better custom widgets dev [#185](https://github.com/CartoDB/carto-react/pull/185)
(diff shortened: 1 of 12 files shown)


VictorVelarde:
I've added it also to FormulaWidget, for consistency (and because default 0.00 value for empty might be wrong / misleading under certain operations). I also reduced a bit the IMO very long text (@borja-munoz can maybe give us feedback on that during QA)

image

VictorVelarde (approved):
LGTM, so 🚀

Clebal:
Sorry @VictorVelarde, I don't agree. From my experience in PS projects, when the count of elements is 0, the client wants to see 0, no "No data". That's why I didn't added there.

Another example, if the sum of revenue of a certain store is 0, it isn't the same as no data.

VictorVelarde:
I understand that displaying 0.0 might be find assuming Count or Sum, but there are other operations that can be applied to Formula, where 0 is not, IMO, fine. Having said that I have no issues in reverting that commit, let's just ping @borja-munoz for confirmation on this before merging it

VictorVelarde:
I'm gonna do one thing, I'm gonna revert my commit, so to merge this as Sergio set it, and if Borja wants to include my FormulaWidget change, we can do another PR with cherry-pick later

borja-munoz:
I think for COUNT operations makes sense to show 0 if there is no data in the current viewport. For the other aggregation operations (SUM, AVG, MIN, MAX), I think it is better to show the no data message.

Sorry for introducing a third option. If there is no time now to implement it, I'd keep Sergio's approach for now, but revisit this in the future.

PR #193, opened 2021-10-14
Bump deck.gl to 8.6.0

felixpalmer:


VictorVelarde:
@felixpalmer I would please ask you to have the corresponding PR for the upgrade in the carto-react-template project as well, so some QA can be done from the user's perspective, loading this linked lib version

PR #223, opened 2021-11-05
Improve Widgets calculations sync with tiles

Clebal:
Story details: https://app.shortcut.com/cartoteam/story/148156

Approach:

I store in a ref the timeoutId of the debounced function. Everytime that an update in tiles or viewport can happens, the debounce is cleared.
Using fetch, we know that a tile is being fetched. Knowing that, we can stop the debounce because we know in advance that in a few seconds a tile will be loaded, avoiding unnecessary calculations.

### DIFF
--- a/packages/react-api/src/hooks/useCartoLayerProps.js
+++ b/packages/react-api/src/hooks/useCartoLayerProps.js
@@ -7,12 +7,12 @@ export default function useCartoLayerProps({
   source,
   uniqueIdProperty,
   viewportFeatures = true,
-  debounceTimeOut = 500
+  viewporFeaturesDebounceTimeout = 500
 }) {
   const [onViewportLoad, onDataLoad, fetch] = useViewportFeatures(
     source,
     uniqueIdProperty,
-    debounceTimeOut
+    viewporFeaturesDebounceTimeout
   );
 
   let props = {};
--- a/packages/react-api/src/hooks/useCartoLayerProps.js
+++ b/packages/react-api/src/hooks/useCartoLayerProps.js
@@ -6,9 +6,14 @@ import { MAP_TYPES, API_VERSIONS } from '@deck.gl/carto';
 export default function useCartoLayerProps({
   source,
   uniqueIdProperty,
-  viewportFeatures = true
+  viewportFeatures = true,
+  debounceTimeOut = 500
 }) {
-  const [onViewportLoad, onDataLoad] = useViewportFeatures(source, uniqueIdProperty);
+  const [onViewportLoad, onDataLoad, fetch] = useViewportFeatures(
+    source,
+    uniqueIdProperty,
+    debounceTimeOut
+  );
 
   let props = {};
 
@@ -18,7 +23,8 @@ export default function useCartoLayerProps({
   ) {
     props = {
       binary: true,
-      onViewportLoad: viewportFeatures ? onViewportLoad : null
+      onViewportLoad: viewportFeatures ? onViewportLoad : null,
+      fetch: viewportFeatures ? fetch : null
     };
   } else if (source?.type === MAP_TYPES.QUERY || source?.type === MAP_TYPES.TABLE) {
     props = {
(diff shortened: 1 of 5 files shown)


packages/react-api/src/hooks/useViewportFeatures.js
```diff
@@ -22,6 +23,14 @@ export default function useViewportFeatures(
   const viewport = useSelector((state) => state.carto.viewport);
   const [tiles, setTiles] = useState([]);
   const [isGeoJSONLoaded, setGeoJSONLoaded] = useState(false);
+  const debounceId = useRef(0);
```
VictorVelarde, line 26: why 0 and not null?

Clebal: Nothing special. I just preferred 0 here. Null would be the same. I can change it if you want.

packages/react-api/src/hooks/useViewportFeatures.js
```diff
@@ -90,8 +101,14 @@ export default function useViewportFeatures(
 
   useEffect(() => {
     if (sourceId && tiles.length && (!isSourceV3 || isSourceTileset)) {
+      clearDebounce();
```
VictorVelarde, line 104: You apply the clearDebounce just for tiles, what would happen to geojson? Doesn't it make sense to apply same approach?

Clebal: No, because geojson is only loaded once and only have one way to be updated: viewport changes. For when viewport changes fastly, we have debounced the function for filtering by viewport.
When we have tiles, the viewport features can be updated when viewport changes and when new tiles are loaded. That's why we need clearDebounce only in tiles, to stop debounce if we know that a change in the feature is coming.

packages/react-api/src/hooks/useCartoLayerProps.js
```diff
@@ -6,9 +6,14 @@ import { MAP_TYPES, API_VERSIONS } from '@deck.gl/carto';
...
   uniqueIdProperty,
-  viewportFeatures = true
+  viewportFeatures = true,
+  debounceTimeOut = 500
 }) {
-  const [onViewportLoad, onDataLoad] = useViewportFeatures(source, uniqueIdProperty);
+  const [onViewportLoad, onDataLoad, fetch] = useViewportFeatures(
+    source,
+    uniqueIdProperty,
+    debounceTimeOut
```
VictorVelarde, line 15: 👍🏻

packages/react-api/src/hooks/useCartoLayerProps.js
```diff
@@ -6,9 +6,14 @@ import { MAP_TYPES, API_VERSIONS } from '@deck.gl/carto';
 export default function useCartoLayerProps({
   source,
   uniqueIdProperty,
-  viewportFeatures = true
+  viewportFeatures = true,
+  debounceTimeOut = 500
```
VictorVelarde, line 10: We should probably be a bit more explicit on this new public parameter in useCartoLayerProps (it looks like a nice name for useViewportFeatures though), to indicate where it applies. How about viewporFeaturesDebounceTimeout ?

packages/react-api/src/hooks/useViewportFeatures.js
VictorVelarde: That looks reasonable.

VictorVelarde:
@Clebal CR looks fine to me. I would like to test it locally, and then see if we can prepare a patch.
We would need also a little PR for updating documentation, adding info about this new parameter. Would you please create that in documentation repo?

packages/react-api/src/hooks/useCartoLayerProps.js
Clebal: Totally agree. I'll change it.

Clebal:
@VictorVelarde sure. I'll do it ASAP.

padawannn (approved):
Looks well. I like the approach

PR #224, opened 2021-11-15
Manage widgets selections through filters so they are more flexible

zbigg:
Widgets that affect datasource filters, use filters state in redux, to derive its own "filter/selection", instead of local state.

- [x] CategoryWidget
- [x] HistogramWidget
- [x] PieWidget
- [x] ~TimeSeriesWidget~ (not in scope)

Story details: https://app.shortcut.com/cartoteam/story/192425

### DIFF
--- a/packages/react-widgets/src/hooks/useWidgetFilterValues.js
+++ b/packages/react-widgets/src/hooks/useWidgetFilterValues.js
@@ -1,0 +1,26 @@
+import { selectSourceById } from '@carto/react-redux/';
+import { useSelector } from 'react-redux';
+import useCustomCompareMemo from './useCustomCompareMemo';
+import { dequal as deepEqual } from 'dequal';
+
+/**
+ * Obtain current widgets's filter values.
+ *
+ * @param  {object} props
+ * @param  {string} props.dataSource - ID of the source to get the filters from.
+ * @param  {string} props.id - ID of the widget that apply the filter you want to exclude.
+ * @param  {string} props.column - name of column of this widget.
+ * @param  {string} props.type - type of filter
+ */
+export function useWidgetFilterValues({ dataSource, id, column, type }) {
+  const columnFilters = useSelector((state) => {
+    // TODO: id ?
+    const dataSourceState = selectSourceById(state, dataSource);
+    const filterTypes = dataSourceState?.filters ? dataSourceState?.filters[column] : {};
+
+    console.log('useWidgetOwnFilters', dataSource, id, filterTypes);
+    return filterTypes ? filterTypes[type]?.values : null;
+  });
+
+  return useCustomCompareMemo(columnFilters, deepEqual);
+}
--- a/packages/react-widgets/src/hooks/useWidgetFilterValues.js
+++ b/packages/react-widgets/src/hooks/useWidgetFilterValues.js
@@ -4,22 +4,23 @@ import useCustomCompareMemo from './useCustomCompareMemo';
 import { dequal as deepEqual } from 'dequal';
 
 /**
- * Obtain current widgets's filter values.
+ * Obtain widget's filter values.
  *
  * @param  {object} props
  * @param  {string} props.dataSource - ID of the source to get the filters from.
- * @param  {string} props.id - ID of the widget that apply the filter you want to exclude.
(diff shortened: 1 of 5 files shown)


zbigg:
At this point i would just call review for CategoryWidget, HistogramWidget,  PieWidget and discuss TimeSeriesWidget separately as it has different model.

TimeSeriesWidget has two separate filte states, i.e "point" aka timelinePosition and "range" timeWindow and they are use same model in filter space, so it would require heuristics to "recreate" timeWindow/timelinePosition setting.

VictorVelarde (commented):
@zbigg The changes look good to me. I've updated with master, as I merged another PR from Sergio (#223). I would just miss a bit of testing if possible

padawannn (commented):
LGFM

zbigg:
@VictorVelarde We agreed that test shall be part of followup changes. Any other obstacles to merging it?

VictorVelarde (approved):
Not at all, go ahead

packages/react-widgets/src/hooks/useWidgetFilterValues.js
```diff
@@ -0,0 +1,27 @@
...
+    const dataSourceState = selectSourceById(state, dataSource);
+
+    const filter = dataSourceState?.filters?.[column]?.[type];
+    if (!filter || filter.owner !== id) {
+      return null;
+    }
+    return filter.values;
+  });
+
+  return useCustomCompareMemo(columnFilters, deepEqual);
```
Clebal, line 26: I think that useCustomCompareMemo isn't needed here because filter values won't have a different reference until filter is updated.

packages/react-widgets/src/hooks/useWidgetFilterValues.js
```diff
@@ -0,0 +1,27 @@
...
+export function useWidgetFilterValues({ dataSource, id, column, type }) {
+  const columnFilters = useSelector((state) => {
+    const dataSourceState = selectSourceById(state, dataSource);
+
+    const filter = dataSourceState?.filters?.[column]?.[type];
+    if (!filter || filter.owner !== id) {
+      return null;
+    }
+    return filter.values;
+  });
```
Clebal, line 24: useSelector re-rendered each time any part of the redux store change. Also, I think it isn't recommended to use so much logic inside of an useSelector.
I would do something similar to this (not tested):

```suggestion
  const { filters } = useSelector((state) => selectSourceById(state, dataSource) || {});
  
  return useMemo(() => {
    const filter = filters?.[column]?.[type];
    if (!filter || filter.owner !== id) {
      return null;
    }
    return filter.values;
  }, [filters, column, type, id]);
```

Clebal (changes requested):

packages/react-widgets/src/hooks/useWidgetFilterValues.js
VictorVelarde: Why not using a new createSelector for this, instead of reading + filtering an existing one?
From my understanding, using a createSelector should be enough to create a memoized selection, but we should check.

Pretty relevant links here: https://redux-toolkit.js.org/api/createSelector

Clebal: Yes, createSelector can be an option but from my understanding of createSelector it's for complex data structures and calculations using different parts of redux state.
I don't think it's the case where we just take filters from redux and the operations made (accessing a key-value object) have a insigficant impact. Even we can discuss if useMemo adds enough perf improvements that worth memoization cost haha.
I think useMemo is good enough.

VictorVelarde: It's true that createSelector allows combining several slices for a query, but they are also used for basic / repeated stuff like filtering: https://github.com/reduxjs/reselect. IMO main point is 'encapsulation', not perf.

In any case, let's leave @zbigg pick an option here

zbigg: I think that @Clebal comments are ok, i've applied them in e391abd.

Clebal (approved):
@zbigg LGTM 🚀

PR #225, opened 2021-11-17
Return raw feature data from workers

Clebal:
Story details: https://app.shortcut.com/cartoteam/story/192546

### Possible questions

- **Why have you modified [buildFeatureFilter](https://github.com/CartoDB/carto-react/pull/225/files#diff-d7b59ea5d65883e4b880e375e86b9cb4f78a24efacc87b3a82822b63bb3556a8R59) fn?**
It's a minor improvement to avoid calling Object.keys for each feature

- **Why that new [applyFilters](https://github.com/CartoDB/carto-react/pull/225/files#diff-d7b59ea5d65883e4b880e375e86b9cb4f78a24efacc87b3a82822b63bb3556a8R75) fn in core?**
I think this can be used in advanced use cases.

- **Do you think that `sorting` might be part of core?**
I don't think so because it's just used in workers, and I don't think it can be used in other parts of C4R or even that C4R have to provide any kind of sorting out of the main think. Idk. Let me know what do you think.

- **What's sortBy structure?**
This can be checked [here](https://github.com/CartoDB/carto-react/pull/225/files#diff-bbaa79a822be2eb3a7132d51faa3ad67db3ff19c3faeff68cf2ed1d15be49b5fR7).
sortBy accepts:
     1. An string (a column) `sortBy: 'column_1'`
     2. Array of strings (multiple columns) `sortBy: ['column_1', 'column_2']`
     3. An array where each element follows [thenBy](https://github.com/Teun/thenBy.js) structure. Passing each child array as arguments to firstBy/thenBy functions. Thank to this we are able to [specify direction for each column](https://github.com/Teun/thenBy.js#sort-descending) or [ignoring cases](https://github.com/Teun/thenBy.js#case-insensitive-sorting) or any other feature of thenBy.

Example of the third option:

```
sortBy: [
    ['column_1', 'desc'],
    ['column_2', { direction: 'asc', ignoreCase:true }]
]
```

- **I saw that thenBy accepts function? Does sortBy accepts function?**
Nope, functions cannot be cloned to workers.

- **If I indicate direction using thenBy structure and I also indicate `sortByDirection`. Which is applied?**
thenBy structure has priority since it's more specific. If sorting is applied by multiple column and some doesn't indicate direction, `sortByDirection` is used (which value by default is 'asc').


packages/react-core/src/filters/Filter.js
```diff
@@ -57,15 +57,23 @@ function passesFilter(columns, filters, feature) {
...
-    const columns = Object.keys(filters);
     const f = feature.properties || feature;
     const featurePassesFilter = passesFilter(columns, filters, f);
 
     return type === 'number' ? Number(featurePassesFilter) : featurePassesFilter;
   };
 }
+
+// Apply certain filters to a collection of features
+export function applyFilters(features, filters) {
```
Josmorsot, line 75: I think that it's not necessary, you should check on "parent" component if need to apply filters or not. Here you're calling to applyFilters but it's not applying anything necessarily

packages/react-workers/src/workers/viewportFeatures.worker.js
```diff
@@ -144,8 +148,34 @@ function getTimeSeries({ filters, column, stepSize, operation, operationColumn }
...
+    } else {
+      data = features;
+    }
+  }
+
+  postMessage({ result: { data, currentPage: page, pages: numberPages } });
+}
+
+function getFilteredFeatures(filters = {}) {
+  return _applyFilters(currentViewportFeatures, filters);
```
Josmorsot, line 180: I would check here if features need filters or not, Is there any reason why you have move it to `filters.js`?

packages/react-workers/src/workers/viewportFeatures.worker.js
```diff
@@ -144,8 +148,34 @@ function getTimeSeries({ filters, column, stepSize, operation, operationColumn }
...
+  let numberPages = 0;
+
+  if (currentViewportFeatures) {
+    const features = applySorting(getFilteredFeatures(filters), {
+      sortBy,
+      sortByDirection
+    });
+
+    if (limit) {
+      numberPages = Math.ceil(features.length / limit);
```
Josmorsot, line 169: You should create an `applyLimit` function and simply call:

```js
data = features
if (limit) data = applyLimit(features, limit)

postMessage(......)
```

packages/react-workers/src/utils/sorting.js
```diff
@@ -0,0 +1,58 @@
...
+    return features;
+  }
+
+  const [firstSortOption, ...othersSortOptions] = normalizeSortByOptions({
+    sortBy,
+    sortByDirection
+  });
+
+  // Create sortFn
+  let sortFn = firstBy(...firstSortOption);
```
Josmorsot, line 25: Instead of add a comment, create a new `createSortFn`. It's just a suggestions, feel free to change or not :) 
Something like:
```js
function createSortFn({ sortBy, sortByDirection }) {
    const sortOptions = normalizeSortByOptions({
        sortBy,
        sortByDirection
     });
    return applySortOptions(sortOptions)
}
```

and call with 

```js
const sortFn = createSortFn({
    sortBy,
    sortByDirection
  });

```

Josmorsot (changes requested):
Nice work! Just minor suggestions :)

packages/react-core/src/filters/Filter.js
Josmorsot: > I think this can be used in advanced use cases.

I've just read it, I think that it could be added in the future if needed. But for now I'll try to move this logic to the parent component instead.

Clebal: Why in the future but not now? I think it's a wrapper to simplify applying C4R filters. I think that the only thing that we can change in the future is to remove the underscore sign and export it to every user. In the meanwhile, even with the underscore, I think it's something that can be useful in PS and we can avoid the `.filter` and the `buildFeatureFilter`, you just pass a feature collection and C4R filters and the magic is done. If, in the future, we want to change how buildFeatureFilter is made or called, I think it's better if we have this wrapper because applyFilters interface won't change.

packages/react-workers/src/workers/viewportFeatures.worker.js
Clebal: I think it's better to have that logic focused in one wrapper that can be used in many other parts that have the Object.keys(filters).length > 0 in every corner of the library or app.

packages/react-workers/src/workers/viewportFeatures.worker.js
Clebal: I like it. I'll change it. In the future, maybe you can add that suggestion like a commit suggestion to just commit the change.

packages/react-workers/src/utils/sorting.js
Clebal: I also like it!

packages/react-workers/src/workers/viewportFeatures.worker.js
Clebal: Updated

packages/react-workers/src/utils/sorting.js
Clebal: Updated

packages/react-core/src/filters/Filter.js
Josmorsot: I'm not talking about `applyFilters` function, I'm talking about checking if the features must be filtered or not. My suggestion was to move the condition to the component that is calling to `applyFilters`, anyways.... I think that you're right, you should check it always in this function.

LGTM

packages/react-workers/src/utils/sorting.js
```diff
@@ -16,21 +16,29 @@ export function applySorting(features, { sortBy = [], sortByDirection = 'asc' }
...
+
+// Aux
+function createSortFn({ sortBy, sortByDirection }) {
   const [firstSortOption, ...othersSortOptions] = normalizeSortByOptions({
     sortBy,
     sortByDirection
   });
 
-  // Create sortFn
   let sortFn = firstBy(...firstSortOption);
```
Josmorsot, line 34: I'd extract it to a new function too.: 
```js
function createSortFn({ sortBy, sortByDirection }) {
    const sortOptions = normalizeSortByOptions({
        sortBy,
        sortByDirection
     });
    return applySortOptions(sortOptions)
}
```

But LGTM

Josmorsot (approved):
LGTM, wait for double review before merge 🚀

Clebal:
@VictorVelarde @padawannn

packages/react-workers/__tests__/sorting.test.js
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    test('if sortBy is array of string', async () => {
+      expect(applySorting(features, { sortBy: ['column1', 'column2'] })).toEqual(
+        commonSortedFeatures
+      );
+    });
+
+    test('if sortBy is array of arrays', async () => {
+      expect(applySorting(features, { sortBy: [['column1'], ['column2']] })).toEqual(
```
VictorVelarde, line 44: Why is this array of arrays use case required? Could you please share an example or context on it?

packages/react-workers/__tests__/sorting.test.js
```diff
@@ -0,0 +1,79 @@
...
+  describe('should correctly understand sortOptions', () => {
+    test('if undefined', async () => {
+      expect(applySorting(features)).toEqual(features);
+    });
+
+    test('if sortBy is string', async () => {
+      expect(applySorting(features, { sortBy: 'column1' })).toEqual(commonSortedFeatures);
+    });
+
+    test('if sortBy is array of string', async () => {
```
VictorVelarde, line 37: I guess this is just using both columns, with default 'asc' order. If that's the case, then I would probably suggest to rename the test to something like...

```suggestion
    test('if sortBy uses 2 columns', async () => {
```

packages/react-workers/__tests__/sorting.test.js
Clebal: You might want to order ascendant by column1 and descendant by column 2. You can do that using array of arrays.

packages/react-workers/__tests__/sorting.test.js
Clebal: Fine!

packages/react-workers/__tests__/sorting.test.js
Clebal: In the test I just wanted to test the structure, if that works, the different order also works

packages/react-workers/__tests__/sorting.test.js
```diff
@@ -0,0 +1,79 @@
...
+      );
+
+      expect(
+        applySorting(features, { sortBy: [['column1']], sortByDirection: 'desc' })
+      ).toEqual(commonSortedFeatures2);
+    });
+
+    test('if sort direction is applied inside sortBy and sortByDirection is also used, sortBy has priority', async () => {
+      expect(
+        applySorting(features, { sortBy: [['column1', 'desc']], sortByDirection: 'asc' })
```
VictorVelarde, line 68: I see this is really flexible, with quite a lot of overloads / combinations to pass options... I'm just wondering if that could cause issues and should be just simplified, what do you think?

packages/react-workers/src/utils/sorting.js
```diff
@@ -0,0 +1,66 @@
...
+ * @param {string | string[] | object[]} [sortOptions.sortBy] - One or more columns to sort by
+ * @param {string} [sortOptions.sortByDirection] - Direction by the columns will be sorted
+ */
+export function applySorting(features, { sortBy = [], sortByDirection = 'asc' } = {}) {
+  const isValidSortBy =
+    (Array.isArray(sortBy) && sortBy.length) || // sortBy can be an array of columns
+    typeof sortBy === 'string'; // or just one column
+
+  if (!isValidSortBy) {
+    return features;
```
VictorVelarde, line 16: No error here? Shouldn't we just throw with a proper message?

packages/react-workers/src/utils/sorting.js
```diff
@@ -0,0 +1,66 @@
...
+// Aux
+function createSortFn({ sortBy, sortByDirection }) {
+  const [firstSortOption, ...othersSortOptions] = normalizeSortByOptions({
+    sortBy,
+    sortByDirection
+  });
+
+  let sortFn = firstBy(...firstSortOption);
+  for (let sortOptions of othersSortOptions) {
+    sortFn = sortFn.thenBy(...sortOptions);
```
VictorVelarde, line 36: Nice function building here

packages/react-workers/__tests__/sorting.test.js
Clebal: I think we shouldn't be afraid of that because there isn't many tricks that can be done, and all the possible tricks are resolved in one function (normaliseOptions if my memory doesn't lie), then if there's any bug, we know where we can find them, then I wouldn't be worried about that, I think it's pretty controlled.

packages/react-workers/src/utils/sorting.js
```diff
@@ -0,0 +1,66 @@
...
+        return [sortByEl, sortByDirection];
+      }
+
+      // sortBy is ['column', { ... }]
+      if (typeof sortByEl[1] === 'object') {
+        return [sortByEl[0], { direction: sortByDirection, ...sortByEl[1] }];
+      }
+    }
+    return sortByEl;
+  });
```
VictorVelarde, line 65: This is the part I was talking about before. Do we need all those variations regarding sortByEl structure?

packages/react-workers/src/workerMethods.js
```diff
@@ -4,6 +4,7 @@ export const Methods = Object.freeze({
   VIEWPORT_FEATURES_HISTOGRAM: 'viewportFeaturesHistogram',
   VIEWPORT_FEATURES_CATEGORY: 'viewportFeaturesCategory',
   VIEWPORT_FEATURES_SCATTERPLOT: 'viewportFeaturesScatterPlot',
+  VIEWPORT_FEATURES_RAW_FEATURES: 'viewportFeaturesRawFeatures',
```
VictorVelarde, line 7: Why not just VIEPORT_FEATURES_RAW, without adding again the word 'FEATURES' ?

VictorVelarde (approved):
Just small comments, but it looks like a nice approach to me. Remember to add a changelog and rebase master, please

packages/react-workers/src/utils/sorting.js
Clebal: Do you think we should throw an error or just show a console.error?

packages/react-workers/src/utils/sorting.js
Clebal: Thanks 😄

packages/react-workers/src/utils/sorting.js
Clebal: I think that if we want to offer a wide range of possibilities from the simplest to the most complex, the answer is yes. And I think that this kind of function that normalize options should be very common in C4R because these functions concentrate the complexity of making a functionality advanced and makes the rest of the code easy to read. Also, if the function is well documented (like I think this one is), this can be understood by anyone and won't be difficult to maintain. In the BarWidgetUI, I also use another normalize fn because it makes everything easier when you're dealing with different kind of inputs, one from the basic users (I just want to order by one column, I send you a string) and another for the most advanced users (I want to order by multiple columns and each one in a different way). We can offer all these variation just making this 20 lines fn without compromising the rest of the code because we're abstracting those line from that using this normalize fn.
I hope this helps to understand my vision.

VictorVelarde: If you really need that flexibility, then I'm ok with it. I just want to be sure it's needed

packages/react-workers/src/utils/sorting.js
VictorVelarde: Throw in my opinion. Passing an invalid sortby, without noticing, can lead to difficult bugs to find

Clebal: Added

VictorVelarde (approved):
Nice job. 🚀

PR #226, opened 2021-11-22
Add spatial filter

Clebal:
I've uploaded a branch in carto-react-template for testing purposes: `feature/spatial-filter`.

Nebula is enabled by default. Just click on the map.


borja-munoz:
Functionality looks great for masking but I'm not sure if we should call the filter "mask". Masking is an option directly related to visualization but the primary functionality we want to have is geospatial intersection. This is simply another filter that you add to the source. I like masking functionality but, in my opinion, this should be a separate thing.

packages/react-api/src/hooks/tileset/useSpatialFilterTileset.js
```diff
@@ -0,0 +1,137 @@
...
+      debounceIdSetSpatialFilterBuffersRef.current = null;
+    }
+    tilesCacheRef.current = {};
+    analysedFeaturesRef.current = createEmptyAnalysedFeatures();
+  }
+
+  const applySpatialFilter = useCallback(
+    (props) => {
+      // props.data is the same as props.tile.content
+      let { data, uniqueIdProperty = 'cartodb_id' } = props;
```
padawannn, line 71: instance of this `uniqueIdProperty = 'cartodb_id'` I think that we could use the functions that returns que unique property id

padawannn (changes requested):

PR #231, opened 2021-11-30
Remove lock in CategoryWidget if selected categories change

padawannn:
- For the category widget, unlock the widget if there are not a selected category

### DIFF
--- a/packages/react-redux/src/slices/cartoSlice.d.ts
+++ b/packages/react-redux/src/slices/cartoSlice.d.ts
@@ -50,7 +50,6 @@ declare enum CartoActions {
   SET_BASEMAP = 'carto/setBasemap',
   ADD_FILTER = 'carto/addFilter',
   REMOVE_FILTER = 'carto/removeFilter',
-  REMOVE_FILTER_BY_OWNER = 'carto/removeFilterByOwner',
   CLEAR_FILTERS = 'carto/clearFilters',
   SET_VIEWPORT_FEATURES = 'carto/setViewportFeatures',
   REMOVE_VIEWPORT_FEATURES = 'carto/removeViewportFeatures',
@@ -114,13 +113,6 @@ export function removeFilter(
 ): {
   type: CartoActions.REMOVE_FILTER;
   payload: FilterCommonProps;
-};
-
-export function removeFilterByOwner(
-  arg: { id: string, owner: string }
-): {
-  type: CartoActions.REMOVE_FILTER_BY_OWNER;
-  payload: { id: string, owner: string };
 };
 
 export function clearFilters(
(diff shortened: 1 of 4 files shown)


packages/react-redux/src/slices/cartoSlice.d.ts
```diff
@@ -104,7 +105,7 @@ export function setBasemap(
 export function addFilter(
   filter: Filter
 ): {
-  type: CartoActions.ADD_LAYER;
+  type: CartoActions.ADD_FILTER;
```
VictorVelarde, line 108: Was this a BUG??. ⚠️

padawannn: Yes, but it's related to typescript. Does not affect functionality

packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -233,6 +250,16 @@ export const removeFilter = ({ id, column }) => ({
   payload: { id, column }
 });
 
+/**
+ * Action to remove filter by widget owner
+ * @param {id} - sourceId of the filter to remove
+ * @param {owner} - widgetId of the filter to remove
+ */
+export const removeFilterByOwner = ({ id, owner }) => ({
```
VictorVelarde, line 258: Owner is the internal name, but I would probably suggest something more clear for the end user, like 'removeFilterFromWidget ({ sourceId, widgetId })

VictorVelarde: Also, I'm wondering if we need the sourceId, or if we can just iterate on all sources, without asking for the param, Does this make sense?

Can a widget affect several sources? I don't think so... right?

packages/react-ui/src/widgets/CategoryWidgetUI.js
```diff
@@ -300,6 +300,12 @@ function CategoryWidgetUI(props) {
     [maxValue]
   );
 
+  useEffect(() => {
+    if (selectedCategories.length === 0) {
```
VictorVelarde, line 304: Ok, so we somehow assume that changing data works well 'removing' the selected categories...

packages/react-redux/src/slices/cartoSlice.js
padawannn: You're right. I did it like this because it made my life easier for the integration with builder but it would be better to remove it

padawannn: Thinking better of it, this action is not necessary. Removed

packages/react-ui/src/widgets/CategoryWidgetUI.js
padawannn: In Carto 2, when you lock the category widget you cannot select or unselect categories but in c4r you can select and unselect categories although the widget is blocked. I think that it is a bug and we should fix it in this PR.
If we fix it this new `useEffect` works really well, we can talk about this

padawannn: In Carto 2, when you lock the category widget you cannot select or unselect categories but in c4r you can select and unselect categories although the widget is blocked. I think that it is a bug and we should fix it in this PR.
If we fix it this new useEffect works really well, we can talk about this

PR #234, opened 2021-12-03
Make Histogram more flexible, accepting filters with null to select bars

padawannn:
Histogram uses `undefined` value to save the value of its filters if you filter using the first column. The problem is that `stringify` an undefined value in array returns a null value so you cannot recover the selected bar from a stored configuration.


VictorVelarde (approved):
LGTM.

Josmorsot (approved):
👍

PR #237, opened 2021-12-09
Fix build adding peerDependecy of @carto/react-core to @carto/react-ui

Clebal:


packages/react-ui/webpack.config.js
```diff
@@ -8,6 +8,7 @@ const webpackConfig = {
   ...webpackBaseConfig,
   externals: [ 
     /^@material-ui\/.+$/,
+    '@carto/react-core',
```
VictorVelarde, line 11: In react-api we use something like: 
```
  externals: [
    /^@carto\/react-.+$/,
```

Clebal: Yes, but I guessed that we don't need every package.

VictorVelarde (approved):
👍🏻

PR #238, opened 2021-12-09
Refactor useViewportFeatures

Clebal:


packages/react-api/src/hooks/useCartoLayerProps.js
```diff
@@ -1,19 +1,29 @@
...
 export default function useCartoLayerProps({
   source,
   uniqueIdProperty,
   viewportFeatures = true,
-  viewporFeaturesDebounceTimeout = 500
+  viewporFeaturesDebounceTimeout = 250
 }) {
-  const [onViewportLoad, onDataLoad, fetch] = useViewportFeatures(
-    source,
+  const viewport = useSelector((state) => state.carto.viewport);
```
VictorVelarde, line 14: Why have you pushed up this viewport selection and not left it inside the 'viewportFeatures' hooks?

packages/react-api/src/hooks/useCartoLayerProps.js
```diff
@@ -1,19 +1,29 @@
...
+
+  const [onDataLoad] = useGeoJsonFeatures(source, {
+    viewport,
+    uniqueIdProperty,
+    debounceTimeout: viewporFeaturesDebounceTimeout
+  });
```
VictorVelarde, line 20: I wouldn't change to positional params with 'source' here, like we discussed in the RFC, for consistency reasons. I would keep the 'set of params', with:

```
  const [onDataLoad] = useGeoJsonFeatures({
    source, 
    uniqueIdProperty,
    debounceTimeout: viewporFeaturesDebounceTimeout
  })
```

And if you really want to create 2 levels, with optional params:

```
  const [onDataLoad] = useGeoJsonFeatures({
    source, 
    options: { 
           uniqueIdProperty,
           debounceTimeout: viewporFeaturesDebounceTimeout
     }
  })
```

packages/react-api/src/hooks/useCartoLayerProps.js
Clebal: Because each time 'viewportFeatures' hooks changes, useCartoLayerProps changes, if the same change happens in two different hook that are children of the a certain hook, it's better to have that change listener (in this case the selector) in the parent hook. Also, it's a improvement to make `useGeoJsonFeatures` and `useTilesetFeatures` independent from viewport.

packages/react-api/src/hooks/useCartoLayerProps.js
Clebal: Updated.

packages/react-api/src/hooks/useCartoLayerProps.js
VictorVelarde: I'm curious about independent from viewport in the future... you will tell us... 😄

Clebal: I hope to be able to tackle that soon! Many steps have to be taken before 😄

packages/react-api/src/hooks/useGeoJsonFeatures.js
```diff
@@ -0,0 +1,92 @@
...
+  const computeFeaturesGeoJson = useCallback(
+    ({ viewport, uniqueIdProperty }) => {
+      executeTask(sourceId, Methods.VIEWPORT_FEATURES_GEOJSON, {
+        viewport: Float32Array.of(...viewport),
+        uniqueIdProperty
+      })
+        .then(() => {
+          setSourceViewportFeaturesReady(true);
+        })
+        .catch(throwError);
```
VictorVelarde, line 45: Previously we had a specific error case that was ignored: 
`if (error.name === 'AbortError') return;` and it has been removed, why?

VictorVelarde: Also, why changing modern await & try catch with classic thenable?

packages/react-api/src/hooks/useGeoJsonFeatures.js
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }, [source]);
+
+  const onDataLoad = useCallback(
+    (geojson) => {
+      clearDebounce();
+      setSourceViewportFeaturesReady(false);
+      executeTask(sourceId, Methods.LOAD_GEOJSON_FEATURES, { geojson })
+        .then(() => setGeoJsonLoaded(true))
+        .catch(throwError);
```
VictorVelarde, line 86: Same here. I don't mind a lot about then.catch, but what about the previous catch errors we had?

packages/react-api/src/hooks/useGeoJsonFeatures.js
Clebal: Not removed, just it was generalised:

```
export function throwError(error) {
  if (error.name === 'DataCloneError')
    throw new Error(
      `DataCloneError: Unable to retrieve GeoJSON features. Please check that your query is returning a column called "geom" or that you are using the geoColumn property to indicate the geometry column in the table.`
    );
  if (error.name === 'AbortError') return;
  throw error;
}
```

You can find it in [utils.js](https://github.com/CartoDB/carto-react/pull/238/files#diff-a6dd04cfab907cd4586bf8992ef06a0f6de2107df277078f0d2d0c2431ca7ace).

packages/react-api/src/hooks/useGeoJsonFeatures.js
Clebal: Same answer.

VictorVelarde (commented):
Just small comments, but LGTM. I think it's a nice idea to have them separated. 
My only concern is the error management we had and now it's gone.

Let's have tomorrow another quick CR and talk and then we can merge.

Clebal:
@VictorVelarde error management isn't gone, still there. It was just generalised fof having the same one in both hooks.

packages/react-api/src/hooks/useGeoJsonFeatures.js
Clebal: At least for understanding, try catch option is only better when you have multiple await or many lines inside try. With just one promise and using just a few lines, I think try catch makes the code worse to follow and needs more lines than thenable option to make the same.

somethingAsync().then(...).catch(...)

It's better, at least for me, than:

try {
     const res = await somethingAsync()
     doSomething(res)
} catch (error) {
     ...
}

Btw, I don't think it's more modern, it's just another option that has its good/bad use cases like any other option.

packages/react-api/src/hooks/useGeoJsonFeatures.js
VictorVelarde: I didn't see the custom throw. Agree that having it with then is more compact, thus probably more readable in this case.

VictorVelarde:
We will check with @padawannn tomorrow before the merge

packages/react-api/src/hooks/useCartoLayerProps.js
```diff
@@ -1,19 +1,31 @@
...
   source,
   uniqueIdProperty,
   viewportFeatures = true,
-  viewporFeaturesDebounceTimeout = 500
+  viewporFeaturesDebounceTimeout = 250
 }) {
-  const [onViewportLoad, onDataLoad, fetch] = useViewportFeatures(
+  const viewport = useSelector((state) => state.carto.viewport);
+
+  const [onDataLoad] = useGeoJsonFeatures({
```
padawannn, line 16: I like the idea of split this into two hooks

packages/react-api/src/hooks/useGeoJsonFeatures.js
```diff
@@ -0,0 +1,92 @@
...
+        dispatch(setViewportFeaturesReady({ sourceId, ready }));
+      }
+    },
+    [dispatch, sourceId]
+  );
+
+  const computeFeaturesGeoJson = useCallback(
+    ({ viewport, uniqueIdProperty }) => {
+      executeTask(sourceId, Methods.VIEWPORT_FEATURES_GEOJSON, {
+        viewport: Float32Array.of(...viewport),
```
padawannn, line 39: I would remove the `Float32Array` to have a code more readable

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Implement C4R filtering using binary data [#228](https://github.com/CartoDB/carto-react/pull/228)
 - Fix build adding peerDependecy of @carto/react-core to @carto/react-ui [#237](https://github.com/CartoDB/carto-react/pull/237)
+- Refactor useViewportFeatures [#238](https://github.com/CartoDB/carto-react/pull/238)
```
VictorVelarde, line 7: ❤️

padawannn (approved):
LGFM, some suggestions

packages/react-api/src/hooks/useGeoJsonFeatures.js
VictorVelarde: Agree

packages/react-api/src/hooks/useGeoJsonFeatures.js
```diff
@@ -0,0 +1,92 @@
...
+
+export default function useGeoJsonFeatures({
+  source,
+  viewport,
+  uniqueIdProperty = 'cartodb_id',
+  debounceTimeout = 250
+}) {
+  const dispatch = useDispatch();
+  const [isGeoJsonLoaded, setGeoJsonLoaded] = useState(false);
+  const debounceIdRef = useRef(null);
```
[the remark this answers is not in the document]
Clebal: Done.

packages/react-api/src/hooks/useGeoJsonFeatures.js
Clebal: Done.

VictorVelarde (approved):
🚀  nice job!

PR #239, opened 2021-12-14
Add Clear button to TimeSeriesWidget

padawannn:
- [x] If show controls are disabled in Timeseries widget you cannot clear the Timeseries filter --> add Clear btn, as in the rest of widgets
- [x] Also enable the speed button, even though the animation has not started


Clebal (approved):
LGTM

packages/react-ui/__tests__/widgets/TimeSeriesWidgetUI.test.js
```diff
@@ -233,4 +233,18 @@ describe('TimeSeriesWidgetUI', () => {
       jest.runAllTimers();
     });
   });
+
+  test('clear when animation is working', () => {
+    const onStop = jest.fn();
+    render(<Widget isPlaying={true} onStop={onStop} />);
+    fireEvent.click(screen.getByText(/Clear/));
+    setTimeout(() => expect(onStop).toHaveBeenCalledTimes(1), 250);
```
VictorVelarde, line 241: Can't we avoid using a setTimeout for expect? Have you tried with async() and then screen.findByText(/Clear/)

VictorVelarde: Inside testingjavascript materials, there is a little utility, called fire-event-async, that waits for next tick.
That could be useful here:

image

PD: video from https://testingjavascript.com/lessons/svelte-use-dom-testing-library-with-svelte (01:26)

VictorVelarde: So we can try to get a code with something like:

```
    const onStop = jest.fn();
    render(<Widget timeWindow={[1517785200000, 1528063200000]} onStop={onStop} />);
    fireEventAsync.click(await screen.findByText(/Clear/));
    expect(onStop).toHaveBeenCalledTimes(1);
```

VictorVelarde (commented):
Small comment on tests to check, but LGTM

packages/react-ui/__tests__/widgets/TimeSeriesWidgetUI.test.js
padawannn: @VictorVelarde I've removed the `setTimeout `

PR #240, opened 2021-12-15
Avoid unuseful view state updates

Clebal:
### FAQs

- Why have you reduced viewport update debounce time?
  - To make it almost invisible. We also have a debounce in useFeatures hooks, this is almost not needed. The point to use a minimal debounce is that setViewState is called many many times while zooming, which is awful, so a minimal debounce is still needed.

- Why are you deep comparing view state?
  - Because some calles are made with the same view state, and that causes duplicated calculations.

### DIFF
--- a/packages/react-redux/src/slices/cartoSlice.js
+++ b/packages/react-redux/src/slices/cartoSlice.js
@@ -3,6 +3,7 @@ import { WebMercatorViewport } from '@deck.gl/core';
 import { debounce } from '@carto/react-core';
 import { removeWorker } from '@carto/react-workers';
 import { setDefaultCredentials } from '@deck.gl/carto';
+import { dequal } from 'dequal';
 
 /**
  *
@@ -87,8 +88,7 @@ export const createCartoSlice = (initialState) => {
         state.basemap = action.payload;
       },
       setViewState: (state, action) => {
-        const viewState = action.payload;
-        state.viewState = { ...state.viewState, ...viewState };
+        state.viewState = action.payload;
       },
       setViewPort: (state) => {
         state.viewport = new WebMercatorViewport(state.viewState).getBounds();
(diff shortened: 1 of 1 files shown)


packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -280,8 +280,13 @@ export const setViewState = (viewState) => {
       delete viewState[viewProp];
     }
 
-    dispatch(_setViewState(viewState));
-    debouncedSetViewPort(dispatch, _setViewPort);
+    const { carto: currentState } = getState();
+    const newViewState = { ...currentState.viewState, ...viewState };
+    // Deep compare current viewState and the new one.
+    if (!dequal(currentState.viewState, newViewState)) {
```
padawannn, line 286: I no totally agree with this comparison. setViewState only be called if the `viewState` has changed so if the client that uses c4r is calling this function with the same value the problem is on the client-side and we should fix it in the template

packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -254,7 +254,7 @@ export const selectIsViewportFeaturesReadyForSource = (state, id) =>
 
 const debouncedSetViewPort = debounce((dispatch, setViewPort) => {
   dispatch(setViewPort());
-}, 200);
+}, 25);
```
padawannn, line 257: what happens if we remove the debounce? I've seen that kepler doesn't use a debounce for this and we have a debounce in the viewport features calculation

padawannn (changes requested):

packages/react-redux/src/slices/cartoSlice.js
Clebal: When you are moving a layer in nebula.gl, deck.gl calls onViewStateChange with the same viewState. This can also happen in some other cases.

packages/react-redux/src/slices/cartoSlice.js
Clebal: I tried removing the debounce but this action is called many many many times while zooming and panning, I dont think we should update redux state so fastly. That's why I think it's convenience to have at least a very small debounce. Also, in viewport features the behavior was weird. I felt the app strange. I don't remember the details, but with this debounce, it worked perfectly.

packages/react-redux/src/slices/cartoSlice.js
padawannn: I'm a bit worried that we are masking the error. We're going to research about this

PR #241, opened 2021-12-15
Add fn for applying filters to tile content

Clebal:


packages/react-core/src/filters/applyFiltersToTileContent.js
```diff
@@ -0,0 +1,31 @@
...
+    filterFn(idx, data);
+
+  return {
+    ...tileContent,
+    ...['points', 'lines', 'polygons'].reduce((acc, key) => {
+      acc[key] = {
+        ...tileContent[key],
+        attributes: {
+          getFilterValue: {
+            value: tileContent[key].featureIds.value.map(
```
padawannn, line 21: Instead of use `tileContent[key].featureIds` we could use `primitivePolygonIndices` or `pathIndices` or `pointIndices`. If we use `featureIds` we're repeating calculation for lines and polygons

padawannn (changes requested):

packages/react-core/src/filters/applyFiltersToTileContent.js
Clebal: Good catch 👌  I'll change it.

Clebal: Done

packages/react-core/src/filters/applyFiltersToTileContent.js
```diff
@@ -0,0 +1,82 @@
...
+          size: 1
+        }
+      }
+    }
+  };
+}
+
+// Aux
+function applyToPolygonsLinesData(data, applyFiltersFn) {
+  const featuresGetFilterValue = new Uint16Array(data.featureIds.value.length);
```
padawannn, line 59: According to documentation https://github.com/visgl/loaders.gl/blob/master/modules/gis/docs/api-reference/geojson-to-binary.md `featureIds` can be Uint16Array | Uint32Array so we could change this by `Uint32Array`

packages/react-core/src/filters/applyFiltersToTileContent.js
```diff
@@ -0,0 +1,82 @@
...
+
+  let idx = 0,
+    start = null;
+  for (const featureId of data.featureIds.value) {
```
padawannn, line 63: This solution works but it should be better if we do the loop over `primitivePolygonIndices` or `pathIndices` and you would not need to do this condition ```if (featureId === data.featureIds.value[idx + 1])``` or work with two index (idx and idx+1).

In any case we can change it next week, it's not a blocker

padawannn (changes requested):

packages/react-core/src/filters/applyFiltersToTileContent.js
Clebal: Updated.

PR #246, opened 2021-12-16
Add spatial filter actions in Redux

Clebal:
**MERGE #247 BEFORE THIS ONE**

### DIFF
--- a/packages/react-redux/src/slices/cartoSlice.js
+++ b/packages/react-redux/src/slices/cartoSlice.js
@@ -48,6 +48,7 @@ export const createCartoSlice = (initialState) => {
       dataSources: {
         // Auto import dataSources
       },
+      spatialFilter: null,
       viewportFeatures: {},
       viewportFeaturesReady: {},
       ...initialState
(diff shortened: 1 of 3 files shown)


packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -93,6 +94,30 @@ export const createCartoSlice = (initialState) => {
...
+          const source = state.dataSources[sourceId];
+
+          if (source) {
+            source.spatialFilter = null;
+          }
+        } else {
+          state.spatialFilter = null;
+        }
+      },
       addFilter: (state, action) => {
```
padawannn, line 121: I think that `clearFilters` action should also remove the spatial filter for the datasource. Do you agree @VictorVelarde and @Clebal ?

padawannn (changes requested):
I'm wondering if

packages/react-redux/src/slices/cartoSlice.js
borja-munoz: I've not been summoned but I'd say it shouldn't. The `clearFilters` action should remove the filters added by the `addFilter` action. As we have decided to not treat the spatial filter like the other filters, we shouldn't remove it using this action.

Clebal: Agree with @borja-munoz

PR #248, opened 2021-12-16
Fix TS type declarations for BasemapName and CartoSlice

juandjara:
Hello team! I've found some Typescript errors when setting up [a new project](https://github.com/cartodb/initiative-vision) from the template with Carto3 and Typescript and I saw a very easy correction of them so I created this PR. Let me know if you need any more context, tests, docs or anything like that

There are two changes in this PR:
* changing the type of the `createCartoSlice` function. Instead of using the custom `Reducer` type from the `types.ts` file of the C4R library, we are importing the `Reducer` type from `redux`. This is a generic type, and as such, it should be defined as `Reducer<any, AnyAction>`. The VS Code tooltips let me know about this when I saw that the types for adding the cartoSlice to the store in the TS-template were not matching the redux types.

* changing the type of `BasemapName` from using `&` to using `|`. The `&` operator in Typescript creates an intersection between types that does not work for joining two enum types together. I learned about that [here](https://stackoverflow.com/questions/48478361/how-to-merge-two-enums-in-typescript) when I was debugging why VS Code was again complaining in the basemap references scattered across the template. After much research, try, and error, I discovered that you just need to change the `&` for an `|` to mean "This type represents values from this enum OR this other enum"

Edit:
Following @Clebal comment and the [redux-typescript documentation](https://redux.js.org/usage/usage-with-typescript#type-checking-reducers) I've changed the `cartoSlice` type to further indicate the types of the reducer using `Reducer<CartoState, AnyAction>`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Refactor useViewportFeatures [#238](https://github.com/CartoDB/carto-react/pull/238)
 - Add clear button for TimeSeriesWidget widget and enable the speed button even though the animation has not started  [#239](https://github.com/CartoDB/carto-react/pull/239)
 - Improve timeseries animation performance [#243](https://github.com/CartoDB/carto-react/pull/243)
+- Fix TS type declarations for BasemapName and CartoSlice [#248](https://github.com/CartoDB/carto-react/pull/248) 
 
 ## 1.1.3 (2021-12-04)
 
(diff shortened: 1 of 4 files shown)


bbecquet (approved):

VictorVelarde:
Code looks good to me, but we should check against Builder, linked locally, before we merge

Clebal (approved):
@juandjara great job! 🚀

Clebal:
@VictorVelarde it's ready to check agains builder.

juandjara:
@Clebal @VictorVelarde thanks for the quick response!
Just FYI, I've checked in my project (Initiative-Vision) and also ran `yarn build` in the root folder `carto-react` and everything works fine 👌

PR #250, opened 2021-12-23
Spatial filter

padawannn:


packages/react-core/src/filters/tileFeatures.js
```diff
@@ -169,31 +180,42 @@ export function viewportFeaturesBinary({ tiles, viewport, uniqueIdProperty }) {
     }
 
     const { bbox } = tile;
-    const tileIsFullyVisible = isTileFullyVisible(bbox, viewport);
-    const viewportIntersection = bboxPolygon(prepareViewport(tile.bbox, viewport));
+    const bboxToGeom = bboxPolygon([bbox.west, bbox.south, bbox.east, bbox.north]);
```
alasarr, line 183: You're trying to unify the two methods (random geometry and viewport features), that's good from a code perspective, I've not tested, but I think it could be slower:

- I feel `booleanWithin` is slower than just trying to check with regular mercator math if a tile is fully visible. 
- I feel `intersect` operation is slower than calculate with regular mercator math.

Have you tried if you've a the same performance with these changes on viewport feature?

As it's an not a massive operation (we only do once per tile), only 16 times, it's not critical, but just want to share it to double check

padawannn: On my computer, the worse downgrade in performance is 7ms for some tiles when the layer is loading but in most cases is 1, 2, or 3ms  or the same performance.

Captura de pantalla 2021-12-27 a las 11 41 21

I think that we can live with that to make the code more readable. What do you think?

alasarr: Makes sense

alasarr (approved):

PR #251, opened 2021-12-23
WRAPPER WIDGET Configurable margin in widgets [sc-196869]

atoom75:
Allow to pass `margin` as props for `WrapperWidgetUI`

### DIFF
--- a/packages/react-ui/src/widgets/WrapperWidgetUI.js
+++ b/packages/react-ui/src/widgets/WrapperWidgetUI.js
@@ -43,7 +43,7 @@ const useStyles = makeStyles((theme) => ({
     width: '100%',
     height: theme.spacing(0.25)
   },
-  header: ({ expanded, margin }) => ({
+  header: ({ expanded }) => ({
     display: 'flex',
     flexDirection: 'row',
     alignItems: 'center',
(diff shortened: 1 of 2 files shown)


alasarr (approved):

packages/react-ui/src/widgets/WrapperWidgetUI.js
```diff
@@ -43,14 +43,14 @@ const useStyles = makeStyles((theme) => ({
     width: '100%',
     height: theme.spacing(0.25)
   },
-  header: ({ expanded }) => ({
+  header: ({ expanded, margin }) => ({
```
padawannn, line 46: This margin is never used

padawannn (changes requested):
I like the approach but with default values, all widgets are 14px lower in height than the current widgets. I'm wondering if our current widgets are wrong or we need to fix it in this PR

Clebal:
In DrawingToolWidgetUI, we use className prop to override default root styles. It has more possibilities than only margin. What do you think guys? @alasarr @padawannn @atoom75

padawannn:
> In DrawingToolWidgetUI, we use className prop to override default root styles. It has more possibilities than only margin. What do you think guys? @alasarr @padawannn @atoom75

We already had a meeting about this and the idea of this is precisely not to be able to overwrite all the widget styles

padawannn (approved):
After talking with @alasarr we've decided that we can live with these 14px of height difference

PR #252, opened 2021-12-27
Remove sortBy prop default value

Clebal:
Using `[]` as default sortBy value makes the execution to break due to it's an invalid sortBy value.


alasarr:
Can you provide an example of the failure?

Clebal:
@alasarr 

If you want to get the features of certain source with certain filters, but you don't want to sort the features:
```
executeTask(sourceId, Methods.FEATURES_RAW, { filters: sourceFilters })
```

It makes the code break, because in sorting.js we have a validation of sortBy prop.

alasarr (approved):

PR #263, opened 2022-01-02
Fix: DrawingToolLayer disables interactivity of the layers behind it

padawannn:
When you finish drawing, all layers behind the DrawingToolLayer lose their interactivity. For example, all these layers will not show their popup. To solve it we disable the pick DrawingToolLayer property unless you're editing or drawing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Not released
 
+- Fix: DrawingToolLayer disables interactivity of the layers behind it [#263](https://github.com/CartoDB/carto-react/pull/263)
+
 ## 1.2.0-alpha.0 (2021-12-30)
 
 - Track clientId for tracing purposes [#261](https://github.com/CartoDB/carto-react/pull/261)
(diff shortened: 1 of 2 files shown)


Clebal:
Good catch @padawannn. Only the features behind the polygon loss their interactivity, that's why I would just disable polygon fill when it isn't needed:

```
getFillColor: isEdit && [...mainColor.slice(0, 3), 255 * 0.08],
filled: isEdit,
```

But disabling interactivity is also ok I guess.

Clebal (approved):

padawannn:
> Good catch @padawannn. Only the features behind the polygon loss their interactivity, that's why I would just disable polygon fill when it isn't needed:
> 
> ```
> getFillColor: isEdit && [...mainColor.slice(0, 3), 255 * 0.08],
> filled: isEdit,
> ```
> 
> It wasn't a good approach using [0, 0, 0, 0] to make polygon transparent.
> 
> But disabling interactivity is also ok I guess.

Thank @Clebal but I prefer to disable the pickable property because is easier to understarnd

PR #265, opened 2022-01-03
Use zero-based pagination for getRawFeatures

bbecquet:
Small adjustment on the TableWidget and getRawFeatures, following discussion with @Clebal https://github.com/CartoDB/carto-react/pull/154#discussion_r775987648

Instead of using a 1-based pagination, let's switch to 0-based. This makes it consistent with MUI pagination.

### DIFF
--- a/packages/react-widgets/src/widgets/TableWidget.js
+++ b/packages/react-widgets/src/widgets/TableWidget.js
@@ -27,7 +27,7 @@ function TableWidget({
   onError
 }) {
   const [rowsPerPage, setRowsPerPage] = useState(10);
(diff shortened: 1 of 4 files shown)


packages/react-widgets/src/widgets/TableWidget.js
```diff
@@ -42,7 +42,7 @@ function TableWidget({
 
   useEffect(() => {
     // force reset the page to 1 when the viewport or filters change
```
Clebal, line 44:
```suggestion
    // force reset the page to 0 when the viewport or filters change
```

Clebal (approved):
🚀

PR #268, opened 2022-01-07
Disable interactivity for widgets

Clebal:
Story details: https://app.shortcut.com/cartoteam/story/193004

Added property `filterable` in every widget with filtering capabilities.

When `filterable` is false, widget's click event is disabled.

### DIFF
--- a/packages/react-ui/src/widgets/CategoryWidgetUI.js
+++ b/packages/react-ui/src/widgets/CategoryWidgetUI.js
@@ -27,9 +27,15 @@ const useStyles = makeStyles((theme) => ({
     padding: theme.spacing(0, 1, 1, 0)
   },
 
-  element: {
+  selectable: {
     cursor: 'pointer',
 
+    '&:hover $progressbar div': {
+      backgroundColor: theme.palette.secondary.dark
+    }
+  },
+
+  element: {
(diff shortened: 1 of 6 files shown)


eamador (approved):
LGTM

VictorVelarde (commented):
Code LGTM, nice addition @Clebal 

Just a question: to land this (for an alpha) we can do it without updating docs & without any example on carto-react-template, but I think it would be nice to create also those other 2 small companion PRs, is that possible?.

If not, we can do that later, when making official 1.2 release... thoughts ? cc/ @borja-munoz

Clebal:
@VictorVelarde totally agree. But I cannot tackle that right now since I don't have specific dedication for C4R, I'm sorry. I can make it in a future, maybe for the final release.

borja-munoz:
In my opinion, the carto-react-template example is not mandatory in this case, but updating the documentation (adding the property to the reference) is a must.

VictorVelarde:
I've created this task for doc (https://app.shortcut.com/cartoteam/story/202069/widgets-document-new-filterable-prop-for-widgets), but I think we can already land this one

VictorVelarde (approved):
🚀

PR #271, opened 2022-01-11
Add FeatureSelection tool

padawannn:
- Name refactor (https://app.shortcut.com/cartoteam/story/204565/rename-drawingtoolwidget-and-drawingtoollayer). [sc-204565]
- Google Maps compatibility.
- Mask extension integration


packages/react-redux/src/slices/cartoSlice.d.ts
```diff
@@ -149,3 +149,5 @@ export function setDrawingToolEnabled(enabled: boolean): {
 };
 
 export function selectSpatialFilter(state: any, sourceId?: string): object | null;
+
+export function selectDrawingToolMode(state: any): string | false;
```
VictorVelarde, line 153: If 'false' is equivalent to 'None', I think it could be better to add that option to an enum or type

packages/react-widgets/src/layers/EditableCartoGeoJsonLayer.js
```diff
@@ -0,0 +1,30 @@
+import { EditableGeoJsonLayer } from '@nebula.gl/layers';
+
+const EVENT_TYPES = ['anyclick', 'pointermove', 'panstart', 'panmove', 'panend', 'keyup'];
```
VictorVelarde, line 3: Probably a comment with a bit of explanation on this would be really useful, about how this should be better a mechanism in nebula itself, with probably a PR on that in the future

packages/react-widgets/src/layers/EditableCartoGeoJsonLayer.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default class EditableCartoGeoJsonLayer extends EditableGeoJsonLayer {
+  _addEventHandlers() {
+    const eventManager = this._getEventManager();
+    const { eventHandler } = this.state._editableLayerState;
+
+    for (const eventType of EVENT_TYPES) {
+      eventManager.on(eventType, eventHandler, {
+        // give nebula a higher priority so that it can stop propagation to deck.gl's map panning handlers
+        priority: 100
```
VictorVelarde, line 13: Is 100 the "top priority" or how does is work? I mean, is it an arbitrary number?.

packages/react-redux/src/types.d.ts
```diff
@@ -45,7 +45,7 @@ export type CartoState = {
   spatialFilter: Geometry,
   featuresReady: { [key: string]: boolean },
   drawingToolEnabled: boolean,
-  drawingToolMode: typeof DRAW_MODES
+  drawingToolMode: string
```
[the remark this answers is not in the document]
padawannn: Because `typeof DRAW_MODES` doesn't work when I integrate it in typescript project. I accept suggestions

packages/react-redux/src/slices/cartoSlice.d.ts
padawannn: `selectDrawingToolMode` is a selector that return `false` if the `spatialFilter` is null in the store (the default state). I think that creating another enum type could be over-engineering

packages/react-widgets/src/layers/EditableCartoGeoJsonLayer.js
padawannn: I really don't know XD. This code has been copied from the original nebula code
https://github.com/uber/nebula.gl/blob/master/modules/layers/src/layers/editable-layer.ts#L83

packages/react-widgets/src/layers/EditableCartoGeoJsonLayer.js
padawannn: Agree, I'm going to add a link to the original code and write a small explanation

packages/react-widgets/src/layers/EditableCartoGeoJsonLayer.js
VictorVelarde: XD, ok then add this link to code...

packages/react-redux/src/slices/cartoSlice.d.ts
VictorVelarde: I just see there are other options without adding boolean to the mix:
- a) just an enum / type: DrawingToolMode.Type1, DrawingToolMode.Type2,... + DrawingToolMode.None, or more simple
- b) just respecting the null for 'None'. 

"selectDrawingToolMode" doesn't sound like a boolean to me, but a value from a predefined list. If required, we can later do in code something like this

```
const noFilter = selectDrawingToolMode === null
if (noFilter) ...
```

What do you think?

padawannn: I've done the `b` option

packages/react-widgets/src/layers/EditableCartoGeoJsonLayer.js
padawannn: Done

packages/react-widgets/src/layers/EditableCartoGeoJsonLayer.js
padawannn: Done

VictorVelarde (approved):
:+1

VictorVelarde:
@padawannn I'm not sure if you prefer to merge this one together with https://github.com/CartoDB/carto-react/pull/278, but I'm ok with this one, as it is now. 

Is it mergeable independently?

padawannn:
> @padawannn I'm not sure if you prefer to merge this one together with #278, but I'm ok with this one, as it is now.
> 
> Is it mergeable independently? (if that's the case add changelog entry before)

I prefer to merge #278 first due to name refactor requested by Borja

VictorVelarde:
Ok, then adding this as 'blocked' for the moment

packages/react-widgets/src/layers/useEventManager.js
```diff
@@ -0,0 +1,42 @@
...
+// When we're using Google as a base map, Nebula layer doesn't work becase Nebula uses the Deck Gl eventManager
+// so in this case we're defining a new EventManager with the Google maps div. On the other hand, in Builder
+// we don't use the c4r GoogleMap component so we need to send a EventManager as parameter
+export default function useEventManager({
+  eventManager,
+  isEdit,
+  isSelected,
+  selectedMode
+}) {
+  const isGmaps = useSelector((state) => BASEMAPS[state.carto.basemap].type === 'gmaps');
```
Josmorsot, line 16: Is not 'gmaps' in any constant? Should we move it to a new one?

Josmorsot (approved):
LGTM, just minor question/suggestion

packages/react-basemaps/src/basemaps/GoogleMap.js
```diff
@@ -120,12 +121,16 @@ export function GoogleMap(props) {
...
       }
       window.cartoGmap.setOptions(rest);
-      window.cartoDeck.setProps({ layers, getTooltip });
+      window.cartoDeck.setProps({ getTooltip });
     }
-
-    window.cartoDeck.setMap(window.cartoGmap);
   };
 
+  // Set layers should be outside of the useEffect to avoid problems with layers that changes
```
bbecquet, line 128: Won't it cause useless redraws on Deck?

Also, is it normal that the `useEffect` below has no dependency array?

packages/react-widgets/src/layers/FeatureSelectionLayer.js
```diff
@@ -0,0 +1,130 @@
...
+} from '@carto/react-redux';
+import { EDIT_MODES } from '@carto/react-core';
+import { hexToRgb, useTheme } from '@material-ui/core';
+import EditableCartoGeoJsonLayer from './EditableCartoGeoJsonLayer';
+import useEventManager from './useEventManager';
+import MaskLayer from './MaskLayer';
+
+const { ViewMode, TranslateMode, ModifyMode, CompositeMode } = nebulaModes;
+
+// const EditMode = new CompositeMode([new TranslateMode(), new ModifyMode()]);
```
bbecquet, line 18: Comment to remove

packages/react-basemaps/src/basemaps/GoogleMap.js
padawannn: `Won't it cause useless redraws on Deck?`

No, in fact, Deck always should receive a new instance of all layers and it decides if it's necessary to redraw or not. The creation of the layers or setting them within a `useEffect` is a common mistake. Check this example https://deck.gl/docs/get-started/using-with-react#the-deckgl-react-component

`Also, is it normal that the useEffect below has no dependency array?`
In this case yes although I admit that it is difficult to understand. In any case, this PR does not intend to modify that part of the code

packages/react-widgets/src/layers/useEventManager.js
padawannn: We don't have a constant for 'gmaps'. Move it to a new one it's not as easy as it sounds because the basemap package also uses that string so probably we would need to add it to the core package and do a refactor. @VictorVelarde maybe we could do it in another PR

packages/react-widgets/src/layers/FeatureSelectionLayer.js
padawannn: Done, thx

bbecquet (approved):
LGTM

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## Not released
 
 - Upgrade storybook and fix custom-component stories [#303](https://github.com/CartoDB/carto-react/pull/303)
+- Name refactor (DrawingTool for FeatureSelection)
+- Add compatibility between Google Maps and FeatureSelectionLayer
+- MaskExtension integration
```
VictorVelarde, line 8:
```suggestion
- Integrate new deck.gl MaskExtension, required by FeatureSelectionLayer
```

VictorVelarde: ... and also PR links in other entries

VictorVelarde (approved):
:rocket Well done

PR #273, opened 2022-01-12
Remove uniqueIdProperty default value in useGeojsonFeatures

Clebal:
We cannot asume that every layer has cartodb_id since it's not mandatory in Carto 3. Users are facing a problem when cartodb_id is not defined.
This is a wrong behavior added when the useViewportFeatures were refactored, because before the refactor we didn't have a default value for uniqueIdProperty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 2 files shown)


padawannn (approved):

VictorVelarde (approved):
🚀

VictorVelarde:
Patch v1.1.5 released

PR #276, opened 2022-01-13
New InputFile component

builtbyedgar:
Shortcut: https://app.shortcut.com/cartoteam/story/145206/component-input-file


packages/react-ui/src/custom-components/InputFile.js
```diff
@@ -0,0 +1,137 @@
...
+    setOver(false);
+  };
+
+  const handleDrop = (event) => {
+    event.preventDefault();
+
+    const newFiles = [];
+    const items = event.dataTransfer.items;
+
+    for (let i = 0; i < items.length; i++) {
```
Josmorsot, line 74: Why are you not using a forEach instead?

I would change it for:

```ts
items.forEach(item => { .... })
```

Also I think that would be better to move this block to a new function, something like that:
```ts
// handleDrop function
const newFiles = getAllFiles(items)
props.onChange(newFiles)
```

Josmorsot (changes requested):
Try to use forEach instead.

LGTM, just minor suggestions

VictorVelarde:
Same here @builtbyedgar. Tell us if you'll be able to have availability to fix comments soon or not, please

VictorVelarde:
@Josmorsot. You should please take also this one, when you can

packages/react-ui/src/custom-components/InputFile.js
Josmorsot: `items` is not an array, so instead using `forEach` I'm going to use a `for(item of items)`

Josmorsot (approved):

PR #279, opened 2022-01-14
Table widget UI tweaks

bbecquet:
Tweaks to the TableWidget to improve its usability:
 - allow to specify a fixed pixel `height` prop, necessary to make the table content scrollable in case of overflow
 - make the header "sticky", so it doesn't scroll with the content
 - use the `size="small"` variant, which reduces cell padding for a denser layout

### DIFF
--- a/packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
+++ b/packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
@@ -74,7 +74,8 @@ function TableWidgetUI({
   rowsPerPageOptions,
   onSetRowsPerPage,
   onRowClick,
-  height
+  height,
+  dense
 }) {
   const classes = useStyles();
   const paginationRef = useRef(null);
@@ -103,7 +104,7 @@ function TableWidgetUI({
   return (
     <>
       <TableContainer style={fixedHeightStyle}>
-        <Table stickyHeader size='small'>
+        <Table stickyHeader size={dense ? 'small' : 'medium'}>
           <TableHeaderComponent
             columns={columns}
             sorting={sorting}
@@ -197,7 +198,8 @@ TableWidgetUI.defaultProps = {
   sortDirection: 'asc',
   pagination: false,
   rowsPerPage: 10,
-  rowsPerPageOptions: [5, 10, 25]
+  rowsPerPageOptions: [5, 10, 25],
+  dense: false
 };
 
 TableWidgetUI.propTypes = {
@@ -216,7 +218,8 @@ TableWidgetUI.propTypes = {
   rowsPerPageOptions: PropTypes.array,
   onSetRowsPerPage: PropTypes.func,
   onRowClick: PropTypes.func,
-  height: PropTypes.string
+  height: PropTypes.string,
+  dense: PropTypes.bool
 };
 
 export default TableWidgetUI;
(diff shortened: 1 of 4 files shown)


borja-munoz:
Hi Benjamin, please let's discuss changes in the interface/properties, specially in the widget, before implementing them. It is also important to be sure we are prioritizing the right work.

For instance, in this case we already have a `height` property in the `PieWidget` with a different data type, and we should provide a consistent interface among widgets .

bbecquet:
Hi @borja-munoz,
Thanks for the mention of the type of `height`. I changed it so it's a string like the other widgets exposing this prop.
In terms of priority, this change is required to make the content scrollable for the integration in builder, so it's not a new feature :)
Same for the denser layout, I realized the most recent designs use this style.

packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
```diff
@@ -89,10 +94,16 @@ function TableWidgetUI({
...
+    const paginationHeight = paginationRef?.current?.clientHeight || 0;
+    fixedHeightStyle.height = `calc(${height} - ${paginationHeight}px)`;
+  }
+
   return (
     <>
-      <TableContainer>
-        <Table>
+      <TableContainer style={fixedHeightStyle}>
+        <Table stickyHeader size='small'>
```
VictorVelarde, line 106: If this is gonna be 'hardcoded' then we should ping @Clebal about it for opinion

bbecquet: We can easily expose another prop if needed.
But I really think this should be the default. This matches the designs of integration in builder and makes the table much more usable. The default cell padding of MUI is huge which is quite unpractical in most scenarios when space is constrained.

Clebal: I think Jose designed it using default size as medium.

Size small
image

Size medium
image

Jose's design
image

In any case, **I think this should be a prop**.

Clebal: I like 'dense' approach @bbecquet 👏🏻

bbecquet: @Clebal Alright :+1: 
As `size=small|medium` is MUI terminology and isn't very clear, I've expose a boolean prop `dense`, and set it to `false` by default. Builder will use `true`.
I've also added a case in TableWidgetUI storybook to illustrate that.

packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
```diff
@@ -89,10 +91,16 @@ function TableWidgetUI({
     onSetPage(0);
   };
 
+  const fixedHeightStyle = {};
+  if (height) {
+    const paginationHeight = paginationRef?.current?.clientHeight || 0;
+    fixedHeightStyle.height = `calc(${height} - ${paginationHeight}px)`;
+  }
```
Clebal, line 98: What do you think about using a useMemo here? Idk if it's useful, but I find weird to see this kind of logic being executed in every render.

Clebal (approved):
🚀

PR #280, opened 2022-01-17
Fix executeSQL typing

Clebal:
Changes made:

- Added a generic to customise executeSQL output type.
- Fixed executeSQL JSDoc: it had the old description of the fn.

I have some doubts about the need of ExecuteSQLResponse type, because it just adds the promise. But it didn't want to remove it to avoid a breaking change. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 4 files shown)


Josmorsot (approved):
LGTM

PR #281, opened 2022-01-17
Allow to set the initial page size on TableWidget

bbecquet:
Add an `initialPageSize` prop to `TableWidget` to set the number of rows per page when rendering the component first.

This number can be changed after by the user through the pagination UI, but it will change only the internal state of the component, resulting in a possible inconsistency of props vs. state.
The **initial** prefix is meant to emphasize this.


Clebal (commented):
In [this RFC](https://app.shortcut.com/cartoteam/story/197577/rfc-set-expanded-collapsed-value-to-wrapperwidgetui-and-legendwidgetui) I proposed to name this kind of prop as: `default*`. `initial*` seems also ok for me.
I think @borja-munoz should say which one is better.

borja-munoz:
I like `initial` better and we should adapt the other RFC to be consistent.

Clebal (approved):
🚀

PR #282, opened 2022-01-17
Add initialCollapsed prop to LegendWidget

Clebal:
image

Changes made:

- Allow to render LegendWidget initially collapsed.
- Added TS types, prop types and JSDocs.

Breaking change:
- LegendWidgetUI is now dummy.

### DIFF
--- a/packages/react-widgets/src/widgets/LegendWidget.js
+++ b/packages/react-widgets/src/widgets/LegendWidget.js
@@ -1,19 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { updateLayer } from '@carto/react-redux/';
 import { LegendWidgetUI } from '@carto/react-ui';
 import { useDispatch, useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
 
-export default function LegendWidget({ className }) {
+/**
+ * Renders a <LegendWidget /> component
+ * @param  {object} props
+ * @param  {string} props.className - CSS class name
+ * @param  {string} props.initialCollapsed - Define initial collapsed value. false by default.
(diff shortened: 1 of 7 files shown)


packages/react-widgets/src/widgets/LegendWidget.js
```diff
@@ -1,19 +1,29 @@
...
+ * Renders a <LegendWidget /> component
+ * @param  {object} props
+ * @param  {string} props.className - CSS class name
+ * @param  {string} props.initialCollapsed - Define initial collapsed value. false by default.
+ */
+function LegendWidget({ className, initialCollapsed }) {
   const dispatch = useDispatch();
   const layers = useSelector((state) =>
-    Object.values(state.carto.layers).filter((layer) => !!layer.legend)
+    [...Object.values(state.carto.layers), ...Object.values(state.carto.layers)].filter(
```
Josmorsot, line 16: Could you explain what are you doing here? Maybe you can create a selector to be more specific

Josmorsot (approved):
LGTM, just a minor question/suggestion

packages/react-widgets/src/widgets/LegendWidget.js
Clebal: Ups, sorry. That was for testing. I'm going to remove it now

Clebal: Updated.

Clebal:
We need to discuss if it's interesting to have a initial* prop.

@padawannn @VictorVelarde @bbecquet

padawannn (approved):

PR #269, opened 2022-01-11
Improve styles for MaterialUI - ToggleButton & ToggleButtonGroup components

builtbyedgar:
Implement the Carto theme for Toggle Button of Material UI with the current design

Story: [Component - Toggle button](https://app.shortcut.com/cartoteam/story/140023/component-toggle-button)

PR #289, opened 2022-01-20
Remove custom ToggleButtonGroup and ToggleButton default size value

Clebal:
In the PR #269, Edgar modified the default size value for ToggleButtonGroup and ToggleButton components.

In Material UI, both components have medium `size` as default value, instead of `small`.

That change made the projects styles to break:

Using small:
image

Using medium:
image

This PR is to remove small size, and leave the default Material UI value as our default value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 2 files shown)


builtbyedgar (approved):
LGTM

PR #292, opened 2022-01-20
Fix size default value in ToggleButton storybook

Clebal:
Use medium as default value in ToggleButton storybook

image


builtbyedgar (approved):
LGTM

PR #295, opened 2022-01-21
Use LegendComponent in LegendWidgetUI as JSX component to avoid react errors

Clebal:
Story details: https://app.shortcut.com/cartoteam/story/198202

The error happened when you render a legend that uses hooks (icon and proportion) after the first rendering. Because LegendComponent was being called as hook instead of JSX component. Using JSX component, you can change hooks order, but not using a hook.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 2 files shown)


Josmorsot (approved):
LGTM

PR #302, opened 2022-01-25
Fix react-redux FilterBasic typing

flipasg:
# Description

Fix react-redux FilterBasic typing:

- filter property now uses _FilterTypes type based in the FilterTypes constant of react-core package, solving the problems related with the '' type.
- _FilterTypes now is exported in the types.d.ts file, solving the problems related with the import of this type.
- owner now is optional.

## Type of change
- Fix


VictorVelarde (approved):
LGTM

PR #304, opened 2022-01-26
Fix MuiComponents in Storybook 

Josmorsot:
# Description
 This PR includes 2 POC, first applied in `Chip` (based on @Clebal's  PR: #172)

Pros: Works fine with the playground section
Cons: It's not going to be useful in all cases.

I'm in favor of doing it directly in each `stories` file, just using a few helpers function instead of a "big magical" function that solve all cases. 

Second solution applied in `Palette`
Pros: You have the absolute control about what is being displayed.
Cons: It's completely manual and Playground is not sync with code, IMO it should be the last option used.

Why have I used "manual" solution in Palette? There we're applying different styles to each element, using the first solution Storybook is calculating what is `theme.palette.primary.main` and instead display the hexadecimal color: `#XXXXXX`

## Acceptance

First: https://sb-carto-josmorsot.vercel.app/?path=/docs/common-chip--default
Second: https://sb-carto-josmorsot.vercel.app/?path=/docs/carto-theme-palette--default  => just playground story


VictorVelarde:
I would probably prefer 'Second option' (total control of code displayed, even loosing code<-->component sync in Default story), as here I understand we should give top priority to users (quick copy & paste code), but please attach vercel link for acceptance & decision on the approach.

VictorVelarde:
@Clebal and @builtbyedgar you have both worked using Storybook, either for new components or polishing theme... once shared here, it would be nice to receive also your feedback / comments on this change

borja-munoz:
In my opinion, if the playground is not synced with code, I'd hide the "Show code" option and will keep the default story just for testing the different combinations in a visual way. Then we have the other stories to see what code we need to use.

I cannot test the approach followed for the chip component because the storybook is throwing an error.

PR #306, opened 2022-01-27
Fix Timeseries wrong filtering with time window

Josmorsot:
# Description

Shortcut: [(link)](https://app.shortcut.com/cartoteam/story/205677/redshift-public-map-histogram-breaks-when-filtering-with-timeseries)

Timeseries filter was not working properly, array of **time** was not correctly sorted. Why?
`Array.sort()` sort array of number from least to greatest if numbers are > 0 but not with negatives values.
Instead we have to implement a specific callback to sort array of numbers:
`Array.sort((timeA, timeB) => (timeA < timeB ? -1 : 1))`

## Type of change

- Fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Not released
 
 - Upgrade storybook and fix custom-component stories [#303](https://github.com/CartoDB/carto-react/pull/303)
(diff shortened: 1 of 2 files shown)


VictorVelarde (approved):
I'm going to land this quickly to fix production bug. We should add next week (@eamador ?) more test coverage on this filter)

PR #309, opened 2022-02-01
Remove unused mask extension property 

padawannn:
# Description

Remove maskEnabled property and control the mask extension activation by maskId

## Type of change

- Fix


VictorVelarde (approved):
👍🏻

PR #311, opened 2022-02-04
Legend improvements

bbecquet:
- Expose internal legend types components (`LegendCategories`, `LegendIcon`, `LegendRamp`, & `LegendProportion`) to allow re-use outside the high-level `LegendWidget`
 - Add `isStrokeColor` prop to `LegendCategories` to draw colors as circles instead of disks
Capture d’écran de 2022-02-07 15-37-32

### DIFF
--- a/packages/react-ui/src/index.d.ts
+++ b/packages/react-ui/src/index.d.ts
@@ -5,6 +5,10 @@ import FormulaWidgetUI from './widgets/FormulaWidgetUI';
 import HistogramWidgetUI from './widgets/HistogramWidgetUI';
 import PieWidgetUI from './widgets/PieWidgetUI';
 import LegendWidgetUI, { LEGEND_TYPES } from './widgets/legend/LegendWidgetUI';
+import LegendCategories from './widgets/legend/LegendCategories';
+import LegendIcon from './widgets/legend/LegendIcon';
+import LegendProportion from './widgets/legend/LegendProportion';
+import LegendRamp from './widgets/legend/LegendRamp';
 import ScatterPlotWidgetUI from './widgets/ScatterPlotWidgetUI';
 import TimeSeriesWidgetUI from './widgets/TimeSeriesWidgetUI/TimeSeriesWidgetUI';
 import {
@@ -35,6 +39,10 @@ export {
   TableWidgetUI,
   LegendWidgetUI,
   LEGEND_TYPES,
-  NoDataAlert
+  NoDataAlert,
+  // undocumented components
+  LegendCategories as _LegendCategories,
+  LegendIcon as _LegendIcon,
+  LegendProportion as _LegendProportion,
+  LegendRamp as _LegendRamp
 };
-
--- a/packages/react-ui/src/widgets/legend/LegendCategories.js
+++ b/packages/react-ui/src/widgets/legend/LegendCategories.js
@@ -7,12 +7,18 @@ export default function LegendCategories({ legend }) {
     return null;
   }
 
-  const { labels = [], colors = [] } = legend;
+  const { labels = [], colors = [], isStrokeColor = false } = legend;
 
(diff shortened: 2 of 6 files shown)


packages/react-ui/src/index.d.ts
```diff
@@ -35,6 +40,10 @@ export {
   TableWidgetUI,
   LegendWidgetUI,
   LEGEND_TYPES,
+  LegendCategories,
+  LegendIcon,
+  LegendProportion,
+  LegendRamp,
+  LegendWrapper,
```
VictorVelarde, line 47: Until we decide about 'public' api, we should export 'undocument' / internal methods with underscore, as eg `_LegendCategories`

packages/react-ui/src/index.d.ts
```diff
@@ -35,6 +39,10 @@ export {
   TableWidgetUI,
   LegendWidgetUI,
   LEGEND_TYPES,
-  NoDataAlert
+  NoDataAlert,
+  // undocumented components
+  LegendCategories as _LegendCategories,
```
VictorVelarde, line 44: @bbecquet, @Clebal just commented me about this: "components can't be exported with _" (in the past all _ exported elements were functions, indeed). Have you tried that works? If not, we can just keep your initial proposal, without "_"

packages/react-ui/src/widgets/legend/LegendCategories.js
```diff
@@ -59,7 +66,7 @@ function Row({ label, isMax, color = '#000' }) {
           mr={1.5}
           component='span'
           className={classes.circle}
-          style={{ backgroundColor: color }}
+          style={isStrokeColor ? { borderColor: color } : { backgroundColor: color }}
```
VictorVelarde, line 69: 👍🏻

packages/react-ui/src/index.d.ts
bbecquet: @VictorVelarde I've tested it locally and it's ok. I'm renaming them on import anyway (`import { _LegendCategories as LegendCategories } from '@carto/react-ui'`, so it's normal it works.
But I've also tested using them directly in JSX and that works too… Maybe that's weird, as React component names are supposed to start with an uppercase letter, but in the actual render tree they keep the name of component itself, not the exported alias.
Capture d’écran de 2022-02-08 17-31-51

So I would say there is no problem, especially because I'll be the only one to use them for now, but maybe by convention it's not a good practice to export them with `_`?

What's your take on that @Clebal?

VictorVelarde: If "_" works, I think it's better to keep it for the moment, for consistency with other non-public api methods. I would go ahead with it, even if that forces to use an alias with importing (that also serves as a reminder to end users), but let's hear Sergio

Clebal: If an alias without "_" is used in the import, then the rule won't complain.
From my point of view: I wouldn't export it as internal, because in the near future, to accomplish some PS requests, I will document them and make them public. We can make them public now (avoiding a future breaking change in cloud-native), and document them before the publishing of the v1.2.

bbecquet: Ok, I removed the _ and updated the changelog.
Thanks for your input @Clebal

borja-munoz: I agree with Victor approach. In my opinion, we should keep it with the underscore and not document the components yet.

bbecquet: So, I've dropped the commit where I removed the `_`, they are back to being prefixed.
In any case I don't think that's such a big deal as it's so easy to change either way, let's move on with that :)

VictorVelarde: Let's go for the moment with "_".

VictorVelarde (approved):
🚀

PR #312, opened 2022-02-04
Support for tiles in WGS84

padawannn:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/206588/carto-for-react-integration-to-allow-snowflake-tilesets

Currently, functions that calculate tile features for widgets only can work with local tiles coordinates (values between 0 and 1) but with providers like snowflake tiles come in WGS84 format

## Type of change
- Feature


packages/react-api/src/hooks/useTileFeatures.js
```diff
@@ -111,5 +118,10 @@ export default function useTileFeatures({
     [stopAnyCompute]
   );
 
-  return [onViewportLoad, fetch];
+  const onDataLoad = useCallback(({ tiles: [tile] }) => {
+    const tilesFormat = new URLSearchParams(tile).get('format');
+    setTileFormat(tilesFormat || 'mvt');
```
VictorVelarde, line 123: An enum here for TileFormat ? TileFormat.MVT

packages/react-core/src/filters/tileFeatures.js
```diff
@@ -188,10 +194,13 @@ export function tileFeatures({ tiles, viewport, geometry, uniqueIdProperty }) {
     if (!clippedGeometryToIntersect) {
       continue;
     }
-    // Transform the geometry to intersect to tile coordinates [0..1]
+    // Transform the geometry to intersect to tile coordinates [0..1] if the tileFormat is MVT
```
VictorVelarde, line 197: We could probably add a comment with our assumptions here, in case they change in the future, eg (feel free to rewrite):

We assume that MVT tileFormat uses local coordinates, while in the case of 'geojson', the geometries are already in WGS84

packages/react-api/src/hooks/useTileFeatures.js
padawannn: Done

packages/react-core/src/filters/tileFeatures.js
padawannn: Done

PR #314, opened 2022-02-08
Export FilterTypes enum

Clebal:
In order to be used in TypeScript, FilterTypes has to be exported. Before this change, we had to use @ts-ignore everywhere we used it.


eamador (approved):
LGTM!

PR #321, opened 2022-02-10
Fix how to read format tiles param from Maps API

padawannn:
# Description

To read `formatTiles` param from Maps API we're using the URLSearchParams and it could fail if the `formatTiles` is the first param

## Type of change
- Fix


packages/react-api/src/hooks/useTileFeatures.js
```diff
@@ -120,7 +120,7 @@ export default function useTileFeatures({
   );
 
   const onDataLoad = useCallback(({ tiles: [tile] }) => {
-    const tilesFormat = new URLSearchParams(tile).get('formatTiles');
```
VictorVelarde, line 123: I didn't know this could fail on that circumstance

VictorVelarde (approved):
LGTM.

packages/react-api/src/hooks/useTileFeatures.js
padawannn: for example (new URLSearchParams('https://github.com?formatTiles=aaaa')).get('formatTiles')

PR #324, opened 2022-02-11
Add optional slider to modify layer opacity in Legend widget

Clebal:
Template branch for testing: https://github.com/CartoDB/carto-react-template/tree/feature/sc-208745/add-slider-to-modify-layer-opacity-in-legend

image

image

Story:
image

### DIFF
--- a/packages/react-ui/src/widgets/legend/LegendWidgetUI.js
+++ b/packages/react-ui/src/widgets/legend/LegendWidgetUI.js
@@ -49,7 +49,8 @@ function LegendWidgetUI({
   layers = [],
   collapsed,
   onCollapsedChange,
-  onChangeVisibility
+  onChangeVisibility,
+  onChangeOpacity,
 }) {
   const classes = useStyles();
   const isSingle = layers.length === 1;
@@ -61,7 +62,7 @@ function LegendWidgetUI({
         collapsed={collapsed}
         onCollapsedChange={onCollapsedChange}
       >
-        <LegendRows layers={layers} onChangeVisibility={onChangeVisibility} />
+        <LegendRows layers={layers} onChangeVisibility={onChangeVisibility} onChangeOpacity={onChangeOpacity} />
       </LegendContainer>
     </Box>
   );
@@ -77,7 +78,8 @@ LegendWidgetUI.propTypes = {
   layers: PropTypes.array,
   collapsed: PropTypes.bool,
   onCollapsedChange: PropTypes.func,
-  onChangeVisibility: PropTypes.func
+  onChangeVisibility: PropTypes.func,
+  onChangeOpacity: PropTypes.func
 };
 
 export default LegendWidgetUI;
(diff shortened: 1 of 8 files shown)


eamador (approved):
Nice one! 👌🏽

packages/react-ui/src/widgets/legend/LegendWrapper.js
```diff
@@ -150,3 +192,77 @@ function Header({
...
+              className={classes.input}
+              value={Math.round(opacity * 100)}
+              margin='dense'
+              onChange={handleTextFieldChange}
+              InputProps={{
+                step: 1,
+                min: 0,
+                max: 100,
+                type: 'number',
+                inputProps: { 'data-testid': 'opacity-slider' },
```
VictorVelarde, line 258: We always try to avoid using data-testid as selectors for tests. Can you try with others, following if possible https://testing-library.com/docs/queries/about/#priority ?

packages/react-ui/src/widgets/legend/LegendWidgetUI.js
```diff
@@ -61,7 +62,7 @@ function LegendWidgetUI({
         collapsed={collapsed}
         onCollapsedChange={onCollapsedChange}
       >
-        <LegendRows layers={layers} onChangeVisibility={onChangeVisibility} />
+        <LegendRows layers={layers} onChangeVisibility={onChangeVisibility} onChangeOpacity={onChangeOpacity} />
```
VictorVelarde, line 65: I'm getting here "'LegendRows' cannot be used as a JSX component." warning in VS Code, because it looks like it returns an array, should we just add a fragment there?

packages/react-ui/src/widgets/legend/LegendWrapper.js
```diff
@@ -52,6 +70,10 @@ export default function LegendWrapper({
     if (onChangeVisibility) onChangeVisibility({ id, visible: !visible });
   };
 
+  const handleToggleLayerOptions = () => {
```
VictorVelarde, line 73: @Clebal Is this required to this PR?

packages/react-ui/src/widgets/legend/LegendWrapper.js
```diff
@@ -150,3 +192,77 @@ function Header({
...
+        margin: 0
+      }
+    },
+    '& .MuiInputAdornment-positionEnd': {
+      margin: 0
+    }
+  }
+}));
+
+function OpacityControl({ opacity, onChangeOpacity }) {
```
VictorVelarde, line 225: I know this was already like this, all functions inside same file, but I think it will probably make sense to create a folder, and a separated file per component, what do you think?

VictorVelarde:
Minor comments / questions on code, but looks like a nice addition @Clebal 👍🏻 

Also, could you please check behaviour when legend title is looooong ?
image

And add a changelog entry

packages/react-ui/src/widgets/legend/LegendWrapper.js
Clebal: Yes, when layer options is expanded, the opacity control is visible.

packages/react-ui/src/widgets/legend/LegendWrapper.js
Clebal: I agree. I'll modify it.

VictorVelarde (approved):
👍🏻

PR #327, opened 2022-02-14
Dynamic tiling integration

padawannn:
# Description

Adapt `useCartoLayerProps` to work with dynamic tiling. Due to dynamic tiling integration, we cannot know if we're going to load a geojson or a tileset until onDataLoad be executed so we always going to return the binary, onViewportLoad, and fetch properties that will be ignored for NOT Tileset layer

## Type of change
- Feature


VictorVelarde (approved):
LGTM.

PR #331, opened 2022-02-15
Implement stringSearch filter

Clebal:
Implemented following [this RFC](https://app.shortcut.com/cartoteam/story/203833/rfc-implement-search-filter).

Example of use:

```javascript
dispatch(
    addFilter({
        id: '...',
        column: 'certainColumn',
        type: FilterTypes.STRING_SEARCH,
        values: ['example'],
        owner: id
    })
)
```

Example of results:

_First is the filterValue and the second one is the featureValue_
```
stringSearch('á$^s', 'a$^sd') // true
stringSearch('manol', 'Manolo') // true
stringSearch('olo', 'Manólo') // true
stringSearch('Àlcälá dê Guadaí', 'Alcala de Guadaira') // true
stringSearch('1234', '4321') // false
```

Available params (all false by default):

```
caseSensitive
mustStart (the featureValue must start with filterValue)
mustEnd (the featureValue must end with filterValue)
keepSpecialCharacters (normalize fn isn't executed)
```

I've included in this PR, to test correctly the stringSearch, a new file called FilterTypes that have the filter fns, FilterTypes and filterFunctions map. Also, I've included a `FilterTypes.test.js` to test each filter separately instead of all together like in `Filter.test.js`.

 [sc-209082]


packages/react-core/src/filters/FilterTypes.js
```diff
@@ -0,0 +1,74 @@
...
+  });
+}
+
+// Aux
+function escapeRegExp(value) {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+function normalize(data, params) {
+  let normalizedData = '' + data;
```
zbigg, line 69: minor, i think that String(data) is more proper constructor nowadays

packages/react-core/src/filters/Filter.js
```diff
@@ -51,7 +16,11 @@ function passesFilter(columns, filters, feature) {
         throw new Error(`"${filter}" filter is not implemented.`);
       }
 
-      return filterFunction(columnFilters[filter].values, feature[column]);
+      return filterFunction(
```
zbigg, line 19: I am worried about performance of this search, where filter `stringSearch` function creates regexp for _every_ separate feature.

This flow should be somehow reversed, so all "filter functions" are have change to init before actually processing all features.

Proposal, change signature of filter functions from `Record<string, FilterFunction>` to `Record<string, (fiterValues, params) => (value) => boolean)>`

IOW, filterFunction should contain factories of filters and they should be called before loop.

packages/react-core/src/filters/FilterTypes.d.ts
```diff
@@ -0,0 +1,16 @@
+export enum FilterTypes {
+  IN = 'in',
+  BETWEEN = 'between', // [a, b] both are included
+  CLOSED_OPEN = 'closed_open', // [a, b) a is included, b is not
+  TIME = 'time',
+  STRING_SEARCH = 'stringSearch'
+}
+
+export const filterFunctions: Record<
```
zbigg, line 9: UppperCamelCase for types please

packages/react-core/src/filters/FilterTypes.d.ts
```diff
@@ -0,0 +1,16 @@
...
+  string,
+  (
+    filterValues: unknown[],
+    featureValue: unknown,
+    params?: Record<string, unknown>
+  ) => boolean
```
zbigg, line 15: As in previous comment, i would propose

```
type FilterFunction = (featureValue: unknown) => boolean
type FilterFunctionGen (filterValues: unknown[], params?: Record<string, unknown>) => FilterFunction
type FilterFunction = Record<string, FilterFunctionGen>
```

packages/react-core/src/filters/FilterTypes.js
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  return makeIntervalComplete(filterValues).some(checkRange);
+}
+
+// FilterTypes.STRING_SEARCH
+function stringSearch(filterValues, featureValue, params = {}) {
+  const normalizedFeatureValue = normalize(featureValue, params);
+
+  return filterValues.some((filterValue) => {
```
zbigg, line 52: if we have many filter values i would still create one regex by joining with `|`

packages/react-core/src/filters/FilterTypes.d.ts
Clebal: But... it's not a type. It's a constant.

packages/react-core/src/filters/FilterTypes.js
Clebal: Why?

packages/react-core/src/filters/FilterTypes.d.ts
Clebal: But that would be a breaking change. With this line 9, what I did is typing the exported map object `filterFunctions` that's used internally in Filter.js. That's all.

packages/react-core/src/filters/Filter.js
Clebal: @zbigg thanks for this very interesting comment. What you comment isn't small change, it requires changes in many parts of the app and it isn't the goal of this PR. But, I encourage you to open a RFC for further discussion about a possible refactor of filters.

About your performance worries:
I've been testing the performance of my code and I made some changes to improve it:

image

In this _dirty_ benchmark, I execute the code to create the filterValue regexp 1M times.
My original code execution time was nearly 400ms.
Avoiding recreating those static regexp (specialRegExp and normalizedRegExp), avoiding to always execute normalize (this can reduce execution time by half, but it's enabled by default) and using toLowerCase instead of toLocaleLowerCase, the execution time is ~140ms (without normalize, it can be ~70ms).

140ms isn't perfect, and I'm sure that rethinking how the filters are made these numbers would be better, but I think that maybe it's the best we can achieve with what we have right now.

packages/react-core/src/filters/FilterTypes.js
Clebal: I'll take a look 👌🏻

Clebal: It works very well. Thanks @zbigg 🚀

packages/react-core/src/filters/FilterTypes.d.ts
zbigg: Sorry for original comment, misread. 
Agree with @Clebal that changing enum names would be unintended breaking change.

packages/react-core/src/filters/Filter.js
zbigg: Good improvements on static regexps.

> it requires changes in many parts of the app and it
I would argue that this is not big or breaking change. `filterFunctions` is used twice (`passesFilter` and `passesFilterUsingBinary`) in very similar way iterate over filter columns, get filter function and with well proper helper

>  isn't the goal of this PR

If we add new feature and it requires (slight IMO) internal architecture change, it's good place to introduce it.

packages/react-core/src/filters/FilterTypes.js
Clebal: I feel it's a matter of taste haha, I feel confortable with that way, and it's wide known way to cast something to string, but yes, `String(data)` is more explicit. I don't care about what to use, tbh. But in this case, I choose `'' + data` because it's more performant.

packages/react-core/src/filters/Filter.js
Clebal: Why does it require it? Your proposal is good and yes, maybe there are ways to achieve it smaller that I thought, but from my knowledge contributing in this lib, I should keep the PRs as simple and as focused in the task as possible. In this PR I already added more that the strictly necessary adding new tests for filters. I don't feel that adding also that is a good idea, because the current proposal doesn't add a performance problem (IMO) and your proposal is an improvement that have to be done in its own task: if we have a problem with stringSearch, we should revert it without affecting your idea and the same about your proposal. Also, when making your idea, we can make changes in other filters, not only stringSearch.

In this case, let @VictorVelarde decides what to do here 😆  I don't think we will reach an agreement.

VictorVelarde: @Clebal We have talked about this and we think that this can be landed as it is, but it should be improved in the future (especially if you find, like Z suspects, performance issues in your UI). 

Please Sergio, give us feedback about your use of this in templates, once you integrate it

zbigg (approved):
LGTM after changes, agreed that potential performance improvements will be in followup PR

packages/react-core/src/filters/Filter.js
Clebal: https://user-images.githubusercontent.com/9151432/154667811-67495003-fe79-4662-9832-d5849dc28cdd.mov


I've done some tests in Site Selection, where I need this filter and it's works really great. The search has a minor debounce of 100ms, but the filter process is almost instant.

Clebal: I've been testing it in Site Selection and it's quite instant with 23K features, filtering in worker and in layer.

borja-munoz (approved):
LGTM!

VictorVelarde (approved):
LGTM. Nice addition, 👍🏻

PR #332, opened 2022-02-15
fix duplicated logic for category selection in PieWidgetUI

juandjara:
# Description

This PR fixes the behaviour of PieWidgetUI `clickEvent`, removing duplicated logic that was re-appliying UI changes already applied in other `useEffect` calls. Before this changes, you couldn't recieve the `onSelectedCategoriesChange` event without affecting the UI of the chart. This change makes it so you can better control if the component is controlled or uncontrolled. I suspect something similar is happening with `HistogramWidgetUI` but I have not tested that yet. 

## Type of change
- Fix

# Acceptance
Acceptance is described here in terms of the component API

1. create a constant array with `const categories = useRef([])`
2. Render a `<PieWidgetUI />`
3. Pass as props `selectedCategories={categories.current}` and `onSelectedCategoriesChange={(value) => console.log(value)}`
4. When clicked, the UI of the chart should not be modified (the clicked part should not be highlighted)

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation


Clebal (approved):
🚀

PR #337, opened 2022-02-16
Set expanded/collapsed value of each legend row

Clebal:
Added `collapsed` property in layerConfig legend object to manage the collapsed state of the layer's legend.

```javascript
const layerConfig = {
  title: 'Store types',
  visible: true,
  showOpacityControl: true,
  opacity: 0.4,
  legend: {
    attr: 'storetype',
    collapsed: true, // false by default
    type: LEGEND_TYPES.CATEGORY,
    labels: DATA.map((data) => data.label),
    colors: DATA.map((data) => data.color),
  },
};
```

By this way, you can stay informed about what is the state of the legend and indicate the initial value.
Test added for this feature.

### DIFF
--- a/packages/react-redux/src/slices/cartoSlice.js
+++ b/packages/react-redux/src/slices/cartoSlice.js
@@ -78,9 +78,16 @@ export const createCartoSlice = (initialState) => {
       updateLayer: (state, action) => {
         const layer = state.layers[action.payload.id];
         if (layer)
+          // TODO: Study if we should use a deepmerge fn
           state.layers[action.payload.id] = {
             ...layer,
-            ...action.payload.layerAttributes
+            ...action.payload.layerAttributes,
+            ...(layer.legend && {
+              legend: {
+                ...layer.legend,
+                ...action.payload.layerAttributes.legend
+              }
+            })
           };
       },
       removeLayer: (state, action) => {
--- a/packages/react-widgets/src/widgets/LegendWidget.js
+++ b/packages/react-widgets/src/widgets/LegendWidget.js
@@ -39,17 +39,13 @@ function LegendWidget({ className, initialCollapsed }) {
     );
   };
 
-  const handleChangeCollapsed = ({ id, collapsed }) => {
-    if (id) {
-      dispatch(
-        updateLayer({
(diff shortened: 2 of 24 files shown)


packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -78,9 +78,16 @@ export const createCartoSlice = (initialState) => {
       updateLayer: (state, action) => {
         const layer = state.layers[action.payload.id];
         if (layer)
+          // TODO: Study if we should use a deepmerge fn
           state.layers[action.payload.id] = {
             ...layer,
-            ...action.payload.layerAttributes
+            ...action.payload.layerAttributes,
+            ...(layer.legend && {
```
zbigg, line 85: i think state, we're unable to _add_ legent unless one already exists, not sure if that was intended
looks more like it should be:

```
(layer.legend || action.payload.layerAttributes.legend) &&  { legend: { ... }}
```

this version on the other hand prevents from removing legend at all

packages/react-widgets/src/widgets/LegendWidget.js
```diff
@@ -39,12 +39,22 @@ function LegendWidget({ className, initialCollapsed }) {
...
+    );
+  };
+
   return (
     <LegendWidgetUI
       className={className}
       layers={layers}
       onChangeVisibility={handleChangeVisibility}
       onChangeOpacity={handleChangeOpacity}
+      onChangeLegendCollapsed={handleChangeLegendCollapsed}
```
zbigg, line 57: There is already `onCollapsedChange` passed to children, why not reuse it and call dispatch(updateLayer) only as effect response to actually local `collapsed` variable controlled by `setCollapsed` ?

packages/react-redux/src/slices/cartoSlice.js
Clebal: If legend doesn't exist, it's added when `action.payload.attributes` is merged.

zbigg (approved):
LGTM, thanks!!!

VictorVelarde (approved):
👍🏻

PR #325, opened 2022-02-14
Add customLegendTypes prop to customise covered legend types

Clebal:
Added `customLegendTypes` prop in LegendWidget and LegendWidgetUI to customise legend types.

Example:

```javascript
const CUSTOM_LEGEND_TYPES = {
      'my-custom-legend': MyCustomLegend
}

function App {
      return <LegendWidget customLegendTypes={CUSTOM_LEGEND_TYPES} />
}

function MyCustomLegend () {
      return <p>Test</p>
}

//

dispatch(
  updateLayer({
    id: '...',
    layerAttributes: {
      legend: {
        type: 'my-custom-legend',
        ...
      }
    }
  })
)
```

Result:

image

PR #342, opened 2022-02-17
Fix legend render with only visibility toggle

juandjara:
# Description

With the implementation of the latest changes for legend custom types (#325), you can no longer define a `legend` in a redux layer definition that has no content and only a visibility toggle (pic. 1).

Prior to this change, the recommended thing to was to define something like `legend: {}` in order to show a visibility toggle only for the associated layer. But now, when doing that, React crashes because it tries to render `undefined`. This change makes it so you can still have a legend with only the visibility toggle

## Type of change
- Fix

# Acceptance

1. somewhere in you code, where you define the legend data, launch this action
```
dispatch(
  updateLayer(
    id: MY_LAYER_ID,
    layerAttributes: { legend: {} }
  )
)
```

2. Observe React crashing in the console with a message similar to `cannot render undefined`

Pic. 1 (From storybook)
Captura de pantalla 2022-02-17 a las 20 47 18

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Not released
 
+- Set expanded/collapsed value of each legend row [#337](https://github.com/CartoDB/carto-react/pull/337)
+
 ## 1.2
 
 ### 1.2.1-beta.6 (2022-02-16)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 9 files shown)


juandjara:
Thanks for your feedback @Clebal !!
I think everything is ready now, I added a new test case for the legend to check this kind of behaviour

PR #345, opened 2022-02-18
fix duplicated logic for category selection in HistogramWidgetUI

juandjara:
# Description

As with PR #332, this PR fixes the behaviour of HistogramWidgetUI `clickEvent`, removing duplicated logic that was re-appliying UI changes already applied in other `useEffect` calls. Before this changes, you couldn't recieve the `onSelectedBarsChange` event without affecting the UI of the chart. This change makes it so you can better control if the component is controlled or uncontrolled. 

## Type of change
- Fix

# Acceptance
Acceptance is described here in terms of the component API

1. create a constant array with `const categories = useRef([])`
2. Render a `<HistogramWidgetUI />`
3. Pass as props `selectedBars={categories.current}` and `onSelectedBarsChange={(value) => console.log(value)}`
4. When clicked, the UI of the chart should not be modified (the clicked part should not be highlighted)

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Not released
 
+- Fix duplicated logic for category selection in HistogramWidgetUI [#345](https://github.com/CartoDB/carto-react/pull/345)
+
 ## 1.2
 
 ### 1.2.1-beta.7 (2022-02-18)
--- a/packages/react-ui/src/widgets/HistogramWidgetUI.js
+++ b/packages/react-ui/src/widgets/HistogramWidgetUI.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import ReactEcharts from '../custom-components/echarts-for-react';
 import { Grid, Link, Typography, useTheme, makeStyles } from '@material-ui/core';
 import {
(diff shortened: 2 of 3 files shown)


Clebal (changes requested):
https://user-images.githubusercontent.com/9151432/154910948-eefbb546-aab4-44bd-a8f1-f1b54174a6ce.mov

I tested this PR using the template-sample-app-2 and it isn't working. Please, take a look.

juandjara:
Okay, finally I managed to get some time today to look at this. The thing was the `.filter` operation before the `.map` operation was changing the indexes to a weird something. Switching `map((d, i) => i)` with `map((d) => serie.data.indexOf(d))` fixes this.

Clebal:
https://user-images.githubusercontent.com/9151432/155129835-aec2a223-9a30-42e9-9985-dddf86340cf7.mov

It's working for every column except the last one. Can you take a look please?

Clebal (approved):

PR #346, opened 2022-02-21
Add groupByDate tests

Clebal:


packages/react-core/__tests__/operations/groupByDate.test.js
```diff
@@ -0,0 +1,149 @@
...
+        const RESULT = [
+          { name: Date.UTC(1970, 0, 1), value: 4 },
+          { name: Date.UTC(1971, 0, 1), value: 1 }
+        ];
+
+        executeGroupByDateFnTests(ARGS, RESULT);
+      });
+
+      test(GroupDateTypes.MONTHS, () => {
+        ARGS[3] = GroupDateTypes.MONTHS;
```
VictorVelarde, line 66: It would be probably easier to read if your ARGS is a dictionary, not array, and you do 

```
const params = { features, operation, date, grouping }
// ...
params.grouping = XXX
// ...
executeGroupByDateFnTests(params, RESULT);
```

VictorVelarde (approved):
LGTM

packages/react-core/__tests__/operations/groupByDate.test.js
Clebal: Much better! I changed it :)

PR #347, opened 2022-02-22
Use multiple operationColumn in widgets

Clebal:


packages/react-core/__tests__/operations/groupby.test.js
```diff
@@ -24,125 +25,91 @@ const COLUMN = 'test';
...
+      Object.values(AggregationTypes).forEach((operation, idx) => {
+        test(operation, () => {
+          const groups = groupValuesByColumn({
+            data: buildValidFeatures(COLUMN),
+            valuesColumns: [`${COLUMN}_quantitative`, `${COLUMN}_quantitative_2`],
+            joinOperation: AggregationTypes.SUM,
+            keysColumn: `${COLUMN}_qualitative`,
+            operation
+          });
+          expect(groups).toEqual(RESULTS_FOR_MULTIPLE[idx]);
```
bbecquet, line 99: This relies on `RESULTS_FOR_MULTIPLE` to be the same order as the result of `Object.values(AggregationTypes)`. 
But key "order" on objects is not strictly guaranteed, and anyway this isn't very easy to read.

Declaring `RESULTS_FOR_MULTIPLE` as an object with the types as keys would solve this.

packages/react-core/__tests__/operations/histogram.test.js
```diff
@@ -18,87 +19,91 @@ const buildInvalidFeatures = (columnName) => [
...
+    describe('one valuesColumns', () => {
+      Object.values(AggregationTypes).forEach((operation, idx) => {
+        test(operation, () => {
+          const groups = histogram({
+            data: buildValidFeatures(COLUMN),
+            valuesColumns: [COLUMN],
+            ticks,
+            operation
+          });
+          expect(groups).toEqual(RESULTS[idx]);
```
bbecquet, line 70: Same comment here, would be clearer and more robust with `RESULTS` as an object.

packages/react-core/src/operations/groupByDate.d.ts
```diff
@@ -1,9 +1,12 @@
...
+import { GroupDateTypes } from './constants/GroupDateTypes';
+import { AggregationTypes } from './constants/AggregationTypes';
 
-export function groupValuesByDateColumn(
-  data: [],
-  valuesColumn: string,
-  keysColumn: string,
-  operation: GroupDateTypes
-): GroupByFeature;
+export function groupValuesByDateColumn(args: {
```
bbecquet, line 5: :+1:

packages/react-core/__tests__/operations/groupby.test.js
Clebal: I like it. I'll change it. Thanks @bbecquet.

packages/react-core/src/operations/histogram.d.ts
```diff
@@ -1,9 +1,10 @@
...
 import { HistogramFeature } from '../types';
 
-export function histogram(
-  features: [],
-  columnName: string,
-  ticks: number[],
-  operation: AggregationTypes
-): HistogramFeature;
+export function histogram(args: {
+  data: Record<string, unknown>[];
```
padawannn, line 5: Rename features and columnName is not necessary, right? it's an unnecessary risk. I can see others variables rename in other files

padawannn (commented):
Without having done an in-depth review,  I'm worried because a lot of tests are changing in this PR so we cannot merge it until it is tested in builder and templates

packages/react-core/src/operations/histogram.d.ts
Clebal: No, it isn't mandatory. But since these fns are internals, I thought it was a good moment to have the same prop names in every file so they're easier to read.

Clebal:
@padawannn tests are the same but written in another way to avoid duplicated code, and to make them easier to test with multiple operation columns.

packages/react-core/__tests__/operations/groupby.test.js
Clebal: Changed.

bbecquet (approved):
LGTM. Thanks for the changes :+1:

Clebal:
Tested in builder and templates and it's working great.

packages/react-core/__tests__/operations/groupByDate.test.js
```diff
@@ -14,127 +14,141 @@ const DATES_VALUES = [
...
       expect(
-        groupValuesByDateColumn(
-          FEATURES,
-          OPERATION_COLUMN,
-          DATE_COLUMN,
-          '__fake_group_type__',
-          AggregationTypes.COUNT
-        )
+        groupValuesByDateColumn({
+          data: FEATURES,
```
VictorVelarde, line 34: Having homogeneous naming for fields in group-functions looks good to me. Out of curiosity: why 'data' and not 'features'?

Clebal: I think that data was used in two fns and that's why I chose that, but features is also ok. What do you prefer?

packages/react-core/__tests__/operations/groupby.test.js
```diff
@@ -24,125 +25,97 @@ const COLUMN = 'test';
...
-        {
-          name: 'Category 1',
-          value: 3
-        }
-      ]);
+    describe('one valuesColumns', () => {
+      Object.entries(RESULTS).forEach(([operation, result]) => {
+        test(operation, () => {
+          const groups = groupValuesByColumn({
+            data: buildValidFeatures(COLUMN),
```
VictorVelarde, line 71: If we always use same set of 'data' features for tests, using `buildValidFeatures(COLUMN)`... Doesn't it make sense to extract into a variable outside?

VictorVelarde: Indeed, if we always use same 'COLUMN' (='test'), why don't we simplify and have a simple array of features, without any function?

packages/react-core/__tests__/operations/groupby.test.js
```diff
@@ -1,12 +1,13 @@
 import { groupValuesByColumn } from '../../src/operations/groupby';
```
VictorVelarde, line 1: I would suggest renaming this test file to `groupBy.test.js` (not `groupby`)

Clebal: It's a problem to change letter case in macOS because it isn't case sensitive 🤣 I think I know how to solve it using multiple commits hmmm, let me find out tomorrow 👌

packages/react-core/src/operations/aggregation.js
```diff
@@ -0,0 +1,71 @@
...
+export const aggregationFunctions = {
+  [AggregationTypes.COUNT]: (values) => values.length,
+  [AggregationTypes.MIN]: min,
+  [AggregationTypes.MAX]: max,
+  [AggregationTypes.SUM]: sum,
+  [AggregationTypes.AVG]: avg
+};
+
+export function aggregate(feature, keys, operation) {
+  const normalizedKeys = Array.isArray(keys) ? keys : normalizeKeys(keys);
```
VictorVelarde, line 14: Do we need the previous check with `Array.isArray(keys)` ?

packages/react-core/src/operations/groupby.d.ts
```diff
@@ -1,9 +1,10 @@
-import { AggregationTypes } from './aggregation/AggregationTypes';
+import { AggregationTypes } from './constants/AggregationTypes';
```
VictorVelarde, line 1: Filename `groupBy.d.ts` ?

packages/react-core/src/operations/groupby.js
```diff
@@ -1,7 +1,13 @@
-import { AggregationTypes } from './aggregation/AggregationTypes';
-import { aggregationFunctions } from './aggregation/values';
-
-export function groupValuesByColumn(data, valuesColumn, keysColumn, operation) {
+import { AggregationTypes } from './constants/AggregationTypes';
```
VictorVelarde, line 1: Filename `groupBy.js` ?

packages/react-core/src/operations/aggregation.js
```diff
@@ -0,0 +1,71 @@
...
+
+export const aggregationFunctions = {
+  [AggregationTypes.COUNT]: (values) => values.length,
+  [AggregationTypes.MIN]: min,
+  [AggregationTypes.MAX]: max,
+  [AggregationTypes.SUM]: sum,
+  [AggregationTypes.AVG]: avg
+};
+
+export function aggregate(feature, keys, operation) {
```
VictorVelarde, line 13: 3 args already here for aggregate, have you considered using a single object param here, for consistency?

VictorVelarde (approved):
The overall approach looks good to me. I think all changes make sense, and open several new options for widgets.

My proposal would be to merge & make an alpha for further testing / QA in cloud-native, but we can discuss tomorrow in the meeting.

packages/react-core/__tests__/operations/groupby.test.js
Clebal: Done.

packages/react-core/src/operations/aggregation.js
Clebal: Nope, thanks 👌🏻

packages/react-core/src/operations/groupByDate.js
```diff
@@ -55,10 +56,12 @@ export function groupValuesByDateColumn(
         acc.set(groupKey, groupedValues);
       }
 
-      const isValid = item[valuesColumn] !== null && item[valuesColumn] !== undefined;
+      const tempValue = aggregate(item, valuesColumns, joinOperation);
```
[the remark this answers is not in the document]
Clebal: I like it. Modified.

packages/react-core/src/operations/groupby.d.ts
Clebal: Changed.

packages/react-core/src/operations/groupby.js
```diff
@@ -11,12 +17,14 @@ export function groupValuesByColumn(data, valuesColumn, keysColumn, operation) {
 
     accumulator[group] = accumulator[group] || [];
 
+    const tempValue = aggregate(item, valuesColumns, joinOperation);
```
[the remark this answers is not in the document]
Clebal: Changed.

packages/react-core/__tests__/operations/groupby.test.js
Clebal: Improved.

PR #350, opened 2022-02-23
Fix opacity selector input

Clebal:
Instead of using css selector with class name, it's better to use MUI classes prop. In SS we have a number suffix in the classes and the styles modification wasn't being applied.


zbigg (approved):
LGTM

PR #354, opened 2022-02-28
The basemap prop in the initialStateSlice for template base-3-typescript only accepts CARTO base maps names

zbigg:
Story details: https://app.shortcut.com/cartoteam/story/211802


VictorVelarde (approved):
👍🏻 
Add a changelog note, and we're ready

PR #355, opened 2022-02-28
Bump deck.gl to 8.7.0

zbigg:


bbecquet (approved):

VictorVelarde:
@zbigg & @bbecquet for effective bump we need to change also deps on every package, not just root level devDependencies (used for build / testing). Be careful in future bumps 👀

PR #358, opened 2022-03-01
Update tooling to support ts files

Clebal:


packages/react-api/package.json
```diff
@@ -18,15 +18,15 @@
     "lib": "src",
     "test": "__tests__"
   },
-  "types": "src/index.d.ts",
+  "types": "src/index.ts",
```
zbigg, line 21: It is in general this is discouraged to publish 'ts' files as typings
By publishing ts _sources_, you're forcing user to rely on "our" subset of typescript settings which arenot necessarily compatible with that of user project (like noImplicitAny, strictNullChecks...)

most configurations of typescript will not even compile them as they specifically exclude `node_modules`

thus, we still should publish `.d.ts` files, but generated using `tsc`

PR #361, opened 2022-03-21
Use multiple operationColumn in widgets

Clebal:
Documentation PR: https://github.com/CartoDB/documentation/pull/271

**Widgets affected**

- FormulaWidget
- CategoryWidget
- PieWidget
- ScatterPlotWidget
- TimeSeriesWidget

**Example from documentation**

In case you have the `population` disaggregated by gender: `population_m` and `population_f`, you can use an array in `operationColumn` and sum them into a single column using `joinOperation`. By this way, you avoid to have duplicated data in your dataset and you can compose widgets using multiple columns.
  
  ```js
  import { PieWidget } from "@carto/react-widgets";
  
  return (
    <PieWidget
      id="populationByContinent"
      title="Population by continent"
      dataSource="countriesSourceId"
      column="continent"
      operationColumn={["population_m", "population_f"]}
      joinOperation={AggregationTypes.SUM}
      operation={AggregationTypes.SUM}
    />
  );
  ```

### DIFF
--- a/packages/react-widgets/src/models/utils/checkParams.js
+++ b/packages/react-widgets/src/models/utils/checkParams.js
@@ -1,0 +1,7 @@
+export default function checkParams ({ joinOperation, operationColumn }) {
+    if (Array.isArray(operationColumn)) {
+        if (!joinOperation) {
+            throw new Error('Must indicate a join operation when providing multiple operation columns.')
+        }
+    }
+}
--- a/packages/react-widgets/src/models/utils/checkParams.js
+++ b/packages/react-widgets/src/models/utils/checkParams.js
@@ -1,7 +1,0 @@
-export default function checkParams ({ joinOperation, operationColumn }) {
-    if (Array.isArray(operationColumn)) {
-        if (!joinOperation) {
-            throw new Error('Must indicate a join operation when providing multiple operation columns.')
-        }
-    }
-}
--- a/packages/react-widgets/src/widgets/CategoryWidget.js
+++ b/packages/react-widgets/src/widgets/CategoryWidget.js
@@ -18,7 +18,8 @@ const EMPTY_ARRAY = [];
  * @param  {string} props.title - Title to show in the widget header.
  * @param  {string} props.dataSource - ID of the data source to get the data from.
  * @param  {string} props.column - Name of the data source's column to get the data from.
- * @param  {string} [props.operationColumn] - Name of the data source's column to operate with. If not defined it will default to the one defined in `column`.
+ * @param  {string | string[]} [props.operationColumn] - Name of the data source's column to operate with. If not defined it will default to the one defined in `column`. If multiples are provided, they will be merged into a single one using joinOperation property.
+ * @param  {AggregationTypes} [props.joinOperation] - Operation applied to aggregate multiple operation columns into a single one.
  * @param  {string} props.operation - Operation to apply to the operationColumn. Must be one of those defined in `AggregationTypes` object.
  * @param  {Function} [props.formatter] - Function to format each value returned.
  * @param  {Object} [props.labels] - Overwrite category labels.
@@ -36,6 +37,7 @@ function CategoryWidget(props) {
     dataSource,
     column,
     operationColumn,
+    joinOperation,
     operation,
     formatter,
     labels,
@@ -68,6 +70,7 @@ function CategoryWidget(props) {
       getCategories({
         column,
         operationColumn,
+        joinOperation,
         operation,
         filters,
         dataSource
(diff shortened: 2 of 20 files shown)


packages/react-widgets/src/widgets/ScatterPlotWidget.js
```diff
@@ -5,15 +5,18 @@ import { selectAreFeaturesReadyForSource } from '@carto/react-redux';
...
  * Renders a <ScatterPlotWidget /> component
  * @param  props
  * @param  {string} props.id - ID for the widget instance.
  * @param  {string} props.title - Title to show in the widget header.
  * @param  {string} props.dataSource - ID of the data source to get the data from.
- * @param  {string} props.xAxisColumn - Name of the data source's column to get the x axis from.
- * @param  {string} props.yAxisColumn - Name of the data source's column to get the y axis from.
+ * @param  {string | string[]} props.xAxisColumn - Name of the data source's column to get the x axis from. If multiples are provided, they will be merged into a single one using xAxisJoinOperation property.
+ * @param  {AggregationTypes} [props.xAxisJoinOperation] - Operation applied to aggregate multiple xAxis columns into a single one.
+ * @param  {string} props.yAxisColumn - Name of the data source's column to get the y axis from. If multiples are provided, they will be merged into a single one using yAxisJoinOperation property.
```
bbecquet, line 18: Type should be `{string | string[]}` like `xAxisColumn`.

packages/react-widgets/src/widgets/CategoryWidget.js
```diff
@@ -144,7 +148,11 @@ CategoryWidget.propTypes = {
   title: PropTypes.string.isRequired,
   dataSource: PropTypes.string.isRequired,
   column: PropTypes.string.isRequired,
-  operationColumn: PropTypes.string,
+  operationColumn: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.arrayOf(PropTypes.string)
+  ]),
+  joinOperation: PropTypes.oneOf(Object.values(AggregationTypes)),
```
bbecquet, line 155: We could use a custom PropTypes validator on this prop and the same on similar components, to ensure it's defined when the column prop is multiple.

Something like:
```jsx
const columnAggregationOn = (columnPropName) => (props, propName) => {
  if (Array.isArray(props[columnPropName])) {
    if (!props[propName]) {
      return new Error(`Prop ${propName} must be defined if ${columnPropName} is an array of column`);
    }
    if (!Object.values(AggregationTypes)) {
      return new Error(`Prop ${propName} must be a valid aggregation operator`);
    }
  }
};

// and in proptypes declaration: 
component.propTypes = {
    // ...
    operationColumn: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
    joinOperation: columnAggregationOn('operationColumn'),
    // ...
}
```

It could also share code with `checkParams` as it's doing basically the same thing, but PropType validator shouldn't throw.

packages/react-widgets/src/models/utils/checkParams.js
```diff
@@ -0,0 +1,7 @@
+export default function checkParams ({ joinOperation, operationColumn }) {
```
bbecquet, line 1: This name is too generic if it's only meant to test this particular kind of relation between params.

packages/react-widgets/src/widgets/CategoryWidget.js
Clebal: I really like it. Thanks!

Clebal: I love this solution. Already changed @bbecquet 👌🏻

packages/react-widgets/src/models/utils/checkParams.js
Clebal: Removed, it makes more sense to use the custom prop type validator instead of checking in the model.

bbecquet: @Clebal I don't know, prop type is only a warning for the dev but doesn't block in case of error. What happens with the components in case of invalid props? I guess if they don't crash everything that's ok :smile:

Clebal: Idk if the app should break. @borja-munoz what do you think? Should we break the app if joinOperation isn't defined? Or should we just show a warning in console?

borja-munoz: I think it should break. It should behave in the same way that it behaves if we miss a required prop (i.e. `dataSource`).

Clebal: Fixed.

padawannn (approved):
LGFM, I've tested in builder and it's working well

bbecquet (approved):

PR #363, opened 2022-03-23
Add filters logical operator prop to source object

juandjara:
# Description

Shortcut RFC: https://app.shortcut.com/cartoteam/story/216324/rfc-add-filter-condition-prop-to-source-object
Shortcut task: https://app.shortcut.com/cartoteam/story/218323/add-filters-logical-operator-prop-to-source-object

Summary of the change: Added a new `filtersLogicalOperator` param that is passed everywhere the `filters` is used. This lets us decide which filtering logical operator to use in the `passesFilter` function that is used by feature workers and by deck.gl layers.

## Type of change

- Feature

# Acceptance

Please describe how to validate the feature or fix:

* This feature can be validated with the new unit test included in the PR

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-core/src/filters/Filter.js
+++ b/packages/react-core/src/filters/Filter.js
@@ -1,11 +1,12 @@
 import { filterFunctions } from './FilterTypes';
 import { FiltersLogicalOperators } from '../operations/constants/FiltersLogicalOperators';
 
+const logicalOperatorMethods = {
+  [FiltersLogicalOperators.AND]: 'every',
+  [FiltersLogicalOperators.OR]: 'some'
+};
+
 function passesFilter(columns, filters, feature, filtersLogicalOperator) {
-  const logicalOperatorMethods = {
-    [FiltersLogicalOperators.AND]: 'every',
-    [FiltersLogicalOperators.OR]: 'some'
-  };
   const method = logicalOperatorMethods[filtersLogicalOperator];
   return columns[method]((column) => {
     const columnFilters = filters[column];
--- a/packages/react-core/src/filters/Filter.js
+++ b/packages/react-core/src/filters/Filter.js
@@ -1,13 +1,13 @@
 import { filterFunctions } from './FilterTypes';
 import { FiltersLogicalOperators } from '../operations/constants/FiltersLogicalOperators';
 
-const logicalOperatorMethods = {
+const LOGICAL_OPERATOR_METHODS = {
   [FiltersLogicalOperators.AND]: 'every',
   [FiltersLogicalOperators.OR]: 'some'
 };
 
 function passesFilter(columns, filters, feature, filtersLogicalOperator) {
-  const method = logicalOperatorMethods[filtersLogicalOperator];
+  const method = LOGICAL_OPERATOR_METHODS[filtersLogicalOperator];
   return columns[method]((column) => {
     const columnFilters = filters[column];
     const columnFilterTypes = Object.keys(columnFilters);
--- a/packages/react-core/src/filters/Filter.js
+++ b/packages/react-core/src/filters/Filter.js
@@ -1,7 +1,13 @@
 import { filterFunctions } from './FilterTypes';
+import { FiltersLogicalOperators } from '../operations/constants/FiltersLogicalOperators';
 
-function passesFilter(columns, filters, feature) {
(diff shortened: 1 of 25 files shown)


packages/react-core/src/filters/Filter.js
```diff
@@ -1,7 +1,13 @@
 import { filterFunctions } from './FilterTypes';
+import { FiltersLogicalOperators } from '../operations/constants/FiltersLogicalOperators';
 
-function passesFilter(columns, filters, feature) {
-  return columns.every((column) => {
+function passesFilter(columns, filters, feature, filtersLogicalOperator) {
+  const logicalOperatorMethods = {
+    [FiltersLogicalOperators.AND]: 'every',
+    [FiltersLogicalOperators.OR]: 'some'
+  };
```
Clebal, line 8: It's a constant, take it out of the function to avoid instantiate it every time.

Clebal (changes requested):
One comment in the code. Aside the code, I miss the changes in the widgets and models.

packages/react-core/src/filters/Filter.js
juandjara: Good catch !

juandjara:
> Aside the code, I miss the changes in the widgets and models.
Ok, I'm gonna check this too

juandjara:
@Clebal all changes introduced for widgets and models, with conflicts resolved

packages/react-core/src/filters/Filter.js
```diff
@@ -1,7 +1,14 @@
 import { filterFunctions } from './FilterTypes';
+import { FiltersLogicalOperators } from '../operations/constants/FiltersLogicalOperators';
 
-function passesFilter(columns, filters, feature) {
-  return columns.every((column) => {
+const logicalOperatorMethods = {
```
Clebal, line 4: Since it's a constant (in the pure meaning of the word), the convention says that it should be LOGICAL_OPERATOR_METHODS

Clebal (changes requested):

packages/react-core/src/filters/Filter.js
juandjara: Done 👍

packages/react-widgets/src/widgets/CategoryWidget.js
```diff
@@ -74,6 +75,7 @@ function CategoryWidget(props) {
         joinOperation,
         operation,
         filters,
+        filtersLogicalOperator: source?.filtersLogicalOperator,
```
bbecquet, line 78: Instead of doing it separately here, what about modifying the `useSourceFilters` hook to return also the operator?
IMO it makes sense to group that as both `filters` and `filtersLogicalOperator` are now required for the filters to be correctly applied.
It would also spare [this selector](https://github.com/CartoDB/carto-react/pull/363/files#diff-b6744b2ddd16d9987012b38f43d5800cd351e453fd471af46f6dbb649801e738R63) as it's already done in the hook.

(And the same comment applies for other widgets of course :slightly_smiling_face:)

juandjara: Ok, we can generalize it in there. Truth is I was searching for a place to get both props from but did not find any, so I'm ok with `useSourceFilters`. 

> It would also spare [this selector](https://github.com/CartoDB/carto-react/pull/363/files#diff-b6744b2ddd16d9987012b38f43d5800cd351e453fd471af46f6dbb649801e738R63) as it's already done in the hook.

I cannot see that link properly, (maybe my version of github is broken) don't know which file is it pointing to

bbecquet: > I cannot see that link properly, (maybe my version of github is broken) don't know which file is it pointing to

Sorry for the broken link, I was just pointing at some lines above when you retrieve the source by id:
```js
const source = useSelector((state) => selectSourceById(state, dataSource));
```

juandjara: Refactor done for `useSourceFilters`

bbecquet (approved):

PR #368, opened 2022-04-05
Add Global mode to FormulaWidget

Clebal:
Highlights:
- Pass source to model instead of only filters and data source id
- The widget update depends on spatial filter (if exists) when global is enabled. When global is disabled, it depends on isSourceReady.


Josmorsot (approved):
LGTM

packages/react-widgets/src/models/utils.js
```diff
@@ -0,0 +1,42 @@
...
+    return operationColumn;
+  }
+
+  const selectClauseFormatter = SELECT_CLAUSE_BY_JOIN_OPERATION[joinOperation];
+
+  if (!selectClauseFormatter) {
+    throw new Error(`${joinOperation} isn't valid.`);
+  }
+
+  return SELECT_CLAUSE_BY_JOIN_OPERATION[joinOperation](operationColumn);
```
bbecquet, line 41: I think you wanted to reuse `selectClauseFormatter` here.

packages/react-widgets/src/models/utils.js
```diff
@@ -0,0 +1,42 @@
+import { AggregationTypes, _filtersToSQL } from '@carto/react-core';
+
+export function formatTableNameWithFilters(props) {
```
bbecquet, line 3: I'm not completely familiar with the organization of the code between the different packages, but the functions declared in this file are quite low-level SQL manipulations. Wouldn't they be better stored in `react-core` alongside other SQL utils?

packages/react-widgets/src/widgets/FormulaWidget.js
```diff
@@ -19,6 +19,7 @@ import { columnAggregationOn } from './utils/propTypesFns';
  * @param  {AggregationTypes} props.operation - Operation to apply to the operationColumn. Must be one of those defined in `AggregationTypes` object.
  * @param  {Function} [props.formatter] - Function to format each value returned.
  * @param  {boolean} [props.animation] - Enable/disable widget animations on data updates. Enabled by default.
+ * @param  {boolean} [props.global] -
```
bbecquet, line 22: Missing doc. Better fill it now or that's the kind of thing that stays like that forever.

packages/react-widgets/src/widgets/FormulaWidget.js
```diff
@@ -31,28 +32,35 @@ function FormulaWidget({
...
+  const source = useWidgetSource({ dataSource, id });
 
   const [formulaData, setFormulaData] = useState(null);
   const [isLoading, setIsLoading] = useState(true);
 
+  const spatialFilter = useSelector((state) =>
+    global ? selectSpatialFilter(state, source?.id) : null
+  );
+
+  const isReady = global ? true : isSourceReady;
```
bbecquet, line 51: Maybe you can keep the same pattern as the selector above and put the ternary on `global` in the `isSourceReady`.
That way we don't need that many vars to track the readyness.

packages/react-widgets/src/models/utils.js
```diff
@@ -0,0 +1,42 @@
...
+  const { source, spatialFilter } = props;
+  const { data, filters, filtersLogicalOperator } = source;
+
+  const filtersClause = _filtersToSQL(filters, filtersLogicalOperator);
+
+  let query = `(${data} ${filtersClause}) foo`;
```
bbecquet, line 9: I bet we can find a better name than `foo` :)

packages/react-widgets/src/hooks/useWidgetSource.js
```diff
@@ -0,0 +1,29 @@
...
+  const rawSource = useSelector((state) => selectSourceById(state, dataSource));
+
+  const applicableFilters = useMemo(
+    () => getApplicableFilters(rawSource?.filters, id),
+    [rawSource?.filters, id]
+  );
+
+  return useCustomCompareMemo(
+    rawSource && {
+      ...rawSource,
```
VictorVelarde, line 24: Why the ...rawSource here? Do you need other props, beyond filters, to be considered in comparison?

packages/react-widgets/src/widgets/FormulaWidget.js
```diff
@@ -31,28 +32,35 @@ function FormulaWidget({
...
     selectAreFeaturesReadyForSource(state, dataSource)
   );
-  const { filters, filtersLogicalOperator } = useSourceFilters({ dataSource, id });
+  const source = useWidgetSource({ dataSource, id });
 
   const [formulaData, setFormulaData] = useState(null);
   const [isLoading, setIsLoading] = useState(true);
 
+  const spatialFilter = useSelector((state) =>
+    global ? selectSpatialFilter(state, source?.id) : null
```
VictorVelarde, line 48: Are we sure we want spatialFilter for global?

packages/react-widgets/src/models/utils.js
VictorVelarde: ... bar ?

packages/react-widgets/src/models/utils.js
VictorVelarde: But in this case Benjamin queries are very tied to widgets themselves, really not reusable in other contexts.

packages/react-widgets/src/models/utils.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// Operation columns & Join operation
+const SELECT_CLAUSE_BY_JOIN_OPERATION = {
+  [AggregationTypes.AVG]: (operationColumns) =>
+    `(${operationColumns.join(' + ')} / ${operationColumns.length})`,
+  [AggregationTypes.SUM]: (operationColumns) => `(${operationColumns.join(' + ')})`,
+  [AggregationTypes.COUNT]: (operationColumns) => operationColumns[0],
+  [AggregationTypes.MIN]: (operationColumns) => `(${operationColumns.join(' + ')})`,
+  [AggregationTypes.MAX]: (operationColumns) => `(${operationColumns.join(' + ')})`
```
VictorVelarde, line 27: I'm trying to think about 'examples' for a FormulaWidget using several columns and min or max... to understand better usecase, do you have any @Clebal ?

packages/react-widgets/src/models/utils.js
VictorVelarde: indeed

packages/react-widgets/src/widgets/FormulaWidget.js
VictorVelarde: This is my only 'big doubt' in PR, @Clebal

VictorVelarde (commented):
The overall approach looks good to me, :+1

Pieces I miss to move on: test coverage, some cleanup and probably then we're ready to go...

packages/react-widgets/src/models/utils.js
Clebal: Thanks!

packages/react-widgets/src/models/utils.js
Clebal: I agree with Victor, at least in a first iteration I think these kind of stuff should be with the widget.

packages/react-widgets/src/widgets/FormulaWidget.js
Clebal: Yes, I agree with you. I'm sorry that you are reviewing a very dirty code 😆 My idea here was the validation of the general approach from Victor, so I made this very quickly. I'll add this in the following commit

packages/react-widgets/src/widgets/FormulaWidget.js
Clebal: I like it hehe. Thanks.

packages/react-widgets/src/widgets/FormulaWidget.js
Clebal: I talked with Alberto, and he decided to remove spatial filter on global for now.

packages/react-widgets/src/models/utils.js
Clebal: Hmmm I don't have any real example tbh. I think that multiple columns doesn't make sense with some AggregationTypes like count or has limited use cases like in min/max.

I guess that I would want to have multiple columns in FormulaWidget to show the avg of min/max months of the stores.

```
<FormulaWidget
    title="Average revenue in the worst month"
    column={["revenue_jan", "revenue_feb", ...]}
    joinOperation={AggregationTypes.MIN}
    operation={AggregationTypes.AVG}
/>
```

Clebal:
@bbecquet @VictorVelarde thanks for all your comments.

I've already made the changes you requested me and I've also refactored the code to achieve what I had in mind.

Highlights: 

- Implement useWidgetFetch. **Why?** From my experience in C4R, I've seen that the widgets have a lot of common logic. With the global feature, that common logic will increase so we'd have to repeat in every widget. To avoid that, I've created a hook called `useWidgetFetch` that concentrate the common behaviors between widgets.

- Add useCustomCompareEffect. **Why?** useCustomCompareEffect is like our current useCustomCompareMemo to make deep comparisons between those hooks' dependencies. I'm not sure about this hook, but it was necessary to avoid reference problems when calling useWidgetFetch with the params. I could wrap the params in a useMemo, but I think that useCustomCompareEffect makes things easier.

- Implement wrapModelCall fn. **Why?** I created that wrapping function to avoid repeating the same logic about source restriction when implementing the global mode in every widget.

packages/react-widgets/src/widgets/FormulaWidget.js
VictorVelarde: makes sense

packages/react-widgets/src/models/utils.js
VictorVelarde: Thx for the example... But if that's the case, isn't the SELECT wrong using ' + ' that way @Clebal ?

VictorVelarde (commented):
@Clebal Approach looks good, so let's go for completeness (tests, readme, doc...). Also please consider my comment about MIN / MAX and please have some tests on it, to understand generated SQL better, with some examples on jest

packages/react-widgets/src/models/utils.js
Clebal: Yes, this is already changed.

New code is:
```
[AggregationTypes.MIN]: (operationColumns) => `LEAST(${operationColumns.join()})`,
[AggregationTypes.MAX]: (operationColumns) => `GREATEST(${operationColumns.join()})`
```

packages/react-widgets/__tests__/models/utils.test.js
```diff
@@ -0,0 +1,123 @@
...
+const V3_SOURCE = {
+  id: '__test__',
+  type: MAP_TYPES.TABLE,
+  data: '__test__',
+  credentials: {
+    apiVersion: API_VERSIONS.V3
+  }
+};
+
+const SOURCE_WITH_FILTES = {
```
VictorVelarde, line 27: typo
```suggestion
const SOURCE_WITH_FILTERS = {
```

packages/react-widgets/__tests__/models/utils.test.js
```diff
@@ -0,0 +1,123 @@
...
+    test('should format table sources correctly', () => {
+      const query = formatTableNameWithFilters({ source: V3_SOURCE });
+
+      expect(query).toBe(V3_SOURCE.data);
+    });
+
+    test('should format source with filters correctly', () => {
+      const query = formatTableNameWithFilters({ source: SOURCE_WITH_FILTES });
+
+      const whereClause = _filtersToSQL(SOURCE_WITH_FILTES.filters);
```
VictorVelarde: Nit: the call to this internal func `_filtersToSQL` is an implementation detail... I would compare query with a real, manually generated SQL string, so it's more clear

packages/react-widgets/src/models/utils.js
```diff
@@ -0,0 +1,64 @@
+import { AggregationTypes, _filtersToSQL } from '@carto/react-core';
+import { MAP_TYPES, API_VERSIONS } from '@deck.gl/carto';
+
+export function wrapModelCall(props, fromLocal, fromRemote) {
```
VictorVelarde, line 4: Nice idea. Maybe I would suggest a simpler name... like `getModel`. 

Also I suggest using ({props, fromLocal, fromRemote}) as params (in general, when more than one param is passed to a method)

Clebal: About the params, whatever you want but I prefer this way haha.
But I'd keep the name, getModel should return fromLocal or fromRemote, and I prefer the fn to make the call. Maybe `callModel` is ok for you?

VictorVelarde: don't worry, it's just something internal, let's keep your current one

packages/react-widgets/__tests__/models/utils.test.js
Clebal: Fixed.

VictorVelarde (approved):
LGTM. Let's just add changelog entry (and coordinate for an alpha)

PR #369, opened 2022-04-05
Fix regression for PieWidgetUI category selection logic

juandjara:
# Description

This PR is meant to bring back a bug fix that was performed in PR #332 but overwritten in PR #341 (both from version 1.2.1-beta.8)
The bug fix consisted in fixing the category selection logic for the PieWidgetUI component to avoid using `labels` in the selected category comparison and use the categories from `data` instead

## Type of change

- Fix

# Acceptance

Same as PR #332

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Bump deck.gl to 8.7.3 [#360](https://github.com/CartoDB/carto-react/pull/360)
 - Fix aggregation to ignore NULL values [#367](https://github.com/CartoDB/carto-react/pull/367)
+- Fix regression for PieWidgetUI category selection logic [#369](https://github.com/CartoDB/carto-react/pull/369)
 
 ## 1.3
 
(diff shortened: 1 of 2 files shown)


Clebal:
I think we don't have a way to test echarts yet :S

juandjara:
As Sergio says, I don't know how to test this UI related to echarts and would appreciate some guiding here, but I think the minimal test would be something like:

* Render a controlled `<PieWidgetUI />` with props `data`, `labels`, `selectedCategories` and `onSelectedCategoriesChange`.
* `selectedCategories` contain categories present in `data`, that are different from the labels present in `labels`
* When the `<PieWidgetUI />` is rendered, the appropiate categories should be selected in the widget
* Subsequent category selection events should read only categories from `data` and not from `labels` to make sure the data stays consistent

Clebal:
Currently, we don't have a way to test echarts, so I wouldn't worry about that because that should be handled using screenshots and that kind of stuff, and that's not easy.

Clebal (approved):

PR #370, opened 2022-04-07
Add Global mode to CategoryWidget & PieWidget

Clebal:


bbecquet (approved):
LGTM

packages/react-widgets/src/models/CategoryModel.js
```diff
@@ -1,22 +1,55 @@
...
+function buildSqlQueryToGetCategories(props) {
+  const { column, operation, operationColumn, joinOperation } = props;
+
+  const selectValueClause = `${operation}(${
+    operation === AggregationTypes.COUNT
+      ? '*'
+      : formatOperationColumn(operationColumn || column, joinOperation)
+  }) as value`;
+
+  return `SELECT COALESCE(${column}, 'null') as name, ${selectValueClause} FROM ${formatTableNameWithFilters(
```
Josmorsot, line 52: Just to confirm, what warehouses should support this feature?

Josmorsot (approved):
LGTM

packages/react-widgets/src/models/CategoryModel.js
Clebal: Any of them. Do you know any incompatibility here?

packages/react-widgets/src/hooks/useWidgetFetch.js
```diff
@@ -10,7 +10,7 @@ export default function useWidgetFetch(
   { id, dataSource, params, global, onError }
 ) {
   // State
-  const [data, setData] = useState(null);
+  const [data, setData] = useState();
```
VictorVelarde, line 13: Why this change here Sergio?

packages/react-widgets/src/models/CategoryModel.js
VictorVelarde: Nop, but it would be nice to involve Data team in these queries verification. Can you gather them (not just this one) and pass them to @Jesus89 please?

VictorVelarde (approved):
LGTM. Let's coordinate to see how we can merge (& test) a group of several widgets together. Most probably: Formula, Category and Histogram first

packages/react-widgets/src/hooks/useWidgetFetch.js
Clebal: https://github.com/CartoDB/carto-react/pull/370/files#diff-b6744b2ddd16d9987012b38f43d5800cd351e453fd471af46f6dbb649801e738R59

As you can see in that line, we define a default value. A default value can only be assigned in destructuring when it's undefined, not null.

PR #377, opened 2022-04-20
Add Global mode to TimeSeriesWidget

Clebal:


packages/react-core/src/operations/constants/GroupDateTypes.d.ts
```diff
@@ -1,8 +1,8 @@
 export enum GroupDateTypes {
-  YEARS = 'years',
-  MONTHS = 'months',
-  WEEKS = 'weeks',
-  DAYS = 'days',
-  HOURS = 'hours',
-  MINUTES = 'minutes'
+  YEARS = 'year',
```
VictorVelarde, line 2: Out of curiosity, why this change?

Clebal: Because in all the providers, these operations are singular :(

packages/react-widgets/src/models/TimeSeriesModel.js
```diff
@@ -1,22 +1,212 @@
...
+
+    return acc;
+  }, new Map());
+
+  return [...groups.entries()].map(([name, value]) => ({
+    name,
+    value:
+      value.filter((_, idx) => idx % 2 === 0).reduce((acc, [val]) => acc + val, 0) /
+      value.filter((_, idx) => idx % 2 !== 0).reduce((acc, [val]) => acc + val, 0)
+  }));
```
VictorVelarde, line 125: This final part is a bit obscure, can you comment (or use some better variable names)?

Clebal: I agree. I'm going to improve it now.

VictorVelarde (approved):
👍🏻

PR #380, opened 2022-04-21
Add callback prop to TableWidget to know when page size changed

bbecquet:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/219830/tablewidget-viewport-features-count-table-height

Add a function prop to TableWidget to be called with the new page size when it has been changed in the widget.

Capture d’écran de 2022-04-21 15-07-29

## Type of change

- Feature


packages/react-widgets/src/widgets/TableWidget.js
```diff
@@ -51,6 +53,12 @@ function TableWidget({
     setPage(0);
   }, [dataSource, isSourceReady, filters]);
 
+  useEffect(() => {
+    if (typeof onPageSizeChange === 'function') {
```
Clebal, line 57: Why checking the type? Isn't it enough to check if it's defined?

Clebal (approved):

PR #382, opened 2022-04-22
Normalize SQL API response due to providers inconsistency

Clebal:


packages/react-widgets/__tests__/models/utils.test.js
```diff
@@ -94,6 +95,24 @@ describe('utils', () => {
     });
   });
 
+  describe('normalizeObjectKeys', () => {
```
VictorVelarde, line 98: Unit testing the fn in utils itself is nice, but IMO this would be better understood how this is used in a higher level

Indeed, one thing that I miss in the PR is a bit more of context: this is a 'fix' but there is no linked shorcut story about what this is fixing exactly. We should always try to do that. In any case, I'm gonna approve it, but please take my comments into consideration

VictorVelarde (approved):
👍🏻

packages/react-widgets/__tests__/models/utils.test.js
VictorVelarde: As always, try to add a good one-line in the changelog

PR #384, opened 2022-04-27
Fix histogram bar selection

Clebal:
To fix https://app.shortcut.com/cartoteam/story/221511/ac_7xhfwyml-histogram-widget-on-click-second-to-last-column-selects-last-column


VictorVelarde (approved):
LGTM. Thx!

PR #385, opened 2022-04-27
Show warning when dropping features

Josmorsot:
# Description

Shortcut: [(link)](https://app.shortcut.com/cartoteam/story/222853/widgets-display-dropping-features-warning)

Display warning when tilesets are generated using dropping feature strategy

https://user-images.githubusercontent.com/7818205/165553901-a802728b-1b69-45b2-8b51-a63f33eca410.mov

## Type of change
- Feature

### DIFF
--- a/packages/react-api/src/hooks/useTileFeatures.js
+++ b/packages/react-api/src/hooks/useTileFeatures.js
@@ -132,7 +132,8 @@ export default function useTileFeatures({
   return [onDataLoad, onViewportLoad, fetch];
 }
 
-// TODO: Comentario => https://github.com/visgl/loaders.gl/pull/2128
+// WORKAROUND: To read headers and know if the tile is dropping features. 
+// Remove when the new loader is ready => https://github.com/visgl/loaders.gl/pull/2128
 const customFetch = async (url, {layer, loaders, loadOptions, signal}) => {
   loadOptions = loadOptions || layer.getLoadOptions();
   loaders = loaders || layer.props.loaders;
--- a/packages/react-api/src/hooks/useTileFeatures.js
+++ b/packages/react-api/src/hooks/useTileFeatures.js
@@ -111,9 +111,9 @@ export default function useTileFeatures({
 
       debounceIdRef.current = debouncedLoadTiles(tiles);
       const isDroppingFeatures = tiles?.some(tile => tile.content?.isDroppingFeatures)
-      dispatch(setIsDroppingFeatures({ id: source.id, isDroppingFeatures }))
+      dispatch(setIsDroppingFeatures({ id: sourceId, isDroppingFeatures }))
     },
-    [source, stopAnyCompute, setSourceFeaturesReady, debouncedLoadTiles, debounceIdRef, dispatch]
+    [sourceId, stopAnyCompute, setSourceFeaturesReady, debouncedLoadTiles, debounceIdRef, dispatch]
   );
 
   const fetch = useCallback(
--- a/packages/react-api/src/hooks/useTileFeatures.js
+++ b/packages/react-api/src/hooks/useTileFeatures.js
@@ -64,11 +64,14 @@ export default function useTileFeatures({
         return acc;
       }, []);
 
+      const isDroppingFeatures = tiles?.some(tile => tile.content?.isDroppingFeatures)
+      dispatch(setIsDroppingFeatures({ id: sourceId, isDroppingFeatures }))
+
       executeTask(sourceId, Methods.LOAD_TILES, { tiles: cleanedTiles })
         .then(() => setTilesetLoaded(true))
         .catch(throwError);
     },
-    [sourceId, setTilesetLoaded]
+    [sourceId, setTilesetLoaded, dispatch]
   );
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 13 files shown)


VictorVelarde (approved):
LGTM, but I miss some tests.
Also one line in the CHANGELOG describing this new feature

packages/react-widgets/src/widgets/FormulaWidget.js
```diff
@@ -48,12 +53,16 @@ function FormulaWidget({
...
   return (
     <WrapperWidgetUI title={title} isLoading={isLoading} {...wrapperProps}>
-      <FormulaWidgetUI
-        data={data}
-        formatter={formatter}
-        unitBefore={true}
-        animation={animation}
-      />
+      {isDroppingFeatures ? (
+        <NoDataAlert severity="warning" body="Some rows have been filtered at this zoom level. Zoom in to ensure you see all rows in the map." />
```
borja-munoz, line 57: I'm not sure about the body property and this should be configurable (for all the widgets). I think some rows have been "filtered out" and if you zoom in, you don't see all rows in the map, what you do is to ensure that no features in the current viewport have been filtered out. I'd also refer to features/elements instead of rows.

VictorVelarde: Copy from @ernesmb are here: https://app.shortcut.com/cartoteam/story/222812/design-widget-state-when-feature-dropping-is-happening. 

Let's validate those!

borja-munoz: I understand this is the copy that Ernesto wants for Builder and that's fine, but the copy itself needs to be configurable. We can have a default value and then have a way to override this default value by the applications, in the same way that NoDataAlert component works. I think the current implementation does not meet this requirement. I already mentioned that before this got implemented.

packages/react-api/src/hooks/useTileFeatures.js
```diff
@@ -107,14 +110,16 @@ export default function useTileFeatures({
       setSourceFeaturesReady(false);
 
       debounceIdRef.current = debouncedLoadTiles(tiles);
+      const isDroppingFeatures = tiles?.some(tile => tile.content?.isDroppingFeatures)
+      dispatch(setIsDroppingFeatures({ id: source.id, isDroppingFeatures }))
     },
-    [stopAnyCompute, setSourceFeaturesReady, debouncedLoadTiles, debounceIdRef]
+    [source, stopAnyCompute, setSourceFeaturesReady, debouncedLoadTiles, debounceIdRef, dispatch]
```
Clebal, line 116: In the line 30, you have sourceId variable with the `source.id`. Here you are using source as dependency when you're only using its id property, I don't think it's a good practice.
Also, I'd put this logic in loadTiles fn, I don't think it's good to set if it's dropping features before loading the tiles, so both are debounced.

packages/react-api/src/hooks/useTileFeatures.js
```diff
@@ -126,3 +131,14 @@ export default function useTileFeatures({
 
   return [onDataLoad, onViewportLoad, fetch];
 }
+
+// TODO: Comentario => https://github.com/visgl/loaders.gl/pull/2128
```
Clebal, line 135: I understand that this isn't a TODO, right? I mean, are you expecting this to change in the feature?
Btw, "comentario" ;)

packages/react-widgets/src/widgets/CategoryWidget.js
```diff
@@ -52,6 +52,8 @@ function CategoryWidget(props) {
   } = props;
   const dispatch = useDispatch();
 
+  const source = useSelector((state) => selectSourceById(state, dataSource))
+  const isDroppingFeatures = source?.isDroppingFeatures
```
Clebal, line 56: I would create a specific selector for dropping features property so we don't have to declare a variable with the full source that may change a lot. Maybe, you can put that logic inside useWidgetFetch where we are already have the source object. Take a look into that hook.

packages/react-widgets/src/widgets/FormulaWidget.js
Clebal: I agree with Borja. In others widgets, we have the noDataAlertProps but they are being overwritten when isDroppingFeatures is true:

```
<NoDataAlert {...noDataAlertProps} {...(isDroppingFeatures ? { severity: 'warning', body: 'Some rows have been filtered at this zoom level. Zoom in to ensure you see all rows in the map.' } : {})}/>
```

> I'd also refer to features/elements instead of rows.

This for sure!

Clebal (changes requested):
👋🏻  Good job @Josmorsot, I added some comments that I think will be very good for your code and the app performance. Ping me if you want to talk about them 👌🏻

VictorVelarde (approved):
👍🏻

Clebal (approved):
Vamos!

packages/react-ui/src/widgets/NoDataAlert.js
```diff
@@ -1,17 +1,33 @@
...
+  </Box>
+) : (
+  <Box mt={-1} />
+)}
 
 function NoDataAlert({
   title = 'No data available',
-  body = 'There are no results for the combination of filters applied to your data. Try tweaking your filters, or zoom and pan the map to adjust the Map View.'
+  body = 'There are no results for the combination of filters applied to your data. Try tweaking your filters, or zoom and pan the map to adjust the Map View.',
+  severity = undefined
```
bbecquet, line 17: The `severity` prop is never passed anywhere so this is removing the orange style on all existing alerts.

bbecquet (changes requested):

packages/react-ui/src/widgets/NoDataAlert.js
VictorVelarde: AFAIK this is desired. We want to change current 'severity', to make it less prominent (not just for dropping, but for current use case).

This was already discussed with Design & validated with product.

PR #387, opened 2022-04-28
GeocoderWidget support for LDS-based gecoding.

zbigg:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/225672/address-search-in-builder

Enchance GeocoderWidget so it can geocode using LDS and can be embedded in builder
 - accept credentials via props
 - for V3 credentials, use LDS service
 - accept `zoomControl` property for disabling default behavior of trying to control view (not needed or desired in builder)

Full backward compatibility for V3 credentials.

## Type of change

- Feature

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation


packages/react-api/src/api/SQL.js
```diff
@@ -100,3 +101,47 @@ function createRequest({ credentials, connection, query, opts = {} }) {
   });
   return postRequest(postUrl, rawParams, requestOpts);
 }
+
```
zbigg, line 104: This code is not common, i.e completly SQL-related so moved here.

packages/react-widgets/src/widgets/GeocoderWidget.js
```diff
@@ -53,12 +53,17 @@ const SearchIcon = (args) => (
  * @param  {object} props
  * @param  {Object} [props.className] - Material-UI withStyle class for styling
  * @param  {Function} [props.onError] - Function to handle error messages from the widget.
+ * @param  {Object=} [props.credentials] - Credentials to use, otherwise widgets checks redus state.
```
Clebal, line 56:
```suggestion
 * @param  {object} [props.credentials] - Credentials to use, otherwise widgets checks redux state.
```

packages/react-widgets/src/widgets/GeocoderWidget.js
```diff
@@ -53,12 +53,17 @@ const SearchIcon = (args) => (
...
+ * @param  {string=} props.method - method, either `sql` or `lds`, see geocodeStreetPoint
+ * @param  {Boolean=} [props.zoomControl] - Optional, default true. Whether control should zoom map on result.
```
Clebal, line 58: Why the =?

packages/react-widgets/src/widgets/GeocoderWidget.js
```diff
@@ -53,12 +53,17 @@ const SearchIcon = (args) => (
...
+ *
+ * @see geocodeStreetPoint
  */
 function GeocoderWidget(props) {
   const inputRef = useRef();
   const oauthCredentials = useSelector(selectOAuthCredentials);
   const globalCredentials = useSelector((state) => state.carto.credentials);
-  const credentials = oauthCredentials || globalCredentials;
+  const credentials = props.credentials || oauthCredentials || globalCredentials;
```
Clebal, line 66: Why do we need to add credentials and method to the props? I guessed that the implementation of this would be just check the globalCredentials apiVersion and if it's V3 it would use lds and if it isn't use sql. Idk if I'm missing something. Idk why we have oauthCredentials here ?¿ The global credentials is created using those credentials.

packages/react-widgets/src/widgets/GeocoderWidget.js
Clebal, line 58: When I read `zoomControl` I though that it would add a zoom control with `+` and `-` 😆  Maybe `zoomToResult: boolean` (like the fn) would be clearer?

zbigg: I've found that this is the [JSDoc syntax for optional parameters](https://stackoverflow.com/questions/5873927/how-to-indicate-param-is-optional-using-inline-jsdoc).

Thanks for this VSCode doersn't complain on missing params ...

packages/react-widgets/src/widgets/GeocoderWidget.js
zbigg: Actually, this is just proposition, not particularly tied to this approach.

I've later found that i can push those credentials through redux too, if that's problematic i can remove it and rely only on `carto.credentials` ...

packages/react-widgets/src/widgets/GeocoderWidget.js
Clebal: 😮 Great so

packages/react-widgets/src/models/GeocodingModel.js
```diff
@@ -1,15 +1,56 @@
...
+import { executeSQL, ldsGeocode } from '@carto/react-api';
+import { API_VERSIONS } from '@deck.gl/carto';
 
 /**
- * Street-Level Geocoder
- * @param {*} credentials
- * @param {*} { searchText, city, state, country } all optional but searchText
+ * Street-Level Geocoder.
  *
+ * Geocode street given as address and optionally county into geo location (latitude, longitute).
```
borja-munoz, line 9: Small typos here: 'county', 'longitute'

packages/react-widgets/src/widgets/GeocoderWidget.js
```diff
@@ -106,10 +111,13 @@ function GeocoderWidget(props) {
         setIsLoading(true);
         const result = await geocodeStreetPoint(credentials, {
           searchText,
+          method: props.method,
```
borja-munoz, line 114: Does the developer needs the set the method prop in the widget declaration?

I mean, do you need to do something like this?:

<GeocoderWidget method='lds'>

If possible, we should come up with a way that works without having to specify the method explicitly. I don't know how the solution could look like but we have an environment variable that we can check to see if we have a CARTO 2 or CARTO 3 application.

packages/react-widgets/src/widgets/GeocoderWidget.js
VictorVelarde: Aggree

packages/react-widgets/src/widgets/GeocoderWidget.js
zbigg: After brief discussion, we agreed with @Clebal that this component should use Redux `carto.credentials`.
We can set it using `setCredentials` action from `react-redux`.
It's cleaner and better aligned with rest of ecosystem.

Clebal (approved):
Great work @zbigg!

packages/react-widgets/src/widgets/GeocoderWidget.js
zbigg: @borja-munoz 
This was only proposition, as i thought it would good to be explicit.
In general, i agree so i've already removed this props from last patch.

As for now, we always 
 - use LDS for V3 credentials
 - use SQL for V2 credentials (old not changed behavior)

borja-munoz (approved):
LGTM! 🚀

packages/react-widgets/src/widgets/GeocoderWidget.js
VictorVelarde: Allright then, we'll see how / where that is set from cloud-native side

packages/react-widgets/src/widgets/GeocoderWidget.js
VictorVelarde: one param less to document, 👍🏻

VictorVelarde (approved):
LGTM, nice job

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Not released
 
+- GeocoderWidget support for LDS queries [#387](https://github.com/CartoDB/carto-react/pull/387)
```
VictorVelarde, line 5:
```suggestion
- GeocoderWidget support for LDS queries in CARTO 3 [#387](https://github.com/CartoDB/carto-react/pull/387)
```

PR #388, opened 2022-04-28
HistogramWidget/HistogramWidgetUI refactor

Clebal:
If you want to test it using builder, and these lines in HistogramWidgetUI:

```
  // TODO: JUST FOR BUILDER LINK
  theme.typography.charts = cartoThemeOptions.typography.charts;
  theme.palette.charts = cartoThemeOptions.palette.charts;
  theme.palette.secondary.main = cartoThemeOptions.palette.secondary.main;
  theme.palette.other = cartoThemeOptions.palette.other;
```

image
image
image

https://user-images.githubusercontent.com/9151432/165898163-3463e5b9-46fd-4326-89b5-7efbd63e50f0.mov


packages/react-ui/src/widgets/HistogramWidgetUI/HistogramWidgetUI.js
```diff
@@ -0,0 +1,352 @@
...
+  tooltip,
+  tooltipFormatter,
+  animation,
+  filterable: _filterable,
+  height
+}) {
+  const classes = useStyles();
+  const theme = useTheme();
+
+  const filterable = _filterable && !!onSelectedBarsChange;
```
VictorVelarde, line 48: Why is this so? I'm not sure I understand it well

packages/react-ui/src/widgets/HistogramWidgetUI/HistogramWidgetUI.js
```diff
@@ -0,0 +1,352 @@
...
+    }),
+    [theme, tooltip, tooltipFormatter, xAxisFormatter, yAxisFormatter]
+  );
+
+  // xAxis
+  const xAxisOptions = useMemo(
+    () => ({
+      min,
+      max,
+      interval: (max - min) / formattedData.length,
```
VictorVelarde, line 94: We always validate we have >0 here, right?

Clebal: We validate it in HistogramWidget, but not in HistogramWidgetUI

packages/react-ui/src/widgets/HistogramWidgetUI/HistogramWidgetUI.js
```diff
@@ -0,0 +1,352 @@
...
+        : theme.palette.secondary.main;
+
+      return { value: item, itemStyle: { color } };
+    });
+
+    return {
+      type: 'custom',
+      cursor: 'pointer',
+      markArea: markAreaOptions,
+      renderItem: function (params, api) {
```
VictorVelarde, line 188: Probably my lack of knowledge of some internals, but it's a bit hard to understand what value 0, 1 and 2 from api.value. I don't know if it exists at all, but it would be nice to have something like: `var [xxxxxx, xxxxxx, yValue] = api.values()  // or something similar` to have a good name per each one

bbecquet (approved):
:white_check_mark: and tested manually on builder, really great! Just some nitpicking suggestions.
Great that you separated interactivity in its own hook :+1:

packages/react-ui/src/widgets/HistogramWidgetUI/HistogramWidgetUI.js
Clebal: I copied this fn without knowing each details, but it's true, I'll improve this fn readability.

packages/react-core/__tests__/operations/histogram.test.js
```diff
@@ -29,80 +57,164 @@ describe('histogram', () => {
...
+      ).toEqual({});
     });
 
-    test('should return an array of bins with same length as ticks plus two, due to the addition of extreme values', () => {
-      const h = histogram({
-        data: VALID_DATA,
-        valuesColumns: [COLUMN],
-        ticks,
-        operation: AggregationTypes.COUNT
+    describe('should return an array of bins with same length as ticks plus one, due to the addition of extreme values', () => {
```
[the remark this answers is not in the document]
Clebal: Agree.

packages/react-core/src/operations/histogram.js
```diff
@@ -1,13 +1,36 @@
...
+  if (useBins) {
+    min = aggregationFunctions[AggregationTypes.MIN](data, valuesColumns, joinOperation);
+    max = aggregationFunctions[AggregationTypes.MAX](data, valuesColumns, joinOperation);
+
+    if (!isFinite(min) || !isFinite(max)) {
+      throw new Error('Cannot calculate histogram without valid data');
+    }
+
+    for (let i = 1; i < bins; i += 1) {
+      ticks.push(min + (max - min) * (i / bins));
```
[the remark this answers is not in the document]
Clebal: In order to this code being executed, the ticks should be empty, so it shouldn't be harmful, I found the code struggling trying to keep two ticks variables. Maybe if I copy the ticks array you feel a bit safer? :)

packages/react-ui/src/widgets/HistogramWidgetUI/HistogramWidgetUI.js
```diff
@@ -0,0 +1,352 @@
...
+        const isLast = params.dataIndex === formattedData.length - 1;
+        const isFirst = params.dataIndex === 0;
+        var yValue = api.value(2);
```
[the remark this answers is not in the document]
Clebal: Yes, sorry, I copied this code, I didn't realize that

packages/react-ui/src/widgets/HistogramWidgetUI/useHistogramInteractivity.js
```diff
@@ -0,0 +1,125 @@
...
+        onSelectedBarsChange(
+          newSelectedBars.size === data.length ? [] : Array.from(newSelectedBars)
+        );
+      }
+    },
+    [onSelectedBarsChange, selectedBars, data.length]
+  );
+
+  return {
+    markAreaOptions: markArea.length === 2 && {
```
VictorVelarde, line 104: why markArea.length === 2 ... what does it mean?

Clebal: markArea is the area created to filter the columns. Only when there are two items it's an area with a start and an end

packages/react-ui/src/widgets/HistogramWidgetUI/HistogramWidgetUI.js
Clebal: I checked that even if formattedData.length is 0 nothing happens.

VictorVelarde (approved):
Excellent job here!. Just some questions

packages/react-core/src/operations/histogram.js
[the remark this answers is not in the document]
Clebal: Implemented shallow copy of ticks array to make the code safer

PR #391, opened 2022-04-29
Simplify 'no data' implementation on FormulaWidget

bbecquet:
# Description

Follow-up to https://github.com/CartoDB/carto-react/pull/389 which was merged too quickly.

- Removes the need for `feature_count`, just return `null` as computed value in the case of no data present (except for COUNT which will return 0). But as `null` is considered in `useWidgetFetch` as a loading state, we still need to return an object and not just a scalar.
- Remove the explicit passing of an empty placeholder prop, as `FormulaWidgetUI` already supports this when no data is provided.

## Type of change

- Fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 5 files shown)


Clebal (approved):
Great job Benjamin 🚀

PR #394, opened 2022-04-29
Use alias in query sources

Clebal:


Josmorsot (approved):

packages/react-widgets/__tests__/models/utils.test.js
```diff
@@ -77,7 +77,7 @@ describe('utils', () => {
       };
       const query = formatTableNameWithFilters({ source });
 
-      expect(query).toBe(`(${source.data.replace(';', '')})`);
+      expect(query).toBe(`(SELECT * FROM test) foo`);
```
VictorVelarde, line 80: Probably 'foo' here is not the best name... @Clebal, don't you think? `queryAlias`, as in the PR title would be better

Clebal: queryAlias is too long. foo is a classic label for something that isn't useful like in this case. I would even use only `q`.

VictorVelarde: I know... 😄 
image
https://www.genbeta.com/desarrollo/rfc-3092-el-curioso-origen-de-las-variables-foo-y-bar

Clebal: Now that I know that, I like it more hahaha.

packages/react-widgets/src/models/utils.js
```diff
@@ -27,7 +27,7 @@ export function formatTableNameWithFilters(props) {
   const whereClause = _filtersToSQL(filters, filtersLogicalOperator);
 
   const formattedSourceData =
-    source.type === MAP_TYPES.QUERY ? `(${data.replace(';', '')})` : data;
+    source.type === MAP_TYPES.QUERY ? `(${data.replace(';', '')}) foo` : data;
```
borja-munoz, line 30: I'd use another alias for the subquery. If I read a query like this, I'm going to think it's like a test.

Clebal: The alias won't be used, so I don't want the user to think that the alias is useful.

PR #397, opened 2022-05-03
Fix HistogramWidgetUI min/max placement

Clebal:
When using decimals, the placement wasn't being done correctly.

Before:

image

After:

image


eamador (approved):
nice one 🚀

PR #398, opened 2022-05-04
Pass layer obj to LegendComponent

Clebal:
In order to know which is the layer that is rendering the legend we need to pass layer obj as props. It's very very useful when rendering custom legends.

### DIFF
--- a/packages/react-ui/src/widgets/legend/LegendWidgetUI.js
+++ b/packages/react-ui/src/widgets/legend/LegendWidgetUI.js
@@ -189,7 +189,10 @@ function LegendRows({
     <>
       {layers.map(
         (
-          {
+          layer,
+          index
+        ) => {
+          const {
             id,
             title,
             switchable,
@@ -197,9 +200,8 @@ function LegendRows({
             showOpacityControl = false,
             opacity = 1,
             legend = {}
-          },
-          index
-        ) => {
+          } = layer;
+
           const {
             type = LEGEND_TYPES.CUSTOM,
             collapsible = true,
@@ -233,7 +235,7 @@ function LegendRows({
                 onChangeVisibility={onChangeVisibility}
                 onChangeCollapsed={onChangeCollapsed}
               >
-                <LegendComponent legend={legend} />
+                <LegendComponent layer={layer} legend={legend} />
               </LegendWrapper>
               {!isSingle && !isLast && <Divider />}
             </Fragment>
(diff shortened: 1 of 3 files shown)


packages/react-ui/src/widgets/legend/LegendWidgetUI.js
```diff
@@ -233,7 +235,7 @@ function LegendRows({
                 onChangeVisibility={onChangeVisibility}
                 onChangeCollapsed={onChangeCollapsed}
               >
-                <LegendComponent legend={legend} />
+                <LegendComponent layer={layer} legend={legend} />
```
eamador, line 238: i see that `LegendComponent` can be:
```ts
const LEGEND_COMPONENT_BY_TYPE = {
  [LEGEND_TYPES.CATEGORY]: LegendCategories,
  [LEGEND_TYPES.ICON]: LegendIcon,
  [LEGEND_TYPES.CONTINUOUS_RAMP]: (args) => <LegendRamp {...args} isContinuous={true} />,
  [LEGEND_TYPES.BINS]: (args) => <LegendRamp {...args} isContinuous={false} />,
  [LEGEND_TYPES.PROPORTION]: LegendProportion,
  [LEGEND_TYPES.CUSTOM]: ({ legend }) => legend.children
};
```
or 
```ts
customLegendTypes
```
or
```ts
UnknownLegend
```

But i don't see those components receiving the layer object, for instance `LegendCategories`

https://github.com/CartoDB/carto-react/blob/76cba7e25d3b64592a68a08244639e405f9ae9c4/packages/react-ui/src/widgets/legend/LegendCategories.js#L6

Clebal: Yes, the goal of adding this is to have it in custom legend types.

```
const LegendComponent = LEGEND_COMPONENT_BY_TYPE[type] || customLegendTypes[type] || UnknownLegend;
```

Using customLegendTypes the dev can build a component that will be rendered as custom legend and that can be reused for multiple layers. In the default legend components, the layer isn't important, but it could be in the custom legends (that' why I'm adding it, because it's necessary in Site Selection)

Clebal: LegendCategories and the others default components just ignore this layer prop because they don't need it.

eamador: understood! thank you for the explanation!

eamador (approved):
nice one!

PR #401, opened 2022-05-05
Fix deprecated warning in HistogramWidgetUI

Clebal:
api.style() is deprecated in echarts, so a warning was shown in each render. I've changed the code so we don't use any deprecated fn.


eamador (approved):
LGTM

PR #402, opened 2022-05-05
Use static bins in HistogramWidget

Clebal:
- Added `min: number` and `max: number` props as mandatory.
- Remove code to calculate viewport min/max.
- Restore the old HistogramModel and histogram aggregation fn, and their tests.
- Calculate ticks manually outside the widget fetch, and keep them until min/max change.
- Fixed minor issues in HistogramWidgetUI when the max Y axis value is 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Not released
 
 - Fix filters with null as upper or lower bound [#403](https://github.com/CartoDB/carto-react/pull/403)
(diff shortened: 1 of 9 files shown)


eamador (approved):
I haven't tested it but it looks good!

VictorVelarde (approved):
LGTM @Clebal. Let's go for an alpha. But before, let's prepare a quick bump to brand new https://www.npmjs.com/package/@deck.gl/core/v/8.7.9

PR #407, opened 2022-05-12
Fixes in HistogramWidgetUI

Clebal:


packages/react-ui/src/widgets/HistogramWidgetUI/HistogramWidgetUI.js
```diff
@@ -180,6 +174,15 @@ function HistogramWidgetUI({
 
   // Series
   const seriesOptions = useMemo(() => {
+    const dataWithColor = formattedData.map((item, idx) => {
+      const isDisabled = selectedBars.length && selectedBars.indexOf(idx) === -1;
```
eamador, line 178: I'd rather go with `selectedBars.includes(idx)` instead of `selectedBars.indexOf(idx) === -1`, what do you think?

Clebal: I think that both are the same, right? Any advantages of using includes?

eamador (approved):
LGTM 👍 , just a minor suggestion, proceed to implement it if you agree, or just leave as it's done now, it is just a matter of readability

packages/react-ui/src/widgets/HistogramWidgetUI/HistogramWidgetUI.js
eamador: It is just for readability, i think it's clearer with "includes" and you don't have to compare it with `-1`

VictorVelarde: +1 on readability.

PR #411, opened 2022-05-18
FIX: Cannot filter map with category widget with boolean values. 

Josmorsot:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/230271/ac-pbwqze2p-cannot-filter-category-widget-for-boolean-values

Category Widget don't filter properly when column is `boolean`

## Type of change

- Fix

# Acceptance

1. Create a builder map with a layer that has a boolean field. E.g: from team account
 ```sql
     SELECT *, price_num > 500 as price_over_500_bool FROM `carto-dw-ac-7xhfwyml.shared.la_airbnb` 
```
2. Create a category widget based on the boolean column (e.g: `price_over_500_bool`)

https://user-images.githubusercontent.com/7818205/168979059-9d4271a9-41bc-4934-8547-053a6b222282.mov

### DIFF
--- a/packages/react-core/src/operations/groupBy.js
+++ b/packages/react-core/src/operations/groupBy.js
@@ -11,11 +11,11 @@ export function groupValuesByColumn({
   if (Array.isArray(data) && data.length === 0) {
     return null;
   }
-
   const groups = data.reduce((accumulator, item) => {
     const group = item[keysColumn];
 
-    accumulator[group] = accumulator[group] || [];
+    const values = accumulator.get(group) || []
+    accumulator.set(group, values)
 
     const aggregatedValue = aggregate(item, valuesColumns, joinOperation);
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by Clebal — CHANGES_REQUESTED** [review]
I haven't tested it but the code looks great. The only thing I miss is a change in the CategoryWidgetUI type in types.d.ts

File: packages/react-core/src/operations/groupBy.js [hunk 1]
```diff
@@ -24,16 +24,18 @@ export function groupValuesByColumn({
...
 
     if (isValid) {
-      accumulator[group].push(aggregatedValue);
+      values.push(aggregatedValue)
+      accumulator.set(group, values);
     }
 
     return accumulator;
-  }, {});
+  }, new Map());
```
VictorVelarde, line 32: Drop a couple of lines about why we're using this

VictorVelarde (commented):
Let's add proper tests, especially for boolean, and we'll be ready to an alpha

Clebal (approved):
LGTM! As @VictorVelarde said, add the changelog entry and you're ready to merge.

VictorVelarde:
I'll prepare an alpha, getting also latest deck.gl 8.8.x alpha

PR #412, opened 2022-05-19
Autocalculate min/max in HistogramWidget

Clebal:
Now the min/max props in HistogramWidget are optionals.


VictorVelarde (commented):
Make the change to new _getStats, and we'll go ahed with this one 👍🏻

VictorVelarde (approved):
🚀

PR #415, opened 2022-05-24
Use new Query Models in SQL API for global widgets

Clebal:
This PR covers the changes needed to use the new Model API in the global widgets.

In order to use the Model API, a new api fn has been implemented in @carto/react-api called `executeModel` following the same path that with Stats API, LDS API and SQL API.

Updated widgets:

- [x] CategoryWidget (also Pie)
- [x] FormulaWidget
- [x] HistogramWidget
- [x] TimeSeriesWidget

This PR is pending of merge of [this one](https://github.com/CartoDB/cloud-native/pull/6840).

### DIFF
--- a/packages/react-api/src/api/model.js
+++ b/packages/react-api/src/api/model.js
@@ -1,0 +1,66 @@
+import { assert, checkCredentials, makeCall } from './common';
+import { MAP_TYPES, API_VERSIONS } from '@deck.gl/carto';
+
+const URL_LENGTH = 2048;
+
+const AVAILABLE_MODELS = ['category', 'histogram', 'formula', 'timeseries'];
+
+/**
+ * Execute a SQL model request.
+ *
+ * @param { object } props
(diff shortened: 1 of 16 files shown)


packages/react-api/src/api/model.js
```diff
@@ -0,0 +1,66 @@
+import { assert, checkCredentials, makeCall } from './common';
+import { MAP_TYPES, API_VERSIONS } from '@deck.gl/carto';
+
+const URL_LENGTH = 2048;
+
+const AVAILABLE_MODELS = ['category', 'histogram', 'formula', 'timeseries'];
```
VictorVelarde, line 6: I would probably suggest creating an enum or set of consts, but up to you

VictorVelarde (commented):
Approach looks good to me, but before merging I think we should land new backend api, and include TimeSeries.

eamador (approved):

PR #416, opened 2022-05-30
Add layer options extension mechanism in LegendWidget

juandjara:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/221433/add-extensions-mechanism-for-layer-options-in-the-legend

This change addresses the need for layer option extension in legend. See [RFC ch219693](https://app.shortcut.com/cartoteam/story/219693/rfc-add-extensions-mechanism-for-layer-options-in-the-legend) for motivation. With the new config, a user will be able to add an arbitrary number of configuration sections to a legend item that will update the associated layer.

## Type of change

- Feature

# Acceptance

The new config of the `LegendWidget` looks like this:

1. somewhere in your code, you define the extensions keys that will be available for the legend items
```
export const LAYER_OPTIONS = {
  PALETTE_SELECTOR: 'PALETTE_SELECTOR',
};
```

2. In your layer config, you can include which sections this layer will show
```
const layerConfig = {
  title: 'Store types',
  visible: true,
  opacity: 1,
  showOpacityControl: true,
  options: [LAYER_OPTIONS.PALETTE_SELECTOR],
  legend: {
    attr: 'storetype',
    type: LEGEND_TYPES.CATEGORY,
    labels: [...CATEGORIES, OTHERS_KEY],
    colors: [...DEFAULT_COLORS, OTHERS_COLOR],
  },
};
```

3. Then, inside your component that renders the legend, you will define a const like this, mapping extension keys to extension components. Every extension component will be passed two props:
* layerId: string - ID of the related layer
* optionsKey: string - options key associated with this component

```
const LAYER_OPTIONS_COMPONENTS = {
  [LAYER_OPTIONS.PALETTE_SELECTOR]: PaletteSelector,
};
```

4. And then pass it to the LegendWidget with the `customLayerOptions` prop
```
<LegendWidget customLayerOptions={LAYER_OPTIONS_COMPONENTS} />
```

Using this mechanism, the LegendWidget will asociate the extension keys with the extension components and render accordingly.

You can see an example of how this will be implemented for a particular use case in the `template-sample-app-3` in [this commit](https://github.com/CartoDB/carto-react-template/commit/a9b7680f03fa0dbeda56ccb8d7c056242b025e22)

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry (TODO)
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests (TODO)
- Documentation (TODO)


juandjara:
Setting this to a draft PR because tests and docs are missing at the moment. I would rather validate the approach and the props / variables names before continuing.

packages/react-ui/src/widgets/legend/LegendWrapper.js
```diff
@@ -33,9 +33,22 @@ const useStyles = makeStyles((theme) => ({
...
+  return (
+    <p>
+      Unknown configuration key{' '}
+      <em>
+        <strong>{optionsKey}</strong>
+      </em>
+    </p>
+  );
+}
+
```
Clebal, line 46: I'd be more specific here and normalise the names.
optionsKey -> optionKey
UnknownLegendOption -> UnknownLayerOption
Unknown configuration key  -> Unknown layer option

packages/react-ui/src/widgets/legend/LegendWrapper.js
```diff
@@ -101,6 +114,10 @@ export default function LegendWrapper({
                     opacity={opacity}
                     onChangeOpacity={handleChangeOpacity}
                   />
+                  {selectedLayerOptions.map((key) => {
+                    const Component = customLayerOptions[key] || UnknownLegendOption;
+                    return <Component key={key} optionsKey={key} layerId={id} />;
```
Clebal, line 119: I think optionKey is a good addition, but in the task it says that the layer is passed as prop, not only the layer id.

Clebal (changes requested):
It's very well done, just minor changes. Congrats @juandjara

packages/react-ui/src/widgets/legend/LegendWrapper.js
juandjara: But passing the whole layer as a prop would mean more modifications to LegendWrapper. Meaning you have to pass to that component the whole layer of get the whole layer via a hook in that componente. Are you ok with that ?

packages/react-ui/src/widgets/legend/LegendWrapper.js
juandjara: Ok, let's do it

packages/react-ui/src/widgets/legend/LegendWrapper.js
Clebal: Modify LegendWrapper to receive `layerOptions: React.Fn[]` instead of `selectedLayerOptions` and `customLayerOptions`.
`layerOptions` will be:
```js
const layerOptions = layer.options.map((key) => {
                    const Component = customLayerOptions[key] || UnknownLegendOption;
                    return <Component key={key} optionsKey={key} layer={layer} />;
})
```
So you have the layer obj and you don't put that logic in LegendWrapper making it easier. In the LegendWrapper you only render the layerOptions in JSX and that's all.

juandjara: Okay so `layerOptions` will be passed to `LegendWrapper` kind of like a `children` prop

Clebal: Yep 👌🏻 I think that'd work

juandjara:
Related documentation PR: https://github.com/CartoDB/documentation/pull/323

Clebal (approved):
LGTM

PR #417, opened 2022-05-31
Tiles features calculation for spatial indexes

padawannn:
# Description

- It remains to modify the `useCartoLayerProps` that will do in other PR
- Tests pending

## Type of change

- Feature

### DIFF
--- a/packages/react-core/src/filters/tileFeaturesSpatialIndex.js
+++ b/packages/react-core/src/filters/tileFeaturesSpatialIndex.js
@@ -1,6 +1,6 @@
 import { indexes } from '@mapbox/tile-cover';
 // h3-js has a known problem that does not allow us to use it in a web-worker. To solve
-// it we're overwriting the node_module package after install it (check postinstall script in package.json)
+// it we're overwriting the node_module package after install it (check postinstall script in the package.json and the patches/h3-js+3.7.2.patch file)
 // To know more about the h3-js issue check https://github.com/uber/h3-js/issues/117
 import { h3GetResolution, polyfill } from 'h3-js';
 import bboxClip from '@turf/bbox-clip';
@@ -62,6 +62,9 @@ function getResolution(tiles, spatialIndex) {
   }
 }
 
+const bboxWest = [-180, -90, 0, 90];
+const bboxEast = [0, -90, 180, 90];
+
 function getCellsCoverGeometry(geometry, spatialIndex, resolution) {
   if (spatialIndex === SpatialIndex.QUADINT || spatialIndex === SpatialIndex.QUADKEY) {
     return indexes(geometry, {
@@ -71,9 +74,6 @@ function getCellsCoverGeometry(geometry, spatialIndex, resolution) {
   }
 
   if (spatialIndex === SpatialIndex.H3) {
-    const bboxWest = [-180, -90, 0, 90];
-    const bboxEast = [0, -90, 180, 90];
-
     // The current H3 polyfill algorithm can't deal with polygon segments of greater than 180 degrees longitude
     // so we clip the geometry to be sure that none of them is greater than 180 degrees
     // https://github.com/uber/h3-js/issues/24#issuecomment-431893796
--- a/packages/react-core/webpack.config.js
+++ b/packages/react-core/webpack.config.js
@@ -7,7 +7,7 @@ const webpackBaseConfig = require('../../webpack.base');
 const webpackConfig = {
   ...webpackBaseConfig,
   // Be careful what you add as external because it can significantly affect the size of the worker package.
-  externals: [/^@turf\/.+$/, /^@mapbox\/.+$/, 'h3-js'],
+  externals: [/^@turf\/.+$/, '@mapbox/tile-cover', 'h3-js'],
   output: {
     path: path.resolve(__dirname, './dist'),
     filename: 'index.js',


package.json
```diff
@@ -60,7 +64,8 @@
     "storybook:build": "lerna --scope @carto/react-ui exec -- npm run storybook:build",
     "storybook:publish": "lerna --scope @carto/react-ui exec -- npm run storybook:publish",
     "publish:prerelease": "yarn run build:clean && lerna publish --dist-tag prerelease --force-publish --preid rc",
-    "publish:release": "yarn run build:clean && lerna publish --force-publish"
+    "publish:release": "yarn run build:clean && lerna publish --force-publish",
+    "postinstall": "patch-package"
```
padawannn, line 68: This comment https://github.com/CartoDB/carto-react/pull/417/files#diff-337ead04a20e03f578a3166d4928bcf79a396bfd9df9e750662aeb689a5d5aceR2

patches/h3-js+3.7.2.patch
```diff
@@ -0,0 +1,34 @@
+diff --git a/node_modules/h3-js/dist/browser/h3-js.es.js b/node_modules/h3-js/dist/browser/h3-js.es.js
```
padawannn, line 1: This comment https://github.com/CartoDB/carto-react/pull/417/files#diff-337ead04a20e03f578a3166d4928bcf79a396bfd9df9e750662aeb689a5d5aceR2

packages/react-core/src/filters/tileFeaturesSpatialIndex.js
```diff
@@ -0,0 +1,94 @@
...
+  if (spatialIndex === SpatialIndex.QUADINT || spatialIndex === SpatialIndex.QUADKEY) {
+    return indexes(geometry, {
+      min_zoom: resolution,
+      max_zoom: resolution
+    });
+  }
+
+  if (spatialIndex === SpatialIndex.H3) {
+    const bboxWest = [-180, -90, 0, 90];
+    const bboxEast = [0, -90, 180, 90];
```
Clebal, line 75: If they're constants, maybe we should put them out of the fn so we don't redeclare them in each execution.

packages/react-core/src/operations/constants/SpatialIndexTypes.d.ts
```diff
@@ -0,0 +1,6 @@
+export enum SpatialIndex {
+  H3 = 'h3',
+  S2 = 's2',
+  QUADKEY = 'quadkey',
+  QUADINT = 'quadint'
+}
```
Clebal, line 6: I think that in TS the enum keys are in pascal case.
https://www.typescriptlang.org/docs/handbook/enums.html

Clebal (approved):
@padawannn everything looks great, just minor comments, not blockers

packages/react-core/src/operations/constants/SpatialIndexTypes.d.ts
padawannn: > I think that in TS the enum keys are in pascal case. https://www.typescriptlang.org/docs/handbook/enums.html

I would say that we are not following that convention in the project
- https://github.com/CartoDB/carto-react/blob/fad57802b35fd7c467e5bfb36b470191f740ebce/packages/react-core/src/filters/FilterTypes.d.ts#L1
- https://github.com/CartoDB/carto-react/blob/2b603f3e86d3d04df4e6232768a4bb62c891674a/packages/react-redux/src/slices/cartoSlice.d.ts#L57
- https://github.com/CartoDB/carto-react/blob/6b62898ebeaa902d6ddfb17a2e3d37b8eeebafca/packages/react-basemaps/src/basemaps/basemaps.d.ts#L1

Clebal: Yes... I know, but I thought we would do it correctly hereinafter hahaha. But it's ok.

packages/react-core/src/filters/tileFeaturesSpatialIndex.js
```diff
@@ -0,0 +1,94 @@
+import { indexes } from '@mapbox/tile-cover';
+// h3-js has a known problem that does not allow us to use it in a web-worker. To solve
+// it we're overwriting the node_module package after install it (check postinstall script in package.json)
```
felixpalmer, line 3: I would mention the `patches/h3-js+3.7.2.patch` file here. Until I found that I had no idea how this worked

packages/react-core/src/filters/tileFeaturesSpatialIndex.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  if (spatialIndex === SpatialIndex.H3) {
+    const bboxWest = [-180, -90, 0, 90];
+    const bboxEast = [0, -90, 180, 90];
+
+    // The current H3 polyfill algorithm can't deal with polygon segments of greater than 180 degrees longitude
+    // so we clip the geometry to be sure that none of them is greater than 180 degrees
+    // https://github.com/uber/h3-js/issues/24#issuecomment-431893796
+    return polyfill(
```
felixpalmer, line 80: Will the resolution always be high relative to the viewport? I had to oversample in deck to avoid edge effects

felixpalmer (approved):

packages/react-core/src/filters/tileFeaturesSpatialIndex.js
padawannn: Done

packages/react-core/webpack.config.js
```diff
@@ -7,7 +7,7 @@ const webpackBaseConfig = require('../../webpack.base');
 const webpackConfig = {
   ...webpackBaseConfig,
   // Be careful what you add as external because it can significantly affect the size of the worker package.
-  externals: [/^@turf\/.+$/],
+  externals: [/^@turf\/.+$/, /^@mapbox\/.+$/, 'h3-js'],
```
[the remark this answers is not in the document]
padawannn: Done

packages/react-core/src/filters/tileFeaturesSpatialIndex.js
padawannn: I don't understand very well, probably due to my lack of knowledge

felixpalmer: Perhaps I've got myself confused, I was just saying that doing it this way means we only capture hexagons that have their center in the polygon. If that is by design that is fine. 

For background: in deck, when filling the viewport so that I capture hexagons which are only partially in the viewport I polyfill at a higher resolution and then get the parents

PR #425, opened 2022-06-08
Spatial index support for useCartoLayerProps

padawannn:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/233636/c4r-h3-qk-qint-support-usecartolayerprops-to-pass-param

Adapt useCartoLayerProps to support Spatial Index
- 8.8.0-alpha.6 DeckGl version is required.
- At high zoom levels we can get failed requests. Check this thread https://cartoteam.slack.com/archives/C035NAP4YHZ/p1654682511529729
- Remember that to test tables with spatial index the `geoColumn` and the `aggregationExp` are mandatories but are no necessary for static tilesets with spatial index

## Type of change
- Feature

### DIFF
--- a/packages/react-api/src/hooks/useCartoLayerProps.js
+++ b/packages/react-api/src/hooks/useCartoLayerProps.js
@@ -1,11 +1,10 @@
-import { useCallback, useState } from 'react';
+import { useCallback } from 'react';
 import { useSelector } from 'react-redux';
 import { selectSpatialFilter } from '@carto/react-redux';
 import useGeojsonFeatures from './useGeojsonFeatures';
 import useTileFeatures from './useTileFeatures';
 import { getDataFilterExtensionProps } from './dataFilterExtensionUtil';
 import { getMaskExtensionProps } from './maskExtensionUtil';
-import { SpatialIndex } from '@carto/react-core';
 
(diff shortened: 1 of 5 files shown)


packages/react-api/src/hooks/useCartoLayerProps.js
```diff
@@ -29,12 +31,16 @@ export default function useCartoLayerProps({
...
+    spatialIndex
   });
 
   const onDataLoad = useCallback(
     (data) => {
       if (data?.tilejson) {
+        setSpatialIndex(
+          Object.values(SpatialIndex).includes(data.scheme) ? data.scheme : undefined
+        );
         return onDataLoadForTile(data);
```
Clebal, line 44: Since the spatialIndex state is only used in useTileFeatures, I think we should move that state inside the hook and move 41-43 lines inside the onDataLoadForTile. What do you think?

Clebal (approved):
Just one minor comment, but you can omit it if you don't find it interesting

packages/react-api/src/hooks/useCartoLayerProps.js
padawannn: Totally agree

PR #426, opened 2022-06-08
Ensure source exists in HistogramWidget before getting stats

sclebal:
After testing the latest alpha in the templates I discovered that it's possible to render the widget before adding the source, so it throws an error when trying to request the HistogramWidget stats. To avoid that error, I added source as a condition to get the stats.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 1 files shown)


VictorVelarde (approved):
LGTM. Do we have to add similar checks in other widgets?

Clebal:
> LGTM. Do we have to add similar checks in other widgets?

No, because other widgets use useWidgetFetch that handle this and others problems. HistogramWidget is the only widget that makes two requests.

PR #427, opened 2022-06-09
Add warning to widgets when column is missing

sclebal:
image

- Implemented InvalidColumnError to throw it when the column used is invalid.
- Used InvalidColumnError to throw errors in `stats.js` and `features.worker.js`.
- Adapted HistogramWidget minMax fetching and useWidgetFetch to handle InvalidColumnError.

### DIFF
--- a/packages/react-widgets/src/widgets/utils/constants.js
+++ b/packages/react-widgets/src/widgets/utils/constants.js
@@ -1,2 +1,2 @@
 export const DEFAULT_INVALID_COLUMN_ERR =
-  'The column selected for this widget is no longer available in the data source.';
+  'One or more columns used in this widget are not available in the data source.';


zbigg (changes requested):

packages/react-core/src/utils/InvalidColumnError.js
```diff
@@ -0,0 +1,16 @@
+const NAME = 'InvalidColumnError';
+const MESSAGE_START = `Uncaught ${NAME}:`;
+
+export class InvalidColumnError extends Error {
+  static message =
+    'The column selected for this widget is no longer available in the data source.';
+
+  constructor(message) {
```
zbigg, line 8: My proposition would be to change this to

super(`ERR_INVALID_COLUMN: ${message || default}`)

and in is just test for existence of `ERR_INVALID_COLUMN: ` in error message.

This is still _string_ test, but at least based on something that looks like error code and _is_ universal and doesn't depend on particular details like adding "Uncaught..." by particular runtime.

packages/react-widgets/__tests__/hooks/useWidgetFetch.test.js
```diff
@@ -79,17 +80,45 @@ describe('useWidgetFetch', () => {
     await act(() => sleep(250));
     expect(screen.queryByText('loading')).not.toBeInTheDocument();
 
-    expect(onError).toBeCalled();
+    expect(onError).toBeCalledTimes(1);
+
+    modelFn.mockRejectedValue(
+      new InvalidColumnError('Uncaught InvalidColumnError: Invalid column')
+    );
```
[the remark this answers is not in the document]
Clebal: Idk why when I throw the error in the test the `Uncaught InvalidColumnError` label doesn't appear but this is the classic label `Uncaught {error class name}` that appears always in JS, that's why I used it to detect the error in the `is` fn.

packages/react-core/src/utils/InvalidColumnError.js
```diff
@@ -0,0 +1,16 @@
+const NAME = 'InvalidColumnError';
+const MESSAGE_START = `Uncaught ${NAME}:`;
+
+export class InvalidColumnError extends Error {
+  static message =
+    'The column selected for this widget is no longer available in the data source.';
```
zbigg, line 6: Minor: i don't think message copy belongs to "react-core" and exception type, it's purely widgets & UI matter (and it's used only once there) so maybe move it there?

This should be in `[useWidgetFetch.js]` (or in UI component).

Clebal: It used in @carto/react-api also.

Clebal: Oh, now I read it correctly. I can move the message to the useWidgetFetch, ok.

zbigg (approved):
LGTM :)

PR #433, opened 2022-06-16
Add layerOrder in LegendWidget

flipasg:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/222396/implement-layerorder-in-legendwidget

**Background**
Currently, the legends are rendered using the Object.values fn with the layers object where we store the layers config (store) in redux.

**Problem**
Object.values fn doesn't guarantee the order ([Source](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order)), so we might have weird behaviour considering that the addLayer action is executed in the same order that the layer are executed in the layers/index.js file.

**Solution**
Add **layerOrder: string[]** prop in LegendWidget. layerOrder is an array of layer ids, when the Object.values is executed, the resulting array must be ordered following the layerOrder prop.

Add **sortArrayByPropValues** utils. Is a function to sort a list of objects based on a property values.

## Type of change

- Feature
- Tests

# Acceptance

**In your App**
1. Go to your **LegendWidget** usage and add **layerOrder** property:

The specified **layerOrder**, from left to the right, should be the order we want from bottom to top in the legend display.

```
<LegendWidget
  layerOrder={['storesLayer2', 'storesLayer2', 'storesLayer0']}
  className={classes.legend}
/>
```

2. Check the legend of your application map in the browser:

The elements should be sorted by id property, from bottom to top. 
- If exists any value not included in the array should be positioned at the top of the legend.
- If exists any value not existing in the layers should be ignored.
- If exist any duplicated value should act as one. Using the first occurrence of this value.

image

**In carto-react**

1. Check the tests by `yarn run test` command
The added test are related with the util **sortArrayByPropValues** .


packages/react-widgets/__tests__/widgets/utils/sortArrayByPropValues.test.js
```diff
@@ -0,0 +1,85 @@
...
+      }
+    ];
+  });
+
+  test('should ignore not existing prop', () => {
+    const ORDER = ['storesLayer3', 'storesLayer2', 'storesLayer1', 'storesLayer0'];
+    sortArrayByPropValues(ELEMENTS, ORDER, 'not_existing_prop');
+    expect(ELEMENTS).toStrictEqual(ELEMENTS);
+  });
+  test('should ignore not existing values', () => {
```
Clebal, line 35:
```suggestion
  });

  test('should ignore not existing values', () => {
```

packages/react-widgets/src/widgets/utils/sortArrayByPropValues.js
```diff
@@ -0,0 +1,24 @@
...
+      ...array.filter((element) => !element.hasOwnProperty(prop)),
+      ...elementsWithProp.filter((item) => uniqueValues.indexOf(item[prop]) === -1),
+      ...elementsWithProp
+        .filter((item) => uniqueValues.indexOf(item[prop]) !== -1)
+        .sort((a, b) => uniqueValues.indexOf(b[prop]) - uniqueValues.indexOf(a[prop]))
+    ];
+  }
+
+  return array;
+}
```
Clebal, line 24: I don't think it's bad but I think that it's overengineering.

Reasons:
1. You create an utils to be used in only one place.
2. This util covers a lot of use cases that aren't necessary, having a certain impact in the performance.

Keep it simple, I think that this is far enough:
```js
Object.values(layers)
    .filter(({ legend }) => legend)
    .sort((layerA, layerB) => order.indexOf(layerB.id) - order.indexOf(layerA.id))
```

You can add there asserts to check that the id exists if you want, but I think it's also unnecessary because you cannot add layers without an id.

Clebal (changes requested):
I think the code is quite complex considering that we're covering a very simple use case.

packages/react-widgets/src/widgets/utils/sortLayersById.js
```diff
@@ -0,0 +1,16 @@
+import { assert } from '@carto/react-core';
+
+/**
+ * Sorts a list of layers based on a list of id property values
+ * @param {any[]} layers list of layers to sort
+ * @param {string[]} layerOrder values to sort by
+ * @returns sorted array
+ */
+export default function sortLayersById(layers = [], layerOrder = []) {
+  const uniqueValues = [...new Set(layerOrder)];
```
Clebal, line 10: Remove this, it's something up to the user and even if there are repeated values it doesn't affect us.

packages/react-widgets/src/widgets/utils/sortLayersById.js
```diff
@@ -0,0 +1,16 @@
+import { assert } from '@carto/react-core';
+
+/**
+ * Sorts a list of layers based on a list of id property values
+ * @param {any[]} layers list of layers to sort
+ * @param {string[]} layerOrder values to sort by
+ * @returns sorted array
+ */
+export default function sortLayersById(layers = [], layerOrder = []) {
```
Clebal, line 9: `sortLayers` much simpler.

packages/react-widgets/src/widgets/utils/sortLayersById.js
```diff
@@ -0,0 +1,16 @@
...
+  const uniqueValues = [...new Set(layerOrder)];
+  assert(
+    layers.every((layer) => layer.id),
+    'Layers must have an id property'
+  );
```
Clebal, line 14: Move this assert to the sort so we avoid looping again over layers.

packages/react-widgets/src/widgets/utils/sortLayersById.js
Clebal: I miss this check at the beginning:

```
if (!layerOrder.length) {
    return layers;
}
```

So we avoid the sort.

packages/react-widgets/src/widgets/LegendWidget.js
```diff
@@ -11,16 +12,21 @@ import PropTypes from 'prop-types';
...
  * @param  {boolean} [props.initialCollapsed] - Define initial collapsed value. false by default.
+ * @param  {string[]} [props.layerOrder] - Array of layer identifiers. Defines the order of layer legends. [] by default.
  */
 function LegendWidget({
   className,
   customLayerOptions,
   customLegendTypes,
-  initialCollapsed
+  initialCollapsed,
+  layerOrder = []
```
Clebal, line 22: You have to fill the PropTypes at the bottom of this file.

packages/react-widgets/__tests__/widgets/utils/sortLayersById.js
```diff
@@ -0,0 +1,72 @@
+import sortLayersById from '../../../src/widgets/utils/sortLayersById';
```
Clebal, line 1: The name of this file should contains `.test.js`

packages/react-widgets/__tests__/widgets/utils/sortLayersById.js
```diff
@@ -0,0 +1,72 @@
...
+
+  test('should add missing ids at end', () => {
+    const ORDER = ['storesLayer1', 'storesLayer2', 'storesLayer0'];
+    const ORDERED_ELEMENTS = sortLayersById(ELEMENTS, ORDER)
+      .map(({ id }) => id)
+      .slice(ORDER.length);
+    expect(ORDERED_ELEMENTS.every((id) => ORDER.indexOf(id) === -1)).toBeTruthy();
+    expect(ORDERED_ELEMENTS.length).toBe(1);
+  });
+  test('should throw an error when elements have no id property', () => {
```
Clebal, line 53:
```suggestion
  });

  test('should throw an error when elements have no id property', () => {
```

Be careful with the spaces.

packages/react-widgets/__tests__/widgets/utils/sortLayersById.js
```diff
@@ -0,0 +1,72 @@
...
+    expect(ORDER.slice(1)).toStrictEqual(
+      sortLayersById(ELEMENTS, ORDER)
+        .map(({ id }) => id)
+        .reverse()
+    );
+  });
+
+  test('should add missing ids at end', () => {
+    const ORDER = ['storesLayer1', 'storesLayer2', 'storesLayer0'];
+    const ORDERED_ELEMENTS = sortLayersById(ELEMENTS, ORDER)
```
Clebal, line 47: I accept that order comes in mayusc, but ordered elements should be camel case

packages/react-widgets/src/widgets/utils/sortLayersById.js
```diff
@@ -0,0 +1,16 @@
+import { assert } from '@carto/react-core';
+
+/**
+ * Sorts a list of layers based on a list of id property values
+ * @param {any[]} layers list of layers to sort
```
Clebal, line 5:
```suggestion
 * @param {{ id: string }[]} layers - array of layers to sort
```

Btw, usually between the type and the description we put a slash

packages/react-widgets/src/widgets/utils/sortLayersById.js
```diff
@@ -0,0 +1,16 @@
...
+ * @param {string[]} layerOrder values to sort by
+ * @returns sorted array
+ */
+export default function sortLayersById(layers = [], layerOrder = []) {
+  const uniqueValues = [...new Set(layerOrder)];
+  assert(
+    layers.every((layer) => layer.id),
+    'Layers must have an id property'
+  );
+  return layers.sort((a, b) => uniqueValues.indexOf(b.id) - uniqueValues.indexOf(a.id));
```
Clebal, line 15:
```suggestion
  return [...layers].sort((layerA, layerB) => {
      assert(layerA.id && layerB.id, 'Layer must have an id')
      return uniqueValues.indexOf(layerB.id) - uniqueValues.indexOf(layerA.id));
  }
```

Clebal: I added `[...layers]` to avoid modifying layers original reference so `sortLayers` can be a pure fn.

packages/react-widgets/__tests__/widgets/utils/sortLayersById.js
```diff
@@ -0,0 +1,72 @@
...
+        source: 'storesSource',
+        title: 'Store types 2'
+      },
+      {
+        id: 'storesLayer3',
+        source: 'storesSource',
+        title: 'Store types 3'
+      }
+    ];
+  });
```
Clebal, line 28: This beforeEach shouldn't be necessary after the reference change I suggested in sortLayers

Clebal (changes requested):

packages/react-widgets/src/widgets/utils/sortLayersById.js
flipasg: perfect! but conceptually  sort (like the Arrays.prototype function), returns the array but also modifies the order of the original ones to.

packages/react-widgets/__tests__/widgets/utils/sortLayersById.js
flipasg: is a beforeAll no each, only executed at the beginning

packages/react-widgets/src/widgets/utils/sortLayersById.js
Clebal: That why I added the `[...layers]`.

packages/react-widgets/src/widgets/LegendWidget.js
Clebal: Remember also to modify the types.d.ts

packages/react-widgets/__tests__/widgets/utils/sortLayersById.js
Clebal: Yep! Sorry, but still unnecessary.

Clebal (changes requested):

Clebal (approved):

packages/react-core/src/index.js
```diff
@@ -12,6 +12,7 @@ export { InvalidColumnError } from './utils/InvalidColumnError';
 export { debounce } from './utils/debounce';
 export { throttle } from './utils/throttle';
 export { randomString } from './utils/randomString';
+export { assert } from './utils/assert';
```
zbigg, line 15: This is internal API, not to be used by public, please export it with as`_assert`.

Then, use it like 

    import { _assert as assert } from '@carto/react-core/';

in other modules

zbigg (commented):
LGTM, but please consider comment about _assert

packages/react-core/src/index.js
flipasg: done

zbigg (approved):

PR #439, opened 2022-06-27
Check null value to avoid crash on category widget

moicalcob:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/239960/ac-pbwqze2p-carto-3-user-s-category-widget-breaking-on-search

We've encountered that there is an issue with category widgets. The whole application crashes if we use this widget filtering a column that contains null values.

## Type of change

- Fix

# Acceptance

1. Add a category widget to a map created filtering a column that contains null values
2. Check that it doesn't fail if we try to search a value in the category widget

# Basic checklist

- [x] Good PR name
- [x] Shortcut link
- [x] Changelog entry
- [x] Just one issue per PR
- [x] GitHub labels
- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/react-ui/src/widgets/CategoryWidgetUI.js
+++ b/packages/react-ui/src/widgets/CategoryWidgetUI.js
@@ -272,7 +272,8 @@ function CategoryWidgetUI(props) {
           ? list.filter((elem) => {
               return (
                 elem.name !== null &&
-                (elem.name?.toLowerCase().indexOf(searchValue.toLowerCase()) !== -1 ||
+                elem.name !== undefined &&
+                (elem.name.toLowerCase().indexOf(searchValue.toLowerCase()) !== -1 ||
                   (labels[elem.name]
                     ? labels[elem.name]
                         .toLowerCase()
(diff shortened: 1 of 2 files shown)


packages/react-ui/src/widgets/CategoryWidgetUI.js
```diff
@@ -271,11 +271,13 @@ function CategoryWidgetUI(props) {
         return searchValue
           ? list.filter((elem) => {
               return (
-                elem.name.toLowerCase().indexOf(searchValue.toLowerCase()) !== -1 ||
-                (labels[elem.name]
-                  ? labels[elem.name].toLowerCase().indexOf(searchValue.toLowerCase()) !==
-                    -1
-                  : false)
+                elem.name !== null &&
+                (elem.name?.toLowerCase().indexOf(searchValue.toLowerCase()) !== -1 ||
```
Clebal, line 275: I would include also `!== undefined` and by that way you can avoid the optional chaining after name.

moicalcob: Done! 😄

PR #451, opened 2022-07-13
LegendCategories: support for custom markers

zbigg:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/244132/implement-support-for-markers-in-builder-i-single-icon

Add support for rendering of 'masked colored icons' in `LegendCategories` widget instead of classic circles.

Rationale: This feature is needed in builder and it's simpler to implement here as we already use  `LegendCategories` as a whole.

Needed by: https://github.com/CartoDB/cloud-native/pull/8559

image

## Type of change

- Feature

# Acceptance

-

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-ui/src/widgets/legend/LegendCategories.js
+++ b/packages/react-ui/src/widgets/legend/LegendCategories.js
@@ -66,8 +66,13 @@ const useStyles = makeStyles((theme) => ({
   marker: {
     whiteSpace: 'nowrap',
     display: 'block',
+    width: '12px',
+    height: '12px',
+    borderRadius: '50%',
     position: 'relative',
-
+    border: '2px solid transparent'
+  },
+  circle: {
     '&::after': {
       position: 'absolute',
       display: ({ isMax }) => (isMax ? 'block' : 'none'),
@@ -80,15 +85,9 @@ const useStyles = makeStyles((theme) => ({
       boxSizing: 'content-box'
     }
   },
-  markerCircle: {
-    width: '12px',
-    height: '12px',
-    border: '2px solid transparent',
-    borderRadius: '50%'
-  },
-  markerIcon: {
-    width: '16px',
-    height: '16px'
+  icon: {
+    maskRepeat: 'no-repeat',
+    maskSize: 'cover'
   },
   flexParent: {
     display: 'flex',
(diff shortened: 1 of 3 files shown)


packages/react-ui/src/widgets/legend/LegendCategories.js
```diff
@@ -121,8 +135,25 @@ function Row({ label, isMax, isStrokeColor, color = '#000' }) {
...
+              icon ? classes.markerIcon : classes.markerCircle
+            ].join(' ')}
+            style={
+              icon
+                ? {
+                    backgroundColor: color,
+                    maskRepeat: 'no-repeat',
+                    maskSize: 'cover',
+                    maskImage: `url(${icon})`,
+                    WebkitMaskRepeat: 'no-repeat',
```
VictorVelarde, line 149: Is this really required? MaterialUI is supposed to detect & add vendor prefixes (https://v4.mui.com/styles/advanced/#css-prefixes)

zbigg: Didn't know about mui. This is not MUI, btw, this is generic react `style` property ...
For sure prefixes are needed with vanilla CSS: https://jsfiddle.net/ongwmLrd/2/
Will check it...

zbigg: It's some other magic mechanism, which understands Firefox 
image
vs Chrome:
image

and ... if we don't add `Webkit*`, it only looks like this:

image

In other words, for Chrome, supposedly unsupported properties aren't even emitted :/ Too many frameworks for me

packages/react-ui/src/widgets/legend/LegendCategories.js
```diff
@@ -76,6 +85,10 @@ const useStyles = makeStyles((theme) => ({
       boxSizing: 'content-box'
     }
   },
+  icon: {
+    maskRepeat: 'no-repeat',
+    maskSize: 'cover'
+  },
```
zbigg, line 91: @VictorVelarde 
So update on story:
 - yes, MUI adds proper prefixes, when used in `makeStyles` like above!
 - but something else (react/mui?) also requires Webkit prefixes ... and filters for inline `style={{...}}`

padawannn (approved):
LGFM

packages/react-ui/src/widgets/legend/LegendCategories.js
VictorVelarde: Great, I didn't notice those were inline styles and they were excluded. This way is better, thx!

PR #454, opened 2022-07-18
Range widget (absolute values)

padawannn:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/244150/implement-new-rangewidget-i-absolute-values

This widget can't be done with a single slider. Check the screenshot:

Captura de pantalla 2022-07-18 a las 16 02 26

- The grey bar represents the min and max value of the entire dataset.
- The light blue bar represents the min and max values taking into account the filters applied and the viewport.
- The intense blue bar represents the selected range and can overlap the other two.

So I've implemented this with three sliders but only one has the interactivity, the other two are to represent the bars effects

## Type of change

- Feature

### DIFF
--- a/packages/react-ui/src/widgets/RangeWidgetUI.js
+++ b/packages/react-ui/src/widgets/RangeWidgetUI.js
@@ -179,7 +179,7 @@ function RangeWidgetUI({ data, min, max, limits, onSelectedRangeChange }) {
 }
 
 RangeWidgetUI.propTypes = {
-  values: PropTypes.arrayOf(PropTypes.number),
+  data: PropTypes.arrayOf(PropTypes.number),
   min: PropTypes.number.isRequired,
   max: PropTypes.number.isRequired,
   limits: PropTypes.arrayOf(PropTypes.number),
--- a/packages/react-ui/src/widgets/RangeWidgetUI.js
+++ b/packages/react-ui/src/widgets/RangeWidgetUI.js
@@ -46,6 +46,17 @@ const useStyles = makeStyles((theme) => ({
   }
 }));
 
+/**
+ * Renders a <RangeWidget /> component
+ * @param  {object} props
+ * @param  {string} props.data - Array of two numbers with the selected values
+ * @param  {string} props.min - The absolute min value
+ * @param  {string} props.max - The absolute max value
+ * @param  {string} props.limits - Array of two numbers that represent a relative min and max values. It is useful to represent the min and max value taking into account other filters.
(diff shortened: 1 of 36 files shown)


packages/react-ui/src/widgets/RangeWidgetUI.js
```diff
@@ -0,0 +1,189 @@
...
+  },
+  sliderWithThumb: {
+    '& .MuiSlider-rail': {
+      color: theme.palette.text.hint
+    },
+    '& .MuiSlider-track': {
+      display: 'none'
+    }
+  },
+  sliderNotThumb: {
```
VictorVelarde, line 18: Creative approach of the underlying component, 👍🏻

packages/react-ui/src/widgets/RangeWidgetUI.js
```diff
@@ -0,0 +1,189 @@
...
+  input: {
+    maxWidth: theme.spacing(9),
+    margin: 0,
+    '& fieldset': {
+      borderWidth: 1
+    }
+  }
+}));
+
+function RangeWidgetUI({ values, min, max, limits, onSelectedRangeChange }) {
```
VictorVelarde, line 49: I think that in other cases we use `data` instead of values for the UI level (eg HistogramWidgetUI). Maybe we can change, so it’s more homogeneous with others. What do you think?

VictorVelarde: I would appreciate a small conversation together to understand this better tomorrow Javi, before we move on.

For example, why this image has a higher value at left than right?

VictorVelarde (commented):
@padawannn The general design, applying a filter ‘range’ based on new control LGTM, but I have some doubts about API (at WidgetUI level), that I would like to understand before merging. 

Also, I’ve pushed a small commit here (WIP), cause if we add a new UI component it also needs some basic storybook examples _(https://github.com/CartoDB/carto-react/commit/c61c689f781165133940b3a3ff46a208022c0c27)

packages/react-ui/src/widgets/RangeWidgetUI.js
padawannn: - I've changed value by data
- I define min & max with 2 different parameters because HistogramWidgetUI does but it's true that we don't support setting min without max, both are mandatories.
- I define limits as an array because limitMin and limitMax sound strange but we can talk about this

```
For example, why this image has a higher value at left than right?
```
The right number is greater than the left number but the right one is cut off because is too big

VictorVelarde (approved):
👍🏻

Issue #456, opened 2022-07-21
[RFC] Allow queryParameters for SQL queries


eamador:
### Target Use Case

Parameter `queryParameters` is now supported in the _SQL & Map APIs_, so we should allow this parameter to be received in `carto-react` in order to have the widgets in sync with the actual features shown in the map.

We developed a PoC to show a typical __datepicker__ use case

https://user-images.githubusercontent.com/9006480/179741156-304ef966-0d53-4893-82ac-fc6ed868f7d1.mov

### New prop
`queryParameters` need to be pass along the following packages / functions:
- **React-api**
> SQL - `executeSQL`
> Hooks - `useCartoLayerProps`

- **React-redux**
> cartoSlice - `addSource`

### Usage example
```js
{
   query: `SELECT * FROM \`heartquakes\` WHERE date >= @start AND date <= @end`,
   queryParameters: [
        {
          name: "start",
          type: "date",
          value: "2020-01-01"
        },
        {
          name: "end",
          type: "date",
          value: "2021-01-01"
        }
      ]
}
```

eamador:
I am not sure if i miss something, @padawannn could you please help here in order to define what changes we need to do and where? 

We'd also need to clarify how widgets should behave, for instance in the global mode..

Thank you!

VictorVelarde:
LGTM. On redux modifying addSource to include in queryParameters should be enough. For useCartoLayerProps, we have a `SourceProps` type where this addition fits.

VictorVelarde:
I think there is another relevant piece here to verify, just in case. When dealing with client-side widgets, this addition to sources wouldn't change the type of results, so I'm assuming client-side calculations should work the same.

But when using widgets in global mode, a request is done to //v3/sql/${source.connection}/model/${model}`; See packages/react-api/src/api/model.js and my question would be: don't we need to send also queryParameters to that endpoint? If we don't do it so, I guess that we would send just templated query in source, is that right?

padawannn:
Yes, the requests to model service need to include the `queryParameters`. The query template is only usable by builder so no service should work with the queryTemple. C4R always should work with the original query, never with the query template.

The histogram and range widgets call to the Stats API that also needs to include the queryParameters.

Remember that in builder we also call the Stats API we must include the queryParameters in these calls

eamador:
Keeping in mind this comments i [have created a draft PR](https://github.com/CartoDB/carto-react/pull/461) so that you could start a quick review and confirm it's ok.

[...]

PR #461, opened 2022-07-25
SQL queryParameters

eamador:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/245734/add-parametrized-queries-support-to-cartoreact
RFC: https://github.com/CartoDB/carto-react/issues/456

## Type of change

- Feature

# Acceptance

We should be able to send `queryParameters` along the following methods:

> SQL - executeSQL
> Hooks - useCartoLayerProps
> cartoSlice - addSource
> model - executeModel
> stats - getStats

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## Not released
-
+- Support for SQL queryParameters
 ## 1.4
 
 ### 1.4.0-alpha.1 (2022-07-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # CHANGELOG
 
 ## Not released
-- Support for SQL queryParameters
+- Support for SQL queryParameters [#461](https://github.com/CartoDB/carto-react/pull/461)
+
 ## 1.4
 
 ### 1.4.0-alpha.3 (2022-07-22)
--- a/packages/react-api/__tests__/api/model.test.js
+++ b/packages/react-api/__tests__/api/model.test.js
@@ -21,7 +21,7 @@ describe('model', () => {
       expect(mockedMakeCall).toHaveBeenCalledWith({
         credentials: TABLE_SOURCE.credentials,
         opts: { method: 'GET' },
-        url: 'https://gcp-us-east1.api.carto.com/v3/sql/carto-ps-bq-developers/model/formula?type=table&source=cartobq.public_account.seattle_collisions&params=%7B%22column%22%3A%22__test__%22%2C%22operation%22%3A%22avg%22%7D&filters=%7B%7D&filtersLogicalOperator=AND'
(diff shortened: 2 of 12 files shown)


packages/react-api/__tests__/api/model.test.js
```diff
@@ -21,7 +21,7 @@ describe('model', () => {
       expect(mockedMakeCall).toHaveBeenCalledWith({
         credentials: TABLE_SOURCE.credentials,
         opts: { method: 'GET' },
-        url: 'https://gcp-us-east1.api.carto.com/v3/sql/carto-ps-bq-developers/model/formula?type=table&source=cartobq.public_account.seattle_collisions&params=%7B%22column%22%3A%22__test__%22%2C%22operation%22%3A%22avg%22%7D&filters=%7B%7D&filtersLogicalOperator=AND'
+        url: 'https://gcp-us-east1.api.carto.com/v3/sql/carto-ps-bq-developers/model/formula?type=table&source=cartobq.public_account.seattle_collisions&params=%7B%22column%22%3A%22__test__%22%2C%22operation%22%3A%22avg%22%7D&queryParameters=&filters=%7B%7D&filtersLogicalOperator=AND'
```
VictorVelarde, line 24: Do we always need to send parameter, even if empty? Wouldn't it be clearer just avoid sending the param in the KVP if empty?

packages/react-api/__tests__/api/model.test.js
```diff
@@ -51,6 +52,7 @@ describe('model', () => {
             type: 'query',
             source: longQuerySource.data,
             params: JSON.stringify(DEFAULT_PARAMS),
+            queryParameters: '',
```
VictorVelarde, line 55: You are testing the 'empty' use case. But I think we should add some tests to see how it deals with different variations of real queryParameters. Probably it would be nice to have even 1 test variation per provider, so we are sure we  pass parameters as expected

packages/react-api/__tests__/hooks/useCartoLayerProps.test.js
```diff
@@ -23,6 +23,7 @@ describe('useCartoLayerProps', () => {
       'connection',
       'credentials',
       'clientId',
+      'queryParameters',
```
VictorVelarde, line 26: If we add something to public api, we should prepare also a small PR at doc repo, so we have this piece updated here: https://docs.carto.com/react/library-reference/api/.

We haven't decided yet if this is gonna be released in alpha mode or as patch.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## Not released
-
+- Support for SQL queryParameters
```
VictorVelarde, line 4: Add PR link here 461

VictorVelarde (commented):
@eamador I just miss a bit more of test coverage, but overall it looks good. It adds the param where expected, as discussed in RFC. 

Where did we have the issue with widgets? I don't see any evident place looking at the code

packages/react-api/__tests__/api/model.test.js
eamador: I have added a new case with a source (query) with `queryParameters` and a _parameterized_ query https://github.com/CartoDB/carto-react/pull/461/commits/2b2047ecdb32aa6488e1cdf1674595bc68deac52

VictorVelarde (approved):
👍🏻

PR #468, opened 2022-09-13
Linked datasources POC

bbecquet:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/256962

### State 

Links are stored in the linked source, i.e. the one that can be filtered as results or filtered applied on the other one, the 'foreign' source (this terminology matches the one used by the backend API on foreign filtererd).

Links can be set directly in the source parameters when adding it.
For example, to describe the `customers` source and declare it will be linked to the `stores` source so that filters applied on `stores` will have effect on `customers`:
```js
{
  id: 'customers',
  type: MAP_TYPES.TABLE,
  connection: 'bqconn-front',
  data: 'carto-dev-data.public.customers',
  foreignFilteringSource: {
    foreignSourceId: 'stores',
    foreignColumn: 'store_id',
    column: 'store_id',
  }
}
```

### Actions

Two new actions

- `ADD_SOURCE_LINK / addSourceLink`, taking `origin` and `linked` as payload, both are `{ sourceId, column }` objects
- `REMOVE_SOURCE_LINK / removeSourceLink`, taking `originSourceId` and `linkedSourceId` as payload

## Type of change

- Feature


alasarr:
I think the current proposal will force the user to do multiple steps:
```javascript
  dispatch(addSource(storesSource));
  dispatch(addSource(customersSource));
  dispatch(addSourceLink...);
```

I propose to simplify it as follows:
```javascript
  const storesSource = {
    id: STORES_SOURCE_ID,
    type: MAP_TYPES.TABLE,
    connection: 'bqconn',
    data: 'carto-demo-data.demo_tables.retail_stores',
  };
  
  const customersSource = {
    id: CUSTOMERS_SOURCE_ID,
    type: MAP_TYPES.TABLE,
    connection: 'bqconn',
    data: 'carto-demo-data.demo_tables.retail_stores',
    linkedSource: {id: STORES_SOURCE_ID, sharedColumn: 'store_id' } 
  };


 const  dispatch(addSource(storesSource));
 const  dispatch(addSource(storesSource));
```

So we only add a new property for [DataSources](https://docs.carto.com/react/guides/data-sources/), the rest is transparent. 

I'm not opposed to keep the `addSourceLink` action available too.

bbecquet:
@padawannn 

> On the other hand, we need to modify the remove source action to automatically remove links if this source was the original source or the linked source to avoid inconsistencies.

Indeed, I will add this.

> Finally, we can not forget that we need to refresh the widget every time we remove or add a link. For the moment, don't focus on this because is the last task but have it in mind

Actually this won't need anything particular as the `useWidget` hook will depend on links so update will be automatic (it already is in my example app).

bbecquet:
@alasarr About the data model…

I put a separate entry in state because I thought it was easier to manage, especially as it needs to be accessed both ways (from origin and from linked), filtered, etc. 

But we discussed with @padawannn and I understand the point that putting the property directly in the `DataSource` model is simpler for C4R-template users.
I think this is no big deal to change for now, so I will.

Compared to your example code above, I will change two things: 
 - replacing `sharedColumn` with two fields, as the column are not necessarily the same
 - changing `linkedSource` with something more explicit, as I think we have a vocabulary problem with "link" which is ambiguous to express the direction of the relationship.

Also, I have a concern about cardinality.
Javi explained to me you agreed to have only 1-1 links for now, but I think it will be a big limitation very soon…
I think having an array right away is better, even if we limit the possibility in UI.
WDYT?

bbecquet:
In the last commit I've changed the data model so that the link can be set in the datasource itself, like you proposed @alasarr.
I've made some naming changes so that it's more explicit:
```js
const customersSource = {
    // … other datasource props
    foreignFilteringSource: {
       foreignSourceId: STORES_SOURCE_ID,
       foreignColumn: 'store_id',
       column: 'store_id'
    } 
};
```
This `foreign` term is the one used by the back-end API that we call in the case of global widgets, so I thought it was good to be consistent.

I'm still worried about the 0 or 1 cardinality of this prop for the near future. But this shouldn't be too hard to turn to multiple foreign filters if needed.

Also I've implemented the filtering for viewport mode on GeoJSON sources. Most of the job is done by the `useForeignFilter` hook which basically reads the link definition and turns it into a real `IN` filter with values fetched from the foreign source. 
cc @padawannn

packages/react-api/src/hooks/useGeojsonFeatures.js
```diff
@@ -22,6 +23,8 @@ export default function useGeojsonFeatures({
 
   const sourceId = source?.id;
 
+  useForeignFilter(source);
```
padawannn, line 26: The Hook approach has some problems.
- You won't be able to move this logic to builder because the feature calculation happend in Sagas. Take a look [here](https://github.com/CartoDB/cloud-native/blob/main/workspace-www/src/features/builder/state/WidgetsSagas.ts#L357). The only possibility would be to copy and adapt this logic in Builder and is would be hard to maintain .

- I don't like the idea of having a `useEffect` to add and remove the `IN` filter. The workflow is going to be difficult to follow because the `executeTask(sourceId, Methods.GEOJSON_FEATURES, {` is executed at some point and the `useEffect` is executed at another moment.

- I know that the initial approach was to add `IN` filter but thinking about the builder use case, imagine that you have widgets linked in viewport mode and you change any widget to global mode. As you have an `IN` filter created from the link this filter will be sent to the server.

- I would like to propose another approach. I think that we could move the logic to the widget model level. Check this [line](https://github.com/CartoDB/carto-react/blob/master/packages/react-widgets/src/models/CategoryModel.js#L13). Just before to execute the `executeTask` I would execute the `Methods.FILTERED_FEATURES_VALUES` and with the returned values I would refresh the `source.filters` with destructuring (to no modify the original object in the store). I know that is less efficient because we execute  `FILTERED_FEATURES_VALUES` for each widget but the code would be cleaner, automatically would work in Builder, and probably will solve a lot of use cases. What do you think?

padawannn (changes requested):

packages/react-api/src/hooks/useGeojsonFeatures.js
bbecquet: Thanks for the review.
I'll try with the approach you propose, probably in another branch.

I understand the problem with builder… I forgot that builder doesn't use that much of C4R actually, which is a pity.

> I know that the initial approach was to add IN filter but thinking about the builder use case, imagine that you have widgets linked in viewport mode and you change any widget to global mode. As you have an IN filter created from the link this filter will be sent to the server.

In this case the other hook dedicated to global mode was replacing the filter with the expression for SQL filtering.

> I don't like the idea of having a useEffect to add and remove the IN filter. The workflow is going to be difficult to follow because the executeTask(sourceId, Methods.GEOJSON_FEATURES, { is executed at some point and the useEffect is executed at another moment.

My point was to work at the dataset/feature level, using the existing filter capabilities because I thought (and I still think in some way) they are the right tool. I don't think it would be more difficult to follow than adding special cases at the widget level. 

But once again I'm probably overlooking a lot of implications. And the builder stuff is a real block.

padawannn: Thinking more about the "add IN filter" approach I think that is going to be difficult to maintain because we need to take into account this special filter in some places. Some examples: 
- The `clear filter` action should not remove this IN filter
- In global mode, we also need to ignore this filter to not include the list of `IDs` in the request
- If a user creates a widget using the same column that we use to link we could have problems. Currently, we can differentiate it by the `owner` property but it is a very weak solution.
- If the user removes the linked source we need to remove the IN filter in other sources.
- If in the future we develop more functionalities related to filters like the first two (clear filter and global modal) we're going to have more "exceptions" like in these first two use cases.

bbecquet: @padawannn so… I keep coming back with new doubts about the tech approach.

The main thing about not working with real filters is that we don't benefit from the existing system built around them, especially to filter features on the map through Deck.gl datafilter extension.
That means we have to add stuff in this code path too, making maintenance more complicated.

For example, I've tried your proposition here:
> Check this [line](https://github.com/CartoDB/carto-react/blob/master/packages/react-widgets/src/models/CategoryModel.js#L13). Just before to execute the executeTask I would execute the Methods.FILTERED_FEATURES_VALUES and with the returned values I would refresh the source.filters with destructuring (to no modify the original object in the store).

It does the job for the widget result, but it has two main drawbacks:
 - it needs to be duplicated in every widget model
 - it doesn't filter features on the map, which should always reflect the features taken into account in the widget
That's why I think we should work as low-level as possible (at datasource/worker level) and not at the widget level.

Otherwise, I've tried another non-filter approach which is to add a new worker action specific to foreign source, to set foreign values/filtering column. Then the worker can filter when needed and it doesn't conflict with user-set filters, which is useful for clear action, etc. But it has the same drawback as above, filtering of features for the map layer has to be done separately. 

I always come back to a "real" filter that can be applied like the others, benefiting from the whole existing filter workflow.

Now regarding the points you raised above, I think most of them are related to how this auto filter would conflict with user-set ones. I think this could be solved by declaring a new type of filter, let's say `FOREIGN_IN`. That way it could 'cohabit' with another `IN` filter set on the same column, be treated like `IN` in terms of actual implementation, but be easily distinguished when needed (ex: when clearing filters, or for global mode)
I don't think it would be more complicated that all the exceptions that currently exist with the time filters…

What do you think?
I'm 100% sure I'm overlooking obvious blocks here… but honestly otherwise I feel stuck.

padawannn: I forgot the map filtering in my previous comment so you're right, the model approach is not valid. Regarding your proposal about declaring a new type of filter, I think that is a very good idea. Only a suggestion:
In c4r the only way to refresh these filter values is using a useEffect as you did [here](https://github.com/CartoDB/carto-react/pull/468/files#diff-d2205706054f73460f419c20cfaee2e46875c9555d7fe39084486cf27d7ceb58R28) and as we talked, we cannot use this useEffect in builder so we will have to move this logic to `sagas`. Is not a problem because we've already done it but my recommendation is to try to create utils functions and redux selectors that we can use in builder sagas. For example `packages/react-api/src/hooks/dataFilterExtensionUtil.js` was create to be able to use easily in builder

padawannn (changes requested):
Remember that clear filter actions should not remove the FOREIGN_IN filters

bbecquet:
@padawannn About the last changes:
- I've added a helper selector (`selectForeignFilterParams`) to simplify the recovery of parameters and foreign source from the hook
- I've added the management of the loading state, now `isSourceReady` depends on `isForeignSourceReady` if the source is filtered by another one
- this `isForeignSourceReady` flag is used as a change trigger for updating the filter values. So, regarding the management of viewport change, there's no need for any specific code added elsewhere. When viewport changes, the feature loading cycle of the source triggers `SET_FEATURES_READY` actions, which will trigger changes to `isForeignSourceReady` and then the hook responsible for setting the filter in `useWidgetFetch`.
It's working fine in my C4R-template site and locally-linked builder.

packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -152,15 +167,41 @@ export const createCartoSlice = (initialState) => {
...
+            source.filters[column] = keptForeignFilter;
+          }
         }
       },
       clearFilters: (state, action) => {
         const { id } = action.payload;
         const source = state.dataSources[id];
 
         if (source && source.filters) {
+          let keptForeignFilter;
```
padawannn, line 189: Syncronize with @Josmorsot because he's changing this action

packages/react-widgets/src/hooks/useWidgetFetch.js
```diff
@@ -15,20 +20,59 @@ export default function useWidgetFetch(
...
+
+  const isSourceReady = useSelector((state) => {
+    const selfReady = global || selectAreFeaturesReadyForSource(state, dataSource);
+    if (hasForeignFilter) {
+      return selfReady && isForeignSourceReady;
+    }
+    return selfReady;
+  });
+
+  useEffect(() => {
```
padawannn, line 38: Great work here, this `useEffect` and the code relative to `isSourceReady` are the keys so that all the system works

packages/react-widgets/src/models/FormulaModel.js
```diff
@@ -7,7 +7,7 @@ export function getFormula(props) {
 }
 
 // From local
-function fromLocal(props) {
+async function fromLocal(props) {
```
padawannn, line 10: can we remove the async

packages/react-widgets/src/hooks/useWidgetFetch.js
```diff
@@ -15,20 +20,59 @@ export default function useWidgetFetch(
...
-  const isSourceReady = useSelector(
-    (state) => global || selectAreFeaturesReadyForSource(state, dataSource)
-  );
   const source = useWidgetSource({ dataSource, id });
+  const { foreignSource, isForeignSourceReady, foreignFilterSelfColumn, foreignColumn } =
+    useSelector((state) => selectForeignFilterParams(state, source));
+  const hasForeignFilter = !!foreignSource;
+
+  const isSourceReady = useSelector((state) => {
+    const selfReady = global || selectAreFeaturesReadyForSource(state, dataSource);
```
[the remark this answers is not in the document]
bbecquet: Agreed, it would be a cleaner solution. I'll change.

VictorVelarde:
Not useful now

PR #470, opened 2022-09-20
Feature Selection tool does not work with Satellite Google base map

padawannn:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/241108/ac-z2gigeei-feature-selection-tool-does-not-work-when-terrain-or-satellite-basemap-is-set-on-the-first-action

Nebula `editable geojson layer`set the `billboard` property false by default and it looks like it is not compatible with Google Raster base maps. For our use cases, I think that this property is not necessary because is relative to the 3D rendering and the extrusion. More details [here](https://deck.gl/docs/api-reference/layers/path-layer#billboard)

This PR set the billboard property to false and add a test to check this property. Also, increase a bit the layer line with to keep the same look when we set the billboard property to false

## Type of change

- Fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Not released
 
+- Set the `EditableGeoJsonLayer` billboard property to false to be compatible with Google raster base maps
+
 ## 1.4
 
 ### 1.4.0 (2002-09-19)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Not released
 
-- Set the `EditableGeoJsonLayer` billboard property to false to be compatible with Google raster base maps
+- Fix the Feature selection tool by setting the `EditableGeoJsonLayer` billboard property to false to be compatible with Google raster base maps
 
 ## 1.4
 
--- a/package.json
+++ b/package.json
@@ -17,6 +17,8 @@
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.57",
+    "@nebula.gl/edit-modes": "^1.0.3",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -17,6 +17,8 @@
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.57",
+    "@nebula.gl/edit-modes": "^1.0.3",
```
padawannn, line 20: Added as a dev dependency to be able to add the `FeatureSelectionLayer.test.js` test

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Not released
 
+- Set the `EditableGeoJsonLayer` billboard property to false to be compatible with Google raster base maps
```
VictorVelarde, line 5: IMO it's better something like:

```suggestion
- Fix the Feature selection tool by setting the `EditableGeoJsonLayer` billboard property to false to be compatible with Google raster base maps
```

packages/react-widgets/__tests__/layers/FeatureSelectionLayer.test.js
```diff
@@ -0,0 +1,21 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { EventManager } from 'mjolnir.js';
+import FeatureSelectionLayer from '../../src/layers/FeatureSelectionLayer';
+import { mockReduxHooks } from '../mockReduxHooks';
+
+describe('FeatureSelectionLayer', () => {
+  // mockReduxHooks();
+  describe('Google Maps compatibility', () => {
```
VictorVelarde, line 8: Not sure if just in raster?

```suggestion
  describe('Google Maps Raster compatibility', () => {
```

VictorVelarde (approved):
LGTM

PR #483, opened 2022-10-06
Allow widgets to clear only its own filters

Josmorsot:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/253299/carto-3-widgets-all-reset-if-using-the-same-column

Having two or more widgets applied over the same column. When you apply and clear filters, all widgets related were reset.

In this example, both widgets are showing information based on the same column. Clearing one is not causing that the other one was reset.

https://user-images.githubusercontent.com/7818205/194520099-f09cd2df-4b8f-49ea-bda8-f88c62817629.mov

 

## Type of change
- Feature


lerna.json
```diff
@@ -4,5 +4,5 @@
   ],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "1.5.0-alpha.1"
+  "version": "1.5.0-alpha.2.filters.1"
```
VictorVelarde, line 7: @Josmorsot We don't usually use custom name alphas for c4r, as we do in kepler. So I would recommend leaving current in the PR as the first step, to merge in master, and then make another PR for the new alpha release (1.5.0-alpha.2)

VictorVelarde (commented):
LGTM, just want to confirm with @padawannn & @bbecquet about the FilterType addition in the redux actions, before landing

packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -148,10 +149,15 @@ export const createCartoSlice = (initialState) => {
         }
       },
       removeFilter: (state, action) => {
-        const { id, column } = action.payload;
+        const { id, column, type } = action.payload;
         const source = state.dataSources[id];
 
         if (source && source.filters && source.filters[column]) {
+          if (type && source.filters[column][type]) {
+            delete source.filters[column][type];
```
padawannn, line 157: After this delete I would check if `source.filters[column]` has more filters and if not I would execute `delete source.filters[column]`

padawannn (changes requested):
@Josmorsot in addition to my comment in the review have you tested in Builder? I would like to know if it can affect the map configuration persistence. I mean, try to load a map, add widgets, add filters, remove filters and reload the map

packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -148,10 +149,15 @@ export const createCartoSlice = (initialState) => {
         }
       },
       removeFilter: (state, action) => {
-        const { id, column } = action.payload;
+        const { id, column, type } = action.payload;
```
bbecquet, line 152: I'm not sure about using the `type` for that… What about using the `owner` property instead? I think it was made just for this kind of use case: identifying which filter belongs to which widget.

packages/react-redux/src/slices/cartoSlice.js
bbecquet: Agreed.

Related, after having worked for some weeks close to this topic, I think the way we store filters could be improved. The 2-levels nested structure is complicated to work with. They are complex to iterate, and we always have to bother about whether the objects exists, or when to completely removing them when they are empty…
I would love to have a flat array here, with `[ ]` as default and easy filtering.

bbecquet (changes requested):

packages/react-redux/src/slices/cartoSlice.js
Josmorsot: yes, I know that would be better to use `owner` instead of `type`. The current structure of filters is using `type` as key, `owner` is a property in the filter, we can't apply more than one filter per type, widgets only applies one type of filter.

IMO, we can delete using `owner` now but has you say in the previous comment we should consider to refactor the filters structure. Maybe using `owner` instead of `type` as key¿?

Josmorsot: @bbecquet Finally I've decided to use `owner` instead.

packages/react-redux/src/slices/cartoSlice.js
Josmorsot: done ✅

bbecquet (commented):
Sorry @Josmorsot for the late reviews. Ok for the change about `owner`, thanks!
There is just this comment from Victor to address I think https://github.com/CartoDB/carto-react/pull/483/files#r989887520, and the rebase on Changelog.

bbecquet (approved):

PR #485, opened 2022-10-07
Add Clear button to RangeWidget

Josmorsot:
# Description

Add **clear** button to reset RangeWidget filters.

https://user-images.githubusercontent.com/7818205/194521726-7b28dfdc-841f-4680-80d5-23e64f551283.mov

## Type of change
- Feature


packages/react-ui/src/widgets/RangeWidgetUI.js
```diff
@@ -138,35 +149,50 @@ function RangeWidgetUI({ data, min, max, limits, onSelectedRangeChange }) {
     onSelectedRangeChange && onSelectedRangeChange(newValues);
   };
 
+  const resetSlider = () => {
+    const newValues = [min, max];
+    setInputsValues(newValues);
+    setSliderValues([...newValues]);
```
VictorVelarde, line 155: Do we need here destructuring?

VictorVelarde (commented):
LGTM, but let's wait for the other PR to land before

packages/react-ui/src/widgets/RangeWidgetUI.js
Josmorsot: Probably not but it's used in other pieces of the component. I assumed that @padawannn found some problem and needed to use destructuring here:
https://github.com/CartoDB/carto-react/blob/8d552890186d2d34bfdac94aed4e32d6766d9217/packages/react-ui/src/widgets/RangeWidgetUI.js#L148

packages/react-ui/src/widgets/RangeWidgetUI.js
```diff
@@ -138,35 +149,50 @@ function RangeWidgetUI({ data, min, max, limits, onSelectedRangeChange }) {
     onSelectedRangeChange && onSelectedRangeChange(newValues);
   };
 
+  const resetSlider = () => {
```
bbecquet, line 152: I think there is a missing call to `onSelectedRangeChange` here, otherwise it's just resetting the UI state but not the underlying filter. And I would expect `onSelectedRangeChange` to simply remove the filter in this case.

bbecquet (changes requested):

packages/react-ui/src/widgets/RangeWidgetUI.js
```diff
@@ -133,40 +144,51 @@ function RangeWidgetUI({ data, min, max, limits, onSelectedRangeChange }) {
...
       (a, b) => a - b
     );
-    setInputsValues(newValues);
-    setSliderValues([...newValues]);
-    onSelectedRangeChange && onSelectedRangeChange(newValues);
+    changeSliderValues(newValues);
+  };
+
+  const resetSlider = () => {
+    changeSliderValues([min, max]);
```
bbecquet, line 151: Ok, that's a nice simplification.
But I think it should really remove the filter in this case. Here it will modify the filter to use min/max values. This will select all features, like if no filter were applied, but with the added cost of having to apply the useless filter to all features.

Look at how other widgets are coded: for example in [https://github.com/CartoDB/carto-react/blob/8f0b65e2d167e7e9748d7ec6ac03bf9b0f986b78/packages/react-widgets/src/widgets/CategoryWidget.js#L92-L97](CategoryWidget), when we clear the selected categories, we actually dispatch a `removeFilter` action.
Same for `BarWidget`, `HistogramWidget`, etc.

We should do the same here, modifying `onSelectedRangeChange` in `RangeWidget` to really remove the filter when the user uses the clear button.

bbecquet: After Slack conversation, it appears I overlooked that RangeWidget can be initialized with its self min/max which are already a default filter that shouldn't be removed. So, the cases to remove the filter are more complex than what I thought, and just updating it will work in every cases, so :+1: with this solution.
We can always come back to it for optimization later if needed.

bbecquet (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Update Storybook to v6.5.12 [#487](https://github.com/CartoDB/carto-react/pull/487)
```
VictorVelarde, line 6: Remove extra line here, please

VictorVelarde (approved):
LGTM

PR #486, opened 2022-10-10
Fix how time series widget manages time zones

moicalcob:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/261050/ac-7xhfwyml-carto-3-timeseries-filtering-makes-geometries-disappear

There is a problem on how we deal with dates on the time series widget. Sometimes we're receiving dates with time zones, and we're ignoring them when calling the `toDate()` function. I've added a new function so that we can get the number of milliseconds since the ECMAScript epoch, adding the proper amount of time that corresponds to the time zone.

## Type of change

- [x] Fix
- Feature
- Refactor
- Tests
- Documentation

# Acceptance

1. Add a time series widgets to a map created using a dataset containing dates with timezones.
2. Configure the time series widget to filter by hour/minutes.
3. Check that the dataset values are filtered correctly.

In order to complete the acceptance in CARTO 3 we have to complete the same steps from Builder.

# Basic checklist

- [x] Good PR name
- [x] Shortcut link
- [ ] Changelog entry
- [ ] Just one issue per PR
- [ ] GitHub labels
- [ ] Proper status & reviewers
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/react-widgets/src/widgets/TimeSeriesWidget.js
+++ b/packages/react-widgets/src/widgets/TimeSeriesWidget.js
@@ -18,6 +18,10 @@ import { columnAggregationOn } from './utils/propTypesFns';
 import useWidgetFetch from '../hooks/useWidgetFetch';
 import WidgetWithAlert from './utils/WidgetWithAlert';
 
+function getUTCTime (date) { 
+  return date.getTime() - (date.getTimezoneOffset() * 60000); 
+};
+
 // Due to the widget groups the data by a certain stepSize, when filtering
 // the filter applied must be a range that represent the grouping range.
 // For example, if you group the data by day, the range must
@@ -145,7 +149,7 @@ function TimeSeriesWidget({
             id: dataSource,
             column,
             type: FilterTypes.TIME,
-            values: [timeWindow.map((date) => date.getTime?.() || date)],
+            values: [timeWindow.map((date) => date instanceof Date ? getUTCTime(date) : date)],
             owner: id
           })
         );


packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -18,6 +18,10 @@ import { columnAggregationOn } from './utils/propTypesFns';
 import useWidgetFetch from '../hooks/useWidgetFetch';
 import WidgetWithAlert from './utils/WidgetWithAlert';
 
+function getUTCTime (date) {
```
bbecquet, line 21: I'm a bit worried about this fix.
I think at least the name of this function is misleading, as it's not returning UTC time. 
`Date.getTime` already returns a UTC ms timestamp, if we apply an offset to it we convert it to a "local timestamp".

If I understand correctly, the problem is that filter values from the widget are UTC, whereas the feature values from the source are manipulated as local dates (is it always the case?), so the filter fails.
Maybe a better place to apply a fix is at the filter level itself "https://github.com/CartoDB/carto-react/blob/57821621b1dc179327a42f0054d930ecf04b1173/packages/react-core/src/filters/FilterTypes.js#L16-L23" or in how `timeWindow` values are created?

Sorry @moicalcob, just trying to understand in more depth. TZ stuff is notoriously confusing so we need to take extra care that it will work everywhere in the world/anytime/with every dataset.
Maybe @padawannn you have some insight here too?

Shylpx:
Hi @moicalcob! Thank you for your work here! It was really useful to understand what was happening. However, as discussed today, it seems this would require a complex solution, and we would need to know if the time includes or not the time zone to properly handle each case. So I've created [this document](https://app.shortcut.com/cartoteam/write/IkRvYyI6I3V1aWQgIjY0MTgzYmJlLThlZjgtNDllMy05OGQzLWUyYjI0OTQ4Mjc2ZSI=) to be reviewed by product team.

Closing this one in the meantime :bow:

PR #498, opened 2022-10-18
New typography toolkit for MUI carto-theme

vmilan:
# Description

New typography toolkit for MUI carto-theme and added several custom variants.

## Type of change

- Feature

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -13,3 +13,9 @@
 Also added some files for shared constants (`themeConstants.js`) and useful functions (`themeUtils.js`).
 
 Removed unused custom `createTheme` function in `carto-theme.js`.
+
+# Typography
+
+Custom variant `charts` has been replaced by `caption`.
+
+`responsiveFontSizes` removed due we want to resize only a few variants through the theme.
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -16,6 +16,6 @@ Removed unused custom `createTheme` function in `carto-theme.js`.
 
 # Typography
 
-Custom variant `charts` has been replaced by `caption`.
+`responsiveFontSizes` simplified due we want to resize only a few variants through the theme.
 
-`responsiveFontSizes` removed due we want to resize only a few variants through the theme.
+Added several custom variants to the typography set. Note that MUI v5 is needed to register them properly.
--- a/packages/react-ui/src/theme/sections/typography.js
+++ b/packages/react-ui/src/theme/sections/typography.js
@@ -106,6 +106,7 @@ const baseTypography = {
 };
 
 const customTypography = {
+  // TODO: Create a wrapper for the Typography component to reduce the number of variants https://app.shortcut.com/cartoteam/story/265549/
   body1Italic: {
     ...baseTypography.body1,
     fontStyle: 'italic'
@@ -132,20 +133,6 @@ const customTypography = {
     fontWeight: 600,
     fontStyle: 'italic'
   },
-  buttonLarge: {
-    fontFamily: 'Inter, sans-serif',
-    fontWeight: 500,
-    fontSize: getPixelToRem(16),
-    lineHeight: 1.5,
-    letterSpacing: '0.25px'
-  },
-  buttonSmall: {
-    fontFamily: 'Inter, sans-serif',
-    fontWeight: 500,
-    fontSize: getPixelToRem(11),
-    lineHeight: 1.454,
-    letterSpacing: '0.4px'
-  },
   captionItalic: {
     ...baseTypography.caption,
     fontStyle: 'italic'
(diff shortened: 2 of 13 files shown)


packages/react-ui/src/theme/sections/typography.js
```diff
@@ -9,117 +11,219 @@ export const themeTypography = {
...
+  body2Strong: {
+    ...baseTypography.body2,
+    fontWeight: 600
+  },
+  body2StrongItalic: {
+    ...baseTypography.body2,
+    fontWeight: 600,
+    fontStyle: 'italic'
+  },
+  buttonLarge: {
```
VictorVelarde, line 135: Are these components-related variables gonna be here in the long-term? Is this the best place for those?

UPGRADE.md
```diff
@@ -13,3 +13,9 @@
 Also added some files for shared constants (`themeConstants.js`) and useful functions (`themeUtils.js`).
 
 Removed unused custom `createTheme` function in `carto-theme.js`.
+
+# Typography
+
+`responsiveFontSizes` simplified due we want to resize only a few variants through the theme.
+
+Added several custom variants to the typography set. Note that MUI v5 is needed to register them properly.
```
VictorVelarde, line 21: Not sure about this comment, we can talk tomorrow "Note that MUI v5 is needed to register them properly"

packages/react-ui/src/theme/themeUtils.js
```diff
@@ -2,3 +2,11 @@ import createSpacing from '@material-ui/core/styles/createSpacing';
 import { SPACING } from './themeConstants';
 
 export const spacing = createSpacing(SPACING);
+
+// Convert pixels to rem
+export function getPixelToRem(px) {
```
VictorVelarde, line 7: There should be probably a function in mui for this, but I'm ok with having our own, Probably we could switch in the future, if required

packages/react-ui/src/theme/sections/typography.js
```diff
@@ -9,117 +11,219 @@ export const themeTypography = {
...
-    fontSize: '0.625rem',
-    lineHeight: 1.6,
-    letterSpacing: '0.150em',
-    textTransform: 'uppercase',
-    fontSmoothing: 'antialiased'
+    fontSize: getPixelToRem(14),
+    lineHeight: 1.428,
+    letterSpacing: 0
+  },
+  codeMediumStrong: {
```
VictorVelarde, line 188: If we're gonna go after several 'nested' concepts here, should we create a proper object (to use keyed names)?

I'm thinking about something similar to (pseudo-code):

```
code : { 
     default: {  /* style definition */ }
     large: {
          default: { /*whatever*/}
          strong: { ... }   
     }
}
```

So we would consume externally something like: `code.large.strong`

Also I'm feeling that this set of custom elements could be maybe separated from official typography in the tree, having like a first level with `custom : { /* previous keys here */ }` Would it be too verbose? or complex? What do you think?

packages/react-ui/src/theme/sections/typography.js
```diff
@@ -9,117 +11,219 @@ export const themeTypography = {
...
+    letterSpacing: 0
+  },
+  codeSmallChip: {
+    fontFamily: '"Overpass Mono", monospace',
+    fontWeight: 600,
+    fontSize: getPixelToRem(12),
+    lineHeight: 1.25,
+    letterSpacing: 0
   },
   charts: {
```
VictorVelarde, line 216: Probably it's ok as a first step, but I would probably suggest treating this for charts in a separated object. Ideally this would land finally in something like a charts.typography (kind-of-internal) section in the future, related to charts and probably not exposed externally (thus no need to include in theme.typography.charts...).

I know this is a challenge now, but I'm sharing thoughts for future direction, we can advance step by step to avoid breaking current charts

. At least let's add a note

packages/react-ui/src/theme/sections/typography.js
```diff
@@ -9,117 +11,219 @@ export const themeTypography = {
...
-    fontSmoothing: 'antialiased'
+    fontSize: getPixelToRem(10),
+    lineHeight: 1.2,
+    letterSpacing: '2px'
   }
 };
+
+export const themeTypography = {
+  ...baseTypography,
+  ...customTypography
```
VictorVelarde, line 228: I'm not sure about including both blocks at the same level... but probably we can move on and have a longer discussion

packages/react-ui/src/theme/sections/typography.js
vmilan: They are going to use them across the platform, so I think they should be here. But you are right that button is not the best naming for a reusable variant. I'll talk to them.

UPGRADE.md
vmilan: I've added the custom variants, but they can't be used right now because they're not part of the theme.

I have been researching for a long time but I didn't find a way to register them in v4. It seems that custom variants (except colors) are not supported in v4: https://github.com/mui/material-ui/issues/15573

Since this is our next step, I added the new variants and left a note to go ahead and upgrade.

packages/react-ui/src/theme/themeUtils.js
vmilan: I didn't find anyone, but maybe you're right. I'll look into it again after the migration.

packages/react-ui/src/theme/sections/typography.js
vmilan: I've talked to them yesterday about this chart variant, because is not in the new palette (this is correct) but we don't have a replacement for it (we need one because is too specific to a particular component).

They told me that they are going to create a new one (reusable) that can replace that, so I left a TODO in the code with this situation. I didn't want no move it until we have the replacement, to see where it belongs.

Is this enough or do you think I should do something different?

packages/react-ui/src/theme/sections/typography.js
vmilan: Good point. I like what you're proposing about nesting, I'll give it a try.

vmilan:
@VictorVelarde I added some comments and cleaned up the number of variants a bit, based on our discussion and the meeting I had with the design team earlier. 

We have an [open issue](https://app.shortcut.com/cartoteam/story/265549/create-a-wrapper-for-the-mui-typography-component) in shortcut to improve the props/variants management, that we can address after we do the MUI update. Once this is done, we will be able to remove most of the new custom variants added.

packages/react-ui/src/theme/sections/typography.js
VictorVelarde: Button styles are gonna be out

VictorVelarde (approved):
🚀

Issue #479, opened 2022-10-05
[RFC] Comparative Widegts


juandjara:
# [RFC] Comparative Widegts

## Background
Many projects from PS (Site Selection, Initiative, etc) need the functionality of comparing two or more sets of widgets side to side, comparing data from many different sources in different form and aggregations.

### Initiative comparative widgets design

![Captura de pantalla 2022-10-05 a las 13 04 56](https://user-images.githubusercontent.com/6652814/194046519-3cae3916-9219-47be-abea-8033c03c2eb6.png)

![Captura de pantalla 2022-10-05 a las 13 05 11](https://user-images.githubusercontent.com/6652814/194046531-e4dafec2-52f2-4f89-b74f-5a8e42183777.png)

![Captura de pantalla 2022-10-05 a las 13 07 48](https://user-images.githubusercontent.com/6652814/194046764-4421b075-29a3-42de-92fa-c34fe2458b91.png)

### Site Selection comparative widgets design

![Captura de pantalla 2022-10-05 a las 12 22 23](https://user-images.githubusercontent.com/6652814/194045362-696a8e35-3cc3-4018-99d0-b4a7bd64bde2.png)

![Captura de pantalla 2022-10-05 a las 13 00 18](https://user-images.githubusercontent.com/6652814/194045455-358244e1-4f6c-4a0c-9d28-3ce5b5fe4148.png)

## Problem
Implementing his feature right now is not possible without ejecting from the Carto4React widgets UI and models completely and rebuilding them all from scratch. The current widgets available in the Carto4React library do not provide this functionality.

This causes that every project that PS has to built that requires that functionality must take care of building the whole widget stack from scratch every time. Every time widgets are built according to the project specific use cases and customizations, thus reducing compatiibility of widgets built between project and defeating the purpose of carto4react library of standarization of the carto brand and look and feel across all its related apps and products.

## Why do we need that
If carto4react would have a comparative widgets stack, that could be use to share the same implementation across many project, and reducing the time of development for the applications, focussing on providing real value for our customers.

## Proposal
The proposed implementation consists on developing new components for the comparative widgets, but only adding the presentation (UI) part.
Developing the logic to extract the data from the models would be left out of the scope of this RFC.

[...]

PR #504, opened 2022-10-24
Comparative Formula Widget

juandjara:
# Description

This PR implements part of the work proposed in RFC #479 

Changes added in this PR include:
* A new widget called `ComparativeFormulaWidgetUI` in the `widgets` folder
* A small npm package added to `dependencies` of `packages/react-ui/package.json` used for animations
* A new component called `AnimatedNumber` in the `custom-components` folder. This component uses the `use-animated-number` npm package and abstracts away the logic for animating a changing number that is required for both formula widget and category widget, removing almost entirely the need for `useEffect` in UI components.
* Typescript typings for both `AnimatedNumber` and `ComparativeFormulaWidgetUI`
* Prop-types validation for both `AnimatedNumber` and `ComparativeFormulaWidgetUI`
* A basic storybook story for `ComparativeFormulaWidgetUI` where all prop combinations can be tested and tweaked. A screenshot of this storybook is added as visual feedback.

Captura de pantalla 2022-10-24 a las 14 20 04

## Type of change

- Feature

# Acceptance

How to validate the feature:

1. Open the local storybook running `yarn storybook:start` in the `react-ui` folder.
2. Pick a set of properties to use for testing by tweaking the storybook configuration.
3. Use this set of properties to render a `ComparativeFormulaWidgetUI` in your sample project, by importing `ComparativeFormulaWidgetUI` from `@carto/react-ui`. You can use [this branch](https://github.com/CartoDB/carto-react-template/tree/feature/comparative-widgets) of `carto-react-template` to skip this step if you want, as a test screen is already setup there. Bear in mind that if this feature is not published to npm you will have to link your local version of `@carto/react-*` packages with `yarn link`.
4. Check that the rendered version of the widget in your sample project looks the same as it looks in the storybook.

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-ui/package.json
+++ b/packages/react-ui/package.json
@@ -72,7 +72,8 @@
     "webpack-cli": "^4.5.0"
   },
   "dependencies": {
-    "@babel/runtime": "^7.13.9"
+    "@babel/runtime": "^7.13.9",
+    "use-animate-number": "^1.0.5"
   },
   "peerDependencies": {
     "@carto/react-core": "^1.5.0-alpha.4",
--- a/packages/react-ui/src/types.d.ts
+++ b/packages/react-ui/src/types.d.ts
@@ -188,3 +188,39 @@ export type LegendRamp = {
     colors?: string | string[] | number[][];
   };
 };
+
+export type AnimationOptions = Partial<{
+  duration: number;
+  enterance: boolean;
+  direct: boolean;
+  disabled: boolean;
+  decimals: number;
+}>;
+
+type AnimatedNumber = {
+  enabled: boolean;
+  value: number;
+  options?: AnimationOptions;
+  formatter: (n: number) => React.ReactNode;
+};
+
+export type FormulaLabels = {
+  prefix?: React.ReactNode;
+  suffix?: React.ReactNode;
+  note?: React.ReactNode;
+};
+
+export type FormulaColors = {
+  [key in keyof FormulaLabels]?: string;
+} & {
+  value?: string;
+};
+
+export type ComparativeFormulaWidgetUI = {
+  data: number[];
+  labels?: FormulaLabels[];
+  colors?: FormulaColors[];
+  animated?: boolean;
+  animationOptions?: AnimationOptions;
+  formatter?: (n: number) => React.ReactNode;
+};
--- a/packages/react-ui/src/custom-components/AnimatedNumber.js
+++ b/packages/react-ui/src/custom-components/AnimatedNumber.js
@@ -1,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import useAnimateNumber from 'use-animate-number';
+
+function countDecimals(n) {
+  if (Math.floor(n) === n) return 0;
+  return String(n).split('.')[1]?.length || 0;
+}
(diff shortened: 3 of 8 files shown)


packages/react-ui/package.json
```diff
@@ -72,7 +72,8 @@
     "webpack-cli": "^4.5.0"
   },
   "dependencies": {
-    "@babel/runtime": "^7.13.9"
+    "@babel/runtime": "^7.13.9",
+    "use-animate-number": "^1.0.5"
```
VictorVelarde, line 76: Hi @juandjara is useAnimateNumber a must? Encapsulating animations in a lib might be a good idea, but I've checked it and it has **GPL-3.0-or-later** license https://www.npmjs.com/package/use-animate-number, so I assume it's not compatible with our current carto-react BSD model. Is there any alternative?

Also, not sure you're fully aware of it, but we already have our own `animateValue` function in utils/animations, been used by the current widgets. 

So my point here is that if we finally make a change, I would say we should do first a refactor PR, adding the lib and applying it in the current widgets that already have animated numbers. Otherwise we would have some widgets with a new lib and some others with custom.

efernandezleon: @juandjara I would remove this new way to use the animated number and discuss and open a new PR (if proceed). Take in mind that a PR implies only one thing, and how we've seen this PR includes more than one.

VictorVelarde: Agree. I'm not against it, but I think it deserves another separated conversation (like the license). Without it, this will be faster

juandjara: @VictorVelarde @efernandezleon as per your comments, I got the concepts separated and created a `useAnimatedNumber` hook based on the `animateValue` util. 
I have pushed the minimal amount of changes to be reviewed in PR #509

packages/react-ui/src/hooks/useAnimatedNumber.js
```diff
@@ -0,0 +1,39 @@
...
+  // starting with a -1 to supress a typescript warning
+  const requestAnimationFrameRef = useRef(-1); 
+
+  // if we want to run the animation on mount, we set the start value as 0 and animate to the start value
+  const [animatedValue, setAnimatedValue] = useState(() => animateOnMount ? 0 : value);
+
+  useEffect(() => {
+    if (!disabled) {
+      animateValue({
+        start: animatedValue || 0,
```
VictorVelarde, line 21: to be used also here... initialValue

packages/react-ui/src/types.d.ts
```diff
@@ -188,3 +188,36 @@ export type LegendRamp = {
     colors?: string | string[] | number[][];
   };
 };
+
+export type AnimationOptions = {
+  duration?: number;
+  animateOnMount?: boolean;
```
VictorVelarde, line 194: to add here the `initialValue?: number`?

packages/react-ui/package.json
VictorVelarde: We'll land first that #509, no prob

PR #506, opened 2022-10-25
New Typography component

vmilan:
# Description

Shortcut: [#265549](https://app.shortcut.com/cartoteam/story/265549/core-create-a-wrapper-for-the-mui-typography-component)

Create a Typography component to extend Mui Typography so we can have custom props to avoid having these kinds of variants:

      'body1Italic',
      'body1Strong',
      'body1StrongItalic',
      'body2Italic',
      'body2Strong',
      'body2StrongItalic',
      'captionItalic',
      'captionStrong',
      'captionStrongItalic',
      'overlineDelicate',
      'code1',
      'code1Strong',
      'code2',
      'code2Strong',
      'code3',
      'code3Strong'


packages/react-ui/src/atoms/Typography/index.js
```diff
@@ -0,0 +1,20 @@
...
+    <MuiTypography {...otherProps}>
+      <span style={{ fontWeight: weight, fontStyle: italic && 'italic' }}>
+        {children}
+      </span>
+    </MuiTypography>
+  );
+};
+
+Typography.propTypes = {
+  weight: PropTypes.oneOf([400 | 500 | 600]),
```
zbigg, line 16: I would propose to use same vocabulary as design i.e `regular`, `semiBold` ...

packages/react-ui/src/atoms/Typography/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Typography as MuiTypography } from '@mui/material';
+
+const Typography = ({ italic = false, weight, children, ...otherProps }) => {
+  return (
+    <MuiTypography {...otherProps}>
+      <span style={{ fontWeight: weight, fontStyle: italic && 'italic' }}>
```
zbigg, line 8: (i know it's MUI5, so not sure)

Isn't it possible to apply custom style on original element? Like `<MuiTypography style={...}/>` or `<MuiTypography className={ourClass}/>`.
I worry that additional dom element is just waste of resources.

packages/react-ui/src/atoms/Typography/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Typography as MuiTypography } from '@mui/material';
+
+const Typography = ({ italic = false, weight, children, ...otherProps }) => {
```
zbigg, line 5: I know it's typography PR, i expect we would have extended support for our `color`s, we've just got much more colors that MUI native `primary`, `secondary`, `textPrimary` ...

packages/react-ui/src/atoms/Typography/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
zbigg, line 1: I don't know if that folder `atoms/Typography` is going to include major submodule with several files, but looks like no.
So, i would question if there is reason to implement it in file that name says nothing - `index.js` looks like "nothing", it's just not searchable, there are hundreds of them in workspace and IMO
I would kindly propose to name it just `src/atoms/Typography.js` and only split into several files if it grows big and unmaintainable.

TLDR: I find folder with lone `index.js` inconvenient and premature optimization.

zbigg (commented):
I really like the idea.

I'm missing also `.d.ts` file, so we can have proper IDE help for using this new component.

packages/react-ui/src/atoms/Typography/index.js
vmilan: Yes, we have indeed, you're gonna be able to do that, like this:

`  <Typography color='default.background'>
    {someText}
  </Typography>`

packages/react-ui/src/atoms/Typography/index.js
vmilan: Ok, I understand, changed!

packages/react-ui/src/atoms/Typography/index.js
vmilan: Yes, sure... making some changes I left this unneeded span, but I already deleted it. Thank you!

packages/react-ui/src/atoms/Typography/index.js
vmilan: Absolutely, I'll do it.

packages/react-ui/storybook/stories/widgetsUI/WrapperWidgetUI.stories.js
```diff
@@ -1,10 +1,8 @@
 import React from 'react';
-import { Typography } from '@mui/material';
 import ColorizeIcon from '@mui/icons-material/Colorize';
 import MenuIcon from '@mui/icons-material/Menu';
 import AddLocationIcon from '@mui/icons-material/AddLocation';
 import WrapperWidgetUI from '.../../../src/widgets/WrapperWidgetUI';
```
vmilan, line 5: Unused imports

packages/react-ui/src/atoms/Typography.js
```diff
@@ -0,0 +1,31 @@
...
+
+const Typography = forwardRef(({ italic, weight, children, ...otherProps }, ref) => {
+  // forwardRef needed to be able to hold a reference, in this way it can be a child for some Mui components, like Tooltip
+  // https://mui.com/material-ui/guides/composition/#caveat-with-refs
+
+  return (
+    <MuiTypography
+      {...otherProps}
+      ref={ref}
+      style={{ fontWeight: FontWeight[weight], fontStyle: italic && 'italic' }}
```
zbigg, line 19: Please include also `style` prop from user

In this shape, user is completly unable to overwrite `style` as we're ignoring /overwriting it.

packages/react-ui/storybook/stories/atoms/Palette.stories.js
```diff
@@ -59,13 +60,19 @@ const ColorBox = ({ colorVariant, colorName }) => {
...
     <Box>
       <Box className={classes.text}>
         <Typography variant='subtitle1'>{colorName}</Typography>
         <Tooltip title={colorValue} enterDelay={600}>
-          <Typography variant='caption' noWrap className={classes.textValue}>
+          <Typography
+            variant='caption'
+            noWrap
+            className={classes.textValue}
+            ref={textRef}
```
Josmorsot, line 74: why are you adding ref here?

packages/react-ui/storybook/stories/atoms/Typography.stories.js
```diff
@@ -28,6 +28,33 @@ const options = {
...
+      defaultValue: false,
+      control: {
+        type: 'boolean'
+      }
+    },
+    color: {
+      description: 'Any theme color, default and custom ones',
+      control: {
+        type: 'select',
+        options: ['primary.dark', 'brand.navyBlue', 'default.main']
```
zbigg, line 48: Where is definition of all those colors available ?

(again, Typography.d.ts would be great)

Josmorsot (approved):
Nice work! LGTM 🚀

packages/react-ui/storybook/stories/atoms/Palette.stories.js
vmilan: Because of this: https://mui.com/material-ui/guides/composition/#caveat-with-refs

Tooltip (and a few others) needs to pass the ref to its children. I had to add the `forwardRef` in Typography to do that because is a custom component.

Without this, I was having this warning in the console:
`Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

packages/react-ui/storybook/stories/atoms/Typography.stories.js
vmilan: Are coming from the [theme](https://github.com/CartoDB/carto-react/blob/feature/design-system/packages/react-ui/src/theme/carto-theme.js) directly, you can use any color that we have declared there. 

This is just a small example to have in the story, just to know you can change the color via props in the Typography component.

packages/react-ui/src/atoms/Typography.js
vmilan: ok!

VictorVelarde: Just putting the otherProps at the end would be enough, wouldn't it?

packages/react-ui/storybook/stories/atoms/Typography.stories.js
VictorVelarde: I guess that should be expressed in the typings

packages/react-ui/src/atoms/Typography.js
zbigg: > Just putting the otherProps 

@VictorVelarde 
It wouldn't work, 
 - `...otherProps` at end would overwrite this component style
 - style from `...otherProps` at beginning is overwritten in this version

Proper "merging" of properties "touched" by component must be manual, like this:

```jsx
function Outer({foo, style, className, ...otherProps}) 
{
    return <Inner 
       {...otherProps}
       className={clsx(foo && 'outerRoot', className)} 
       style={{color: 'red', ...style}}
    />
}
```

vmilan: Done, now styles overrides are working fine.

VictorVelarde: Yeah, I know about override behaviour @zbigg ... my point here is Does it make sense to restrict _otherProps_ just to **style** prop and be explicit about it in the code? 

I was assuming here that if someone wants to play with _style_, beyond defaults, they can provide externally all they require (so no mix, but simple overriding), and allowing us to play with other props. Eg, check Typograghy options in https://mui.com/material-ui/api/typography/ and think about `gutterBottom`, `noWrap` or others.

packages/react-ui/storybook/stories/atoms/Typography.stories.js
vmilan: color is common prop to Mui components (it's a [system property](https://mui.com/system/properties/#properties-reference-table))... do you want to restrict it to only theme colors? 

Right now, from what I've seen, it accepts any string. I mean, wheter a primaryText or a #000000, for example.

I think should be enough as it is, but let me know.

packages/react-ui/src/atoms/Typography.js
zbigg: > I was assuming here that if someone wants to play with style, beyond defaults, they can provide externally all they require 

I would assume differently and would expect our "design system" components behave like proper MUI components. In this case, that "user provided style" and "props-based style" are properly merged _if_ they reasonably not collide.

For example with normal `Typography`, if you say `<Typography "color="primary" style={{cursor: "help"}}>` it just works and we don't care how typography "implements" color.

Nor, we're required to implement "color" by ourselves if we use "style". 

> provide externally all they require  

That would require user to know that `weight: "normal"` translates to `style={{fontWeight: 500}}` and that removes half of usefulness of this component.
(and workarounds for this would add even more api surface, like helper functions, exposing FontWeight mapping ...)

I think that robust APIs should ensure what unrelated features are indeed unrelated (as long as it's possible).

zbigg (approved):
LGTM, Apart from all comments (mostly explorative) i like it very much and we should make more components like this.

packages/react-ui/storybook/stories/atoms/Typography.stories.js
zbigg: Ok, now i see that in MUI5 it's much more robust than in MUI4. In MUI 4, only very few hardcoded colors were accepted ('primary', ...)

Thanks for explanation.

vmilan:
> LGTM, Apart from all comments (mostly explorative) i like it very much and we should make more components like this.

Thanks so much for the feedback, it really helps me.

PR #507, opened 2022-10-25
Spacings, borders and shadows Foundations

vmilan:
# Description

Shortcut: [#266310](https://app.shortcut.com/cartoteam/story/266310/core-adapt-others)

The rest of core elements beyond Colors / Typography:

- Spacings
- Borders
- Shadows


packages/react-ui/src/theme/sections/components.js
```diff
@@ -588,23 +588,23 @@ export const themeComponentsOverrides = {
...
     }
   },
 
   MuiListItemAvatar: {
     root: {
       '& .MuiAvatar-root': {
-        height: spacing(4.5),
-        width: spacing(4.5)
+        height: getSpacing(4.5),
+        width: getSpacing(4.5)
```
VictorVelarde, line 604: This is the type of 'strange' spacing numbers I was talking about... We should consider probably a rounding to the next 'accepted level'. I would suggest creating a new task to discuss with Design about it

packages/react-ui/storybook/.storybook/preview.js
```diff
@@ -37,7 +37,7 @@ export const parameters = {
     storySort: {
       order: [
         'Introduction',
-        'CARTO Theme',
+        'Atoms',
```
VictorVelarde, line 40: Let's validate tomorrow, but ok

VictorVelarde (approved):
@vmilan 🚀

packages/react-ui/src/theme/sections/components.js
vmilan: No, I added it this way to simplify, but you're right. I've just updated this file to be more granular. Let me know if it's ok.

vmilan: Here I just updated the function we use, not the value, but for sure we need to review every spacing we are using to adapt to the new ones that the design team defined. I was thinking of doing it when we start adapting each component.

PR #508, opened 2022-10-26
Fix legend when layer is outline Point

Josmorsot:
# Description

Shortcut: ([link](https://app.shortcut.com/cartoteam/story/266499/ac-7xhfwyml-carto-3-legend-not-rendering-outline-point-values))

Was broken here: https://github.com/CartoDB/carto-react/pull/451

## Type of change

- Fix

### DIFF
--- a/packages/react-ui/src/widgets/legend/LegendCategories.js
+++ b/packages/react-ui/src/widgets/legend/LegendCategories.js
@@ -80,6 +80,7 @@ const useStyles = makeStyles((theme) => ({
     position: 'relative'
   },
   circle: {
+    border: '2px solid transparent',
     '&::after': {
       position: 'absolute',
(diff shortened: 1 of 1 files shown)


packages/react-ui/src/widgets/legend/LegendCategories.js
```diff
@@ -80,6 +80,7 @@ const useStyles = makeStyles((theme) => ({
     position: 'relative'
   },
   circle: {
+    border: '2px solid transparent',
```
VictorVelarde, line 83: And that's it? Wow, nice fix 😄

VictorVelarde: I guess you tested it locally in Builder, right?

VictorVelarde (approved):
🚀

PR #513, opened 2022-10-28
Design System breakpoints and theme types

vmilan:
# Description

Shortcut: [#2666310](https://app.shortcut.com/cartoteam/story/266310/core-adapt-others)
[sc-2666310]

Design System breakpoints and theme types


packages/react-ui/src/theme/themeConstants.js
```diff
@@ -1,2 +1,11 @@
...
 export const SPACING = 8;
+
+// Breakpoints
+export const BREAKPOINTS = {
+  XS: 320,
+  SM: 600,
+  MD: 960,
+  LG: 1280,
+  XL: 1600
+};
```
VictorVelarde, line 11: If we have just a few, maybe, just maybe, it's better to see the values in the carto-theme.js... What do you think?
I see it a bit like too many indirection levels

VictorVelarde (approved):
LGTM. Types look better, indeed!

PR #519, opened 2022-11-02
Fix TileLayer unauthorized error due to breaking changes in deck.gl

Josmorsot:
# Description

Shortcut: [(link)](https://app.shortcut.com/cartoteam/story/268274/c4r-add-authorization-header-when-fetching-tiles)

TileLayer cause `unauthorized` error due to [breaking changes](https://github.com/visgl/deck.gl/pull/7287) in `deck.gl`.

## Type of change

- Fix


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Not released
 
+- Add `Authorization header` fetching tiles [#519](https://github.com/CartoDB/carto-react/pull/519)
```
VictorVelarde, line 5: You always have to describe it in terms of Fix | Add | Remove xxxx feature.... so Add is not 100% clear. It's better to use

```suggestion
- Fix TileLayer unauthorized error due to [breaking changes](https://github.com/visgl/deck.gl/pull/7287) in deck.gl >=8.9.0-alpha.3 and >=8.8.15 [#519](https://github.com/CartoDB/carto-react/pull/519)
```

packages/react-api/src/hooks/useTileFeatures.js
```diff
@@ -157,12 +157,12 @@ export default function useTileFeatures({
 }
 
 // WORKAROUND: To read headers and know if the tile is dropping features.
```
VictorVelarde, line 159: What's the condition to finally remove this work-around in the future? Can we better describe it @Josmorsot ?
Ping @felixpalmer if required, to have the next step 100% clear

Josmorsot: we have no condition to remove this. For now it's the only way to do it.

Issue #499, opened 2022-10-19
queryParameters


Fr4nSG:
Previous version of C4R: `1.3.0`
Actual version of C4R: `1.5.0-alpha.1`

**executeSQL** from **@carto/react-api**
### Error:
````
{
  "error": "Invalid queryParameters, it must be a list of parameters",
  "status": 400,
  "message": "Invalid queryParameters, it must be a list of parameters"
}
````

### Payload:
````
{
  "client": "carto-react",
  "q": "SELECT ....",
  "queryParameters": "[]",
  "access_token": "eyJhbGciOiJSUzI1NiIsInR..."
}
````

Seems like in the new version, the queryParameters is required to specify any value to avoid 400 error like this:
````
 const data = await executeSQL<GetCategoriesRes | Record<string, number>[]>({
    // @ts-ignore
    credentials,
    query,
    connection,
    queryParameters: null, // TODO: avoid error
    opts: { abortController },
  });
````

I just check inside and i found the posible bug is on the `SQL.js` `function createRequest`. (line.58)
````
function createRequest({ credentials, connection, query, opts = {}, **queryParameters = []** }) {
  const { abortController, ...otherOptions } = opts;

  const { apiVersion = API_VERSIONS.V2 } = credentials;

  const rawParams = {
    client: CLIENT,
    q: query?.trim(),
    ...otherOptions,
    queryParameters: queryParameters ? JSON.stringify(queryParameters) : undefined
  };
````

If i downgrade the versions i don't have this issue.

**Comment by Josmorsot** [comment]
Related with #524. Fixed in #525

**Comment by Josmorsot** [comment]
✅  **v1.4.4** released

**Comment by jantolg** [comment]
@Josmorsot I think is not resolved. Problem is in:

The problem is when is not include any queryParameters, function will include **by default []**, the evaluation will be true and this one will include **queryParameters: '[]'**, in case queryParemeters will be **false, null or undefined**, evaluation will return **undefined** and this will be included in queryParamseters as **queryParameters: undefined** . Both cases will return error from API

[Source: https://github.com/CartoDB/carto-react/blob/master/packages/react-api/src/api/SQL.js](https://github.com/CartoDB/carto-react/blob/master/packages/react-api/src/api/SQL.js#L67): 

```
function createRequest({ credentials, connection, query, opts = {}, queryParameters = [] }) {
  const { abortController, ...otherOptions } = opts;

  const { apiVersion = API_VERSIONS.V2 } = credentials;

[...]

PR #531, opened 2022-11-14
Fix executeSQL through POST requests

Josmorsot:
# Description

Shortcut: #499

executeSQL function doesn't work properly through POST

## Type of change

- Fix


eamador (approved):
LGTM 🚀

PR #538, opened 2022-11-23
TextField and TextArea components adapted to the new design system

vmilan:
# Description

[sc-268673]

- TextField and TextArea components  with the new design system applied.
- Custom component to create Password Fields
- Also added some minor details to Tabs stories.

### DIFF
--- a/packages/react-ui/src/components/atoms/PasswordField.js
+++ b/packages/react-ui/src/components/atoms/PasswordField.js
@@ -1,9 +1,8 @@
 import React, { forwardRef, useState } from 'react';
-import PropTypes from 'prop-types';
 import { IconButton, InputAdornment, TextField } from '@mui/material';
 import { VisibilityOffOutlined, VisibilityOutlined } from '@mui/icons-material';
 
-const PasswordField = forwardRef(({ style, ...otherProps }, ref) => {
+const PasswordField = forwardRef(({ ...otherProps }, ref) => {
   // forwardRef needed to be able to hold a reference, in this way it can be a child for some Mui components, like Tooltip
   // https://mui.com/material-ui/guides/composition/#caveat-with-refs
   const [showPassword, setShowPassword] = useState(false);
@@ -13,7 +12,6 @@ const PasswordField = forwardRef(({ style, ...otherProps }, ref) => {
     <TextField
       {...otherProps}
       ref={ref}
-      style={style}
       type={showPassword ? 'text' : 'password'}
       InputProps={{
         endAdornment: (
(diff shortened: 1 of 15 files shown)


packages/react-ui/src/components/atoms/PasswordField.js
```diff
@@ -0,0 +1,35 @@
...
+      ref={ref}
+      style={style}
+      type={showPassword ? 'text' : 'password'}
+      InputProps={{
+        endAdornment: (
+          <InputAdornment position='end'>
+            <IconButton onClick={handleClickShowPassword}>
+              {showPassword ? <VisibilityOutlined /> : <VisibilityOffOutlined />}
+            </IconButton>
+          </InputAdornment>
```
VictorVelarde, line 24: Ok, I see the value here. At the beginning I thought we were just going to manage 'password' mode as hardcode. Obviously this makes more sense

packages/react-ui/src/components/atoms/PasswordField.js
```diff
@@ -0,0 +1,35 @@
+import React, { forwardRef, useState } from 'react';
+import PropTypes from 'prop-types';
+import { IconButton, InputAdornment, TextField } from '@mui/material';
+import { VisibilityOffOutlined, VisibilityOutlined } from '@mui/icons-material';
+
+const PasswordField = forwardRef(({ style, ...otherProps }, ref) => {
```
VictorVelarde, line 6: Why do you extract here 'style' vs 'otherProps'

packages/react-ui/src/types.d.ts
```diff
@@ -219,3 +219,8 @@ type TooltipDataProps = {
   ];
   title?: string;
 };
+
+// PasswordField
+export interface PasswordFieldProps {
```
VictorVelarde, line 224: do we really need to declare this interface, couldn't we just allow same TextField props?

VictorVelarde: See this custom TextField here: https://mui.com/material-ui/react-text-field/#customization (eg  RedditTextField). It basically passes down all props, using **TextFieldProps** .

What do you think?

packages/react-ui/src/components/atoms/PasswordField.js
vmilan: you are right, is not needed. Removed!

packages/react-ui/src/types.d.ts
vmilan: yes, totally, removed.

vmilan: I also cleaned up Button, which had the same unnecessary interface.

PR #541, opened 2022-11-29
Rendering performance improvements

felixpalmer:
# Description

Performance fixes related to https://github.com/CartoDB/carto-kepler/pull/132

- Don't needlessly render FeatureSelectionLayer when there is no mask
- Add a `dataComparator` to avoid layer state updates when mask polygon isn't changing

## Type of change

- Fix

# Acceptance

Smoke tests related to the mask tool. Create/destroy a few masks with/without layers.

### DIFF
--- a/packages/react-widgets/src/layers/FeatureSelectionLayer.js
+++ b/packages/react-widgets/src/layers/FeatureSelectionLayer.js
@@ -63,60 +63,61 @@ export default function FeatureSelectionLayer(
 
   return [
     mask && MaskLayer(),
-    // @ts-ignore
-    new EditableCartoGeoJsonLayer({
-      eventManager: customEventManager,
-      id: 'FeatureSelectionLayer',
-      // billboard need to be false to be compatible with Google Satellite
-      billboard: false,
-      pickable: !!selectedMode,
-      data: {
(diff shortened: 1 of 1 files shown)


bbecquet (approved):
Simple and efficient, LGTM

packages/react-widgets/src/layers/FeatureSelectionLayer.js
```diff
@@ -63,60 +63,61 @@ export default function FeatureSelectionLayer(
...
+        eventManager: customEventManager,
+        id: 'FeatureSelectionLayer',
+        // billboard need to be false to be compatible with Google Satellite
+        billboard: false,
+        pickable: !!selectedMode,
+        data: {
+          type: 'FeatureCollection',
+          features: spatialFilterGeometry ? [spatialFilterGeometry] : []
+        },
+        dataComparator: (data, oldData) => data.features[0] === oldData.features[0],
```
VictorVelarde, line 77:
```suggestion
        // has mask polygon changed?
        dataComparator: (data, oldData) => data.features[0] === oldData.features[0],
```

VictorVelarde (approved):
Great!

PR #545, opened 2022-12-02
UploadField component adapted to the new design system

vmilan:
# Description

[sc-268673]

- UploadField component with the new design system applied.
- Removed InputFile (is replaced by UploadField)
- Also added some minor details in Select (design review).


packages/react-ui/src/components/molecules/UploadField.js
```diff
@@ -0,0 +1,198 @@
...
+      />
+    </>
+  );
+}
+
+UploadField.defaultProps = {
+  placeholder: 'Drag and drop your file or click to browse',
+  dragPlaceholder: 'Drop your file here...',
+  buttonText: 'Browse',
+  accept: 'application/JSON',
```
VictorVelarde, line 176: Is this comma separated, in the case of various types? Should use the opportunity to have accept as an array externally?

packages/react-ui/src/types.d.ts
```diff
@@ -216,7 +213,7 @@ type TooltipDataProps = {
 };
 
 // SelectField
-export interface SelectFieldProps {
+export interface SelectFieldProps extends TextFieldProps {
```
VictorVelarde, line 216: 👍🏻

packages/react-ui/src/components/molecules/UploadField.js
```diff
@@ -0,0 +1,198 @@
...
+
+  return (
+    <>
+      <TextField
+        {...props}
+        ref={textFieldRef}
+        variant={props.variant}
+        placeholder={dragPlaceholderText}
+        value={filesText}
+        label={props.label}
```
VictorVelarde, line 124: All props keeping names should be transmitted with {...props}, right? so I think we can remove:

variant={props.variant}
label={props.label}
helperText={props.helperText}
error={props.error} 

right?

vmilan: I need the placeholder and error because I have to modify them depending on the component is onDrag, the rest you mentioned can be removed, you are right

packages/react-ui/src/components/molecules/UploadField.js
vmilan: I updated the prop, I think I got you right, if not feel free to modify anything. Thank you!

PR #548, opened 2022-12-12
Design changes in UploadField component

vmilan:
# Description

[sc-268673]

- Modifications in UploadField (later design changes in Figma):
    - Added some logic to placeholder text, based on multiple bool and draggable action.
    - Button size based on input size.
- Clean up props


packages/react-ui/src/components/molecules/UploadField.js
```diff
@@ -111,14 +114,26 @@ function UploadField(props) {
...
+  const setPlaceholder = () => {
+    const multipleSuffix = props.multiple ? '(s)' : '';
+    const placeholder = `Drag and drop your file${multipleSuffix} or click to browse`;
+    const dragPlaceholder = `Drop your file${multipleSuffix} here...`;
+
+    let placeholderText;
+    if (dragOver) {
+      placeholderText = dragPlaceholder;
+    } else {
+      placeholderText = props.placeholder || placeholder;
```
bbecquet, line 126: Is it on purpose that `props.placeholder` is still used, whereas it has been removed from propTypes and TS? Looks like a mistake.

BTW, this change means we won't be able to redefine placeholder from this component, being stuck with English and these predefined texts. Is it ok?

packages/react-ui/src/components/molecules/UploadField.js
```diff
@@ -111,14 +114,26 @@ function UploadField(props) {
     setFilesText('');
   };
 
-  const dragPlaceholderText = dragOver ? props.dragPlaceholder : props.placeholder;
+  const setPlaceholder = () => {
```
bbecquet, line 117: I would change the naming here, by convention `setFoo` evokes something that will change some `Foo` state somewhere. And this looks more like a `getPlaceHolder` to me. Maybe a better way still is a `useMemo` depending on `multiple` and `dragOver`.

packages/react-ui/src/components/molecules/UploadField.js
```diff
@@ -167,23 +182,15 @@ function UploadField(props) {
...
   files: [],
   onChange: (files) => files
 };
 
 UploadField.propTypes = {
-  placeholder: PropTypes.string,
-  dragPlaceholder: PropTypes.string,
   buttonText: PropTypes.string,
   accept: PropTypes.array,
-  multiple: PropTypes.bool,
```
bbecquet, line 185: `multiple` is still used in the component, so shouldn't be removed here.
Also, looks like it's missing `size`.

packages/react-ui/src/types.d.ts
```diff
@@ -225,13 +225,10 @@ export interface SelectFieldProps extends TextFieldProps {
   size?: 'small' | 'medium';
 }
 
+// UploadField
 export interface UploadFieldProps extends TextFieldProps {
-  placeholder?: string;
-  dragPlaceholder?: string;
   buttonText?: string;
   accept?: string[];
-  multiple?: boolean;
```
bbecquet, line 233: Same comment as for propTypes: `multiple` is still used in the component, so shouldn't be removed here.

packages/react-ui/src/components/molecules/UploadField.js
```diff
@@ -47,23 +48,25 @@ const useStyles = makeStyles((theme) => ({
   }
 }));
 
-function UploadField(props) {
+function UploadField({ buttonText, accept, files, ...props }) {
```
bbecquet, line 51: For consistency, could be worth destructuring the other ones: `multiple`, `onChange` and `size`

vmilan: I destructured these ones to avoid some console warnings regarding propagation of props:

`React does not recognize the 'buttonText' prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `buttontext` instead. If you accidentally passed it from a parent component, remove it from the DOM element`

`buttonText, accept, files` are custom props, the other ones are from Mui TextField, do you think it's better to destructure them too?

packages/react-ui/src/components/molecules/UploadField.js
bbecquet: Sorry, my mistake about `placeholder`, I hadn't followed the TypeScript base types. So it's ok to use it here. But I think it should be left in proptypes.

And for the second part of my comment, we accept `dragPlaceholder` can't be redefined, then?

VictorVelarde: Proptypes are independent from TypeScript interface about props, yes. Ideally we would do a similar extend on root's proptypes if desired, but I searched for it and it looks like it's not feasible and not required (https://github.com/mui/material-ui/issues/14790). 

Indeed, and this might be a bit controversial, if we have good props definition, I feel like propTypes are probably less relevant now than they were in the past, assuming most our clients will work with TS. Nice little reading: https://medium.com/smallcase-engineering/proptypes-vs-types-in-typescript-45433990a8f9

So if we have to choose I would invest more effort now in TS than in propTypes. Let's try to be pragmatic with the second.

packages/react-ui/src/components/molecules/UploadField.js
```diff
@@ -111,14 +114,26 @@ function UploadField(props) {
...
+    return placeholderText;
+  };
 
   return (
     <>
       <TextField
         {...props}
         ref={textFieldRef}
-        placeholder={dragPlaceholderText}
+        placeholder={setPlaceholder()}
```
VictorVelarde: Benjamin is totally right, this is a getPlaceholder()

vmilan: It's already changed and using useMemo as he suggests 👍

packages/react-ui/src/components/molecules/UploadField.js
vmilan: `dragPlaceholder` it's only shown when you are dragging a file over the input, I think we don't need to change that copy and it's different from the regular placeholder (this can be redefined by props). Anyway, I'll confirm with design.

Regarding ENG translation, I've seen in cloud-native that we use `react-intl` lib. But I didn't see the same in C4R, in many places copies are hardcoded. There is something similar in C4R?

packages/react-ui/src/components/molecules/UploadField.js
vmilan: I recovered back the PropTypes you mentioned 👍

packages/react-ui/src/components/molecules/UploadField.js
vmilan: I've ping design twice with no luck. I would say to leave it as it is (dragPlaceholder) for now and review it in the future if design has a strong opinion on it.

vmilan: Confirmed by design: `dragPlaceholder` is not going to change. So it doesn't need to be accessible by props. The copy is going to be the same in every case.

VictorVelarde (approved):
👍🏻

PR #549, opened 2022-12-13
Fix/gmaps channel

aaranadev:
Currently, C4R uses a beta channel of Google Maps and it's unstable and causes the load script to fail, we need to change to a stable channel.

image (1)


VictorVelarde:
Thx for the fix. I've updated it a bit in another PR, so closing this one

PR #573, opened 2023-01-19
Fix/comparative formula UI

aaranadev:
# Description

Improve the ComparativeFormulaUI props interface with a simplification and an improvement of the docs and test.

### DIFF
--- a/packages/react-ui/src/widgets/comparative/ComparativeFormulaWidgetUI.js
+++ b/packages/react-ui/src/widgets/comparative/ComparativeFormulaWidgetUI.js
@@ -84,7 +84,7 @@ function ComparativeFormulaWidgetUI({
                 </Typography>
               </Box>
             ) : null}
-            <Box fontWeight={!i && 'bold'}>
+            <Box fontWeight={!i ? 'bold' : ''}>
               <AnimatedNumber
                 value={d.value || 0}
                 enabled={animated}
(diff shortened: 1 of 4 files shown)


juandjara (approved):
LGTM

packages/react-ui/src/widgets/comparative/ComparativeFormulaWidgetUI.js
```diff
@@ -58,93 +56,93 @@ function ComparativeFormulaWidgetUI({
...
-    return labels[index] || {};
-  }
+  const processedData = useMemo(
+    () =>
+      data
+        .map((d, i) => ({
+          ...d,
+          color: getColor(colors, i)
+        }))
+        .filter((d) => d.value !== undefined),
```
VictorVelarde, line 66: You're doing here first .map and then .filter, vs previous .filter > .map. Is it intended @aaranadev?

VictorVelarde (approved):
LGTM. We can land it after your filter / map confirmation

packages/react-ui/src/widgets/comparative/ComparativeFormulaWidgetUI.js
aaranadev: Yes, we need to remove the undefined values, but we need to assign the correct color index before the element is removed it.

Example:

If the second value is undefined, we need the value after the first to have the third color and not the second.

PR #585, opened 2023-01-25
Adjustments in theme to fix uses cases in cloud-native

vmilan:
# Description

Shortcut: [#280106](https://app.shortcut.com/cartoteam/story/280106/workflows-general-dragging-area)
sc: [sc-280106]

Includes
- Snackbar default props fix and basic story added
- Dialog tweaks in default props and styles
- Added some help in the upgrade file that we are using in CN migration and will be helpful to PS


VictorVelarde (approved):
👍🏻

PR #592, opened 2023-02-02
Add bearer token in headers instead of using query params

moicalcob:
# Description

Shortcut: (https://app.shortcut.com/cartoteam/story/285013/sql-api-requests-in-c4r-passing-the-token-in-the-url)

We noticed that requests to the APIs V3 were including the authorization token as a query parameter instead of using a header. This change should change this behaviour.

## Type of change

- Fix
- [x] Feature
- Refactor
- Tests
- Documentation

# Acceptance

1. Create a React project and add the @carto/react-api dependency
2. Call to the `executeSQL()` method using both a simple query and a complex one that will be performed through a POST request
3. Check that the GET request contains the authorization token in the request's headers:
Screenshot 2023-02-02 at 14 59 57

4. Check that the POST request contains the authorization token in the request's headers:
Screenshot 2023-02-02 at 14 59 39

# Basic checklist

- [x] Good PR name
- [x] Shortcut link
- [x] Changelog entry
- [x] Just one issue per PR
- [x] GitHub labels
- [ ] Tests
- [x] Documentation

### DIFF
--- a/packages/react-api/src/api/SQL.js
+++ b/packages/react-api/src/api/SQL.js
@@ -108,13 +108,14 @@ function createRequest({
   });
 
   const isGet = getUrl.length < REQUEST_GET_MAX_URL_LENGTH;
-  if (isGet && apiVersion !== API_VERSIONS.V3) {
-    return getRequest(getUrl, requestOpts);
-  }
-  if (isGet && apiVersion === API_VERSIONS.V3) {
-    return getRequest(getUrl, requestOpts, {
-      Authorization: `Bearer ${credentials.accessToken}`
-    });
+  if (isGet) {
+    if (apiVersion === API_VERSIONS.V3) {
+      return getRequest(getUrl, requestOpts, {
+        Authorization: `Bearer ${credentials.accessToken}`
+      });
+    } else {
+      return getRequest(getUrl, requestOpts);
+    }
   }
 
   // Post request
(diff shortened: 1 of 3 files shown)


VictorVelarde (approved):
LGTM, but I would appreciate a quick double-check from someone on backend, and then we'll land it.

Nit: little update to be made, as we've recently landed small change on CLIENT_ID and it's causing a conflict @moicalcob

packages/react-api/src/api/SQL.js
```diff
@@ -129,7 +129,9 @@ function createRequest({
     connection,
     parameters: urlParamsForPost
   });
-  return postRequest(postUrl, payload, requestOpts);
+  return postRequest(postUrl, payload, requestOpts, {
```
thedae, line 132: One question, I don't see the check for V3 here, so I guess we're always sending the bearer token? Even for V1 and V2?

packages/react-api/src/api/SQL.js
```diff
@@ -110,15 +108,17 @@ function createRequest({
...
   const isGet = getUrl.length < REQUEST_GET_MAX_URL_LENGTH;
-  if (isGet) {
+  if (isGet && apiVersion !== API_VERSIONS.V3) {
     return getRequest(getUrl, requestOpts);
   }
+  if (isGet && apiVersion === API_VERSIONS.V3) {
+    return getRequest(getUrl, requestOpts, {
+      Authorization: `Bearer ${credentials.accessToken}`
+    });
+  }
```
thedae, line 118: This could be a little bit more readable:

```
if (isGet) {
    if (apiVersion === API_VERSIONS.V3) {
        // request with header
    } else {
        // request without header
    }
}
```

thedae (changes requested):
Looking good, a question and a minor readability improvement

Thanks for taking care of this

packages/react-api/src/api/SQL.js
moicalcob: Oh, thanks so much! I didn't take that into account 🙇🏼

thedae (approved):

PR #593, opened 2023-02-03
Add sortByColumnType param to fix number sorting in TableWidget

eamador:
# TableWidget - sortByColumnType param + number format

Shortcut: https://app.shortcut.com/cartoteam/story/286648/ac-x5e68ikj-table-widget-ordering-is-wrong-for-numeric-data-type-column

TableWidget accepts a new `sortByColumnType` parameter, if its value is ~**integer** or **float**~ 'number' we will apply number sorting

https://user-images.githubusercontent.com/9006480/216548142-20af37be-0c81-4af0-830c-1fdec6b7b4b5.mov

## Type of change

- Fix

# Acceptance

Sorting by a number column should work properly

### DIFF
--- a/packages/react-widgets/src/models/TableModel.js
+++ b/packages/react-widgets/src/models/TableModel.js
@@ -6,7 +6,7 @@ export function getTable(props) {
 }
 
 function fromLocal(props) {
-  const { source, rowsPerPage, page, sortBy, sortDirection } = props;
+  const { source, rowsPerPage, page, sortBy, sortDirection, sortByColumnType } = props;
 
   return executeTask(source.id, Methods.FEATURES_RAW, {
     filters: source.filters,
@@ -14,6 +14,7 @@ function fromLocal(props) {
     limit: rowsPerPage,
     page,
     sortBy,
-    sortByDirection: sortDirection
+    sortByDirection: sortDirection,
+    sortByColumnType
   });
 }
--- a/packages/react-widgets/src/models/TableModel.js
+++ b/packages/react-widgets/src/models/TableModel.js
@@ -6,6 +6,7 @@ export function getTable(props) {
 }
 
 function fromLocal(props) {
+  // Injecting sortByColumnType externally from metadata gives better results. It allows to avoid deriving type from row value itself (with potential null values)
   const { source, rowsPerPage, page, sortBy, sortDirection, sortByColumnType } = props;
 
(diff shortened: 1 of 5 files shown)


bbecquet (approved):
LGTM

packages/react-workers/src/utils/sorting.js
```diff
@@ -21,20 +25,21 @@ export function applySorting(features, { sortBy, sortByDirection = 'asc' } = {})
   if (!isValidSortBy) {
     throw new Error('Sorting options are bad formatted');
   }
-
+  const isNumberColumn = sortByColumnType === 'integer' || sortByColumnType === 'float';
```
VictorVelarde, line 28: I guess this should be better a simple 'number'

packages/react-workers/src/utils/sorting.js
```diff
@@ -6,8 +6,12 @@ import { firstBy } from 'thenby';
  * @param {object} [sortOptions]
  * @param {string | string[] | object[]} [sortOptions.sortBy] - One or more columns to sort by
  * @param {string} [sortOptions.sortByDirection] - Direction by the columns will be sorted
+ * @param {string} [sortOptions.sortByColumnType] - Column type
```
VictorVelarde, line 9: I think we should enum here the specific allowed types, not just 'string' generic type, to make evident to clients what values are ok.

VictorVelarde, line 9:
```suggestion
 * @param {('number'|'string'|'date')} [sortOptions.sortByColumnType] - Column type
```

packages/react-widgets/src/models/TableModel.js
```diff
@@ -6,14 +6,15 @@ export function getTable(props) {
 }
 
 function fromLocal(props) {
-  const { source, rowsPerPage, page, sortBy, sortDirection } = props;
+  const { source, rowsPerPage, page, sortBy, sortDirection, sortByColumnType } = props;
```
VictorVelarde, line 9:
```suggestion
  // Injecting sortByColumnType externally from metadata gives better results. It allows to avoid deriving type from row value itself (with potential null values)
  const { source, rowsPerPage, page, sortBy, sortDirection, sortByColumnType } = props;
```

VictorVelarde (approved):
👍🏻

PR #597, opened 2023-02-08
Do not create fresh loaders array on every frame

felixpalmer:
Same fix as in Builder: https://github.com/CartoDB/cloud-native/pull/10952


VictorVelarde (approved):
LGTM, thx!

PR #599, opened 2023-02-09
react-workers: getFeaturesRaw returns full features with geometry if asked

zbigg:
# Description

Shortcut: https://app.shortcut.com/cartoteam/epic/274603

* react-workers - `getFeaturesRaw` returns full features with geometry if asked

* Store geometry in calculations of visible features for later use when exporting data.

# Type of change

- Feature

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-workers/src/workers/methods.js
+++ b/packages/react-workers/src/workers/methods.js
@@ -7,12 +7,15 @@ import {
   scatterPlot,
   groupValuesByColumn,
   groupValuesByDateColumn,
-  AggregationTypes
+  AggregationTypes,
+  ResultFormat,
+  InvalidColumnError
 } from '@carto/react-core';
-import { InvalidColumnError } from '@carto/react-core/';
+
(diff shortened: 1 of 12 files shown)


packages/react-workers/src/workers/methods.js
```diff
@@ -31,8 +34,10 @@ export function getTileFeatures({
     uniqueIdProperty,
     tileFormat,
     geoColumName,
-    spatialIndex
+    spatialIndex,
+    resultFormat: ResultFormat.GeoJsonFeature
```
zbigg, line 38: This means that we keep full geojson features always in memory.

I've chosen this as simplest approach that doesn't _modify_ workers "architecture" too much.

We could ask for full features "on-demand", when export is executed though ... this would require only keeping last view filters in state (which may be not necessarily worst thing).

PR #608, opened 2023-03-01
New component: LabelWithIndicator

vmilan:
# Description

Shortcut: [#296467](https://app.shortcut.com/cartoteam/story/296467/core-new-component-labelwithmarker)
[sc-296467]

Designers have the [need to mark inputs as optional or required](https://app.shortcut.com/cartoteam/story/287761/design-visual-distinction-among-required-and-optional-inputs-in-parameters-and-units) depending on the use case. That is, in some forms, we will have the optional ones marked but in others the required ones.

Screenshot 2023-03-01 at 18 59 51

Mui only provides a `required` bool prop to mark an input as mandatory. When is `true`, add an asterisk at the end of the label and the validation logic. 

As this is something quite urgent for `Workflows`, to avoid the need to modify all current required cases, in this PR I temporarily replaced via styles the asterisk with the tag `(required)`, so we would only need the new component to handle the optional mark. 

In the short-term, the use is like:

Required (as it is right now)
```
<TextField
  {...rest}
  label={label}
  required
  variant='filled'
  placeholder={placeholder}
/>
```

Optional field
```
<TextField
  {...rest}
  label={<LabelWithIndicator label={label} />}
  variant='filled'
  placeholder={placeholder}
/>
```

After this is merged, we can remove the asterisk, and handle both, required and optional tags, with the new component. So In the mid-term we will have the following:

Required
```
<TextField
  {...rest}
  label={<LabelWithIndicator label={label}  type='required' />}
  required
  variant='filled'
  placeholder={placeholder}
/>
```

Optional field
```
<TextField
  {...rest}
  label={<LabelWithIndicator label={label} />}
  variant='filled'
  placeholder={placeholder}
/>
```


VictorVelarde (approved):
👍🏻

PR #612, opened 2023-03-06
 Refactor from makeStyles to styled (II) - Several components

VictorVelarde:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/297056

- [x] LabelWithIndicator
- [x] SelectField
- [x] TooltipData
- [x] AppBar (migration to styled + internal refactor)
- [x] UploadField

## Type of change

- Refactor

# Acceptance

CR + storybook change

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation


vmilan:
CR & Tested in Storybook, everything seems fine!

VictorVelarde:
Ready for review @vmilan. Please have a look at storybook for visual validation of changes, thx!

packages/react-ui/src/index.d.ts
```diff
@@ -29,8 +29,8 @@ import Typography from './components/atoms/Typography';
 import Button from './components/atoms/Button';
 import PasswordField from './components/atoms/PasswordField';
 import SelectField from './components/atoms/SelectField';
-import UploadField from './components/molecules/UploadField';
-import AppBar from './components/organisms/AppBar';
+import UploadField from './components/molecules/UploadField/UploadField';
+import AppBar from './components/organisms/AppBar/AppBar';
```
vmilan, line 33: What about what we were discussing, to have an `index.js` instead of `UploadField/UploadField.js `?

zbigg: Please don't use `index.js` or any other convention that creates dozens of files with same name.
It is really hard to navigate in project, where half of files are named `index.js` or `main.yml`.
UploadField.js is much better imo.

(let alone fact, that single file component doesn't isn't enough reason to create _folder_ for it ...), why not having just`molecules/UploadField.js`  ?

`index.js` as concept also isn't ESM friendly, as ESM which is more or less future of JS completly ditched concept of "automatically" resolved "index" files that is CommonJS remnant.

packages/react-ui/src/components/organisms/AppBar/AppBar.js
```diff
@@ -0,0 +1,66 @@
...
+  showBurgerMenu,
+  onClickMenu,
+  ...otherProps
+}) => {
+  return (
+    <MuiAppBar {...otherProps}>
+      <Toolbar>
+        <BrandElements>
+          {showBurgerMenu && <BurguerMenu onClickMenu={onClickMenu} />}
+          {brandLogo && <BrandLogo logo={brandLogo} />}
```
zbigg, line 43: I think that separating _trivial_ components like `BrandLogo`,  `BrandText` abd `SecondaryText` into sparate files/components is just too much.

They could be just `<BrandLogo>{logo}</BrandLogo>` ... where `BrandLogo` is just direct reference to local, small `styled` component without intermediate "functional" component.

Maybe it is ok in this special case, but it's not necessarily good blueprint for whole project to follow.

vmilan:
@VictorVelarde I pushed a change to fix some hierarchy problems with styles in UploadField... apart from that, the visual review was ok

packages/react-ui/src/index.d.ts
VictorVelarde: Happy to talk about this with more calm in a coffee, considering pros / cons. 

I wouldn't discard index.js / .ts as a good way to structure external visibility / exports of components / functions (while keeping internally, not exported, auxiliary files for the main component), keeping the main component in its own file. I think that's a discussion worth to have.

@zbigg if you have references / articules related to ESM about this topic, I will happily review them

VictorVelarde: deck.gl is doing some work related to ESM & keeping index.js, eg here: https://github.com/visgl/deck.gl/blob/master/modules/arcgis/src/index.js

packages/react-ui/src/components/organisms/AppBar/AppBar.js
VictorVelarde: @zbigg I think I want to preserve this, and have a healthy discussion about new explicit rule: one file per component element

vmilan (approved):
👍

packages/react-ui/src/components/molecules/UploadField/StyledUploadField.js
```diff
@@ -3,7 +3,7 @@ import { styled, TextField } from '@mui/material';
 const StyledUploadField = styled(TextField, {
   shouldForwardProp: (prop) => prop !== 'focused'
 })(({ focused, theme }) => ({
-  '& .MuiInputBase-root': {
+  '&.MuiTextField-root .MuiInputBase-root': {
```
VictorVelarde, line 6: Thx @vmilan

PR #619, opened 2023-03-27
Display Carto icon gallery in Storybook

vmilan:
- Story: https://app.shortcut.com/cartoteam/story/302497/tooling-display-carto-icon-gallery-in-storybook
- Autolink: [sc-302497]

Useful to see the available icons (avoid duplicities when you add a new one) and reveal `SvgIcon` component options.

Screenshot 2023-03-29 at 10 56 09


packages/react-ui/src/index.d.ts
```diff
@@ -34,6 +34,7 @@ import AppBar from './components/organisms/AppBar/AppBar';
 import LabelWithIndicator from './components/atoms/LabelWithIndicator';
 import { getCartoColorStylePropsForItem } from './utils/palette';
 import Avatar from './components/molecules/Avatar';
+import { ICON_SIZE_S, ICON_SIZE, ICON_SIZE_M } from './theme/themeconstants';
```
zbigg, line 37: I have minor doubt about naming of those constants.

Looks like`ICON_SIZE_S` means small, and `ICON_SIZE_M` is medium ... then `ICON_SIZE` is normal? So, normal is smaller than medium ?

Maybe we could find more explicit and understantable names for these constants?

zbigg (approved):
LGTM, only one suggestion

packages/react-ui/src/index.d.ts
vmilan: Done! I've changed it to be more explicit

vmilan:
@zbigg I did also the migration to styled in the new story file... 

My idea indeed is to have some shared styled components for Storybook because, in most of the files, I ended up using the same set of them. I'll progressively migrate all makeStyles in stories in the same way.

Thanks for your review!

PR #641, opened 2023-04-12
Migrate 'LegendWrapper' component from makeStyles to styled-component + cleanup

jantolg:
# Description

This PR includes a refactorization of LegendWrapper component, all classes from makestyles are been converted to styled components

Shortcut:  [297037](https://app.shortcut.com/cartoteam/story/297037/migrate-away-from-makestyles-i-widgets)

## Type of change

- Refactor

# Acceptance

The code resultant must return same styles component visualization as the same before changes

1. Overview a general visualization from v1.x to v2.x

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-ui/src/widgets/legend/LegendWrapper.js
+++ b/packages/react-ui/src/widgets/legend/LegendWrapper.js
@@ -1,29 +1,30 @@
-import React, { createRef, useState } from 'react';
-import { Box, Button, Collapse, Grid, Icon, Switch, Tooltip } from '@mui/material';
-import makeStyles from '@mui/styles/makeStyles';
 import { ExpandLess, ExpandMore } from '@mui/icons-material';
-import Note from './Note';
+import {
+  Box,
+  Button,
+  Collapse,
+  Grid,
+  Icon,
+  Switch,
+  ToggleButton,
+  Tooltip,
+  styled
+} from '@mui/material';
+import React, { createRef, useState } from 'react';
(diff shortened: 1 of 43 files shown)


packages/react-ui/src/widgets/legend/LegendWrapper.js
```diff
@@ -82,61 +82,65 @@ export default function LegendWrapper({
       />
       {hasChildren && !!children && (
         <Collapse ref={wrapper} in={expanded} timeout='auto' unmountOnExit>
-          <Box className={classes.content}>
+          <Box sx={{ pt: 0, px: 2, pb: 3 }}>
```
VictorVelarde, line 85: Even if the sx values seem equal to original, the UI renders differently.

I was still getting partially 'hidden' copy, until I changed a bit the values.

Left: previous / Center: this branch / Right: the modified values 
image

The problem to me is that the overall structure is here quite messy. After moving away from makeStyles I think we should make a deep refactor and cleanup of the component

VictorVelarde: @vmilan I'm gonna move on with this approach, to land PR and unblock the rest, but we should check it in the mid-term

VictorVelarde (approved):
👍🏻

PR #655, opened 2023-04-20
Build more accurate ranges for time filters

Shylpx:
# Description

- [Shortcut story](https://app.shortcut.com/cartoteam/story/308300/time-series-widget-precision-issues)

One option to filter using a time series widget is to select a moment in time (instead of using a range). When this happens, we build a range, using that moment as the start date.

For example, if we are aggregating by day, and the user selects 2020-08-20, we create a range to select all timestamps between 2020-08-20 00:00:00 and 2020-08-21 00:00:00. But this range has accuracy issues:

- This is a close range, so if we select 07:04 PM UTC we are not filtering by 07:04 PM UTC, but also by 07:05 PM UTC.
- The list we use to build the end of the range, is not accurate. It works for weeks/days/hours/minutes, but for year/months, they aren't a fixed number of milliseconds. We are wrongly assuming a month has 30 days, and a year has always 365 days:

```
const STEP_SIZE_RANGE_MAPPING = {
  [GroupDateTypes.YEARS]: 60 * 60 * 24 * 365 * 1000,
  [GroupDateTypes.MONTHS]: 60 * 60 * 24 * 30 * 1000,
  ...
};
```

## Type of change

- Fix

# Acceptance

Left size: **new changes**
Right size: **old approach**

1. We filter by `9/28/2922 08:09 PM`, and there are 154 features with this value.
  - Left side shows the correct count in the formula widget
  - Right side is also incluiding 9/28/2022 08:**10** PM features
 
Screenshot from 2023-04-20 10-25-53

2. We aggregate by month, and we have features in 31 august, and we filter by august.
  - Left side shows the correct count
  - Right side, since it assumes all months has 30 days, doesn't show any feature

Screenshot from 2023-04-20 11-26-23


zbigg (commented):
LGTM, but leaving final approval for @VictorVelarde

VictorVelarde (approved):
👍🏻 To be included in the next release, asap

PR #658, opened 2023-04-25
Calculation of widget using maps API

ragazzojp:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/304698

This introduces the possibility for widget to be calculated remotely, so using the maps API. It is the case today for global mode only. This PR adds supports for non-global widgets to offload calculation to the backed. This is also limited to non-tileset data source and data source using V3-API only.

When enabled, widgets are more reactive and don't need to wait for the complete dataset to be downloaded. In demo apps, widgets gets rendered before data is even shown on the map.

The point of control of the feature is the function `isRemoteCalculationSupported`. Note that each widget must have this possibility turned on as well, via the initialisation parameter `attemptRemoteCalculation`, currently hardcoded to `true` for widgets that supports the feature.

No change in public API expected. Carto 4 React users won't be able to control this feature.


packages/react-api/src/api/model.js
```diff
@@ -52,13 +54,21 @@ export function executeModel(props) {
...
 
+  if (spatialFilter) {
+    queryParams.spatialFilter = JSON.stringify(spatialFilter);
+  }
+
   const isGet = url.length + JSON.stringify(queryParams).length <= URL_LENGTH;
   if (isGet) {
     url += '?' + new URLSearchParams(queryParams).toString();
   } else {
+    // undo the JSON.stringify, @todo find a better pattern
```
zbigg, line 65: You can't get much better because you have to measure size of `GET` url few lines above and that must have proper url as input.

(btw, it should create proper URL first, measure and test and not rely of more less accourate assumptions that `JSON.stringify(queryParams).` is same length as `new URLSearchParams(queryParams).toString()` but that's other story)

This is done correctly in `getStats` albeit it's not much more readable.

zbigg (approved):
LGTM

packages/react-api/src/api/model.js
ragazzojp: I'll try to introduce an utility function for this, in a later PR.

ragazzojp:
Im my opinion, we're ready to merge this PR.

VictorVelarde (approved):
👍🏻 really nice job here!

PR #661, opened 2023-04-27
Widgets work on either mask or viewport

ragazzojp:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/303989/widgets-wrongly-operate-on-intersection-between-viewport-and-mask

This PR removes the intersection between viewport and (optional) mask, when filtering features and calculating widgets. 

New behavior:
- If a mask is set, filter by that mask only
- Otherwise filter by viewport

Old behavior:
- If a mask is set, filter by the intersection of that mask and the viewport
- Otherwise filter by viewport

Also includes refactoring of `getGeometryToIntersect`:
- moved closer to the root of render cycle, it avoids unneeded widget recalculation
- better typing via JSDoc
- object returned are GeoJSON `Polygon` or `MultiPolygon` and not `Feature` anymore

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
+- Changed the way widget are calculated when a mask is set: use just the mask, no more intersection between mask and viewport [#661](https://github.com/CartoDB/carto-react/pull/661)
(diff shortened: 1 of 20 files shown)


VictorVelarde (approved):
Overall LGTM. Thx also for the cleanup 👍🏻 . I'm only missing a short description added in the CHANGELOG, and I'll ping PS in slack for awareness

PR #668, opened 2023-05-08
Fix paired buttons spacing when the button is from a different variant

vmilan:
story: https://app.shortcut.com/cartoteam/story/292993/workspace-settings-ui-improvements-bugfixing
[sc-292993]

Found a bug in the theme related to buttons: when you have 2 buttons inline, there is no horizontal padding if they are from different variants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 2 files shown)


zbigg (approved):
LGTM

PR #671, opened 2023-05-10
Fix histogram widget filter for max/min values

moicalcob:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/312083/histogram-widget-with-custom-max-value-displaying-values-outside-of-range#activity-313269

The Histogram widget max and min filters were not working correctly. We were filtering using an open interval, so these min/max values were not really being used.

With this change, we should be able to filter correctly using a custom min/max value.

## Type of change

- [x] Fix

# Acceptance

1. Add a histogram widget with custom min/max values
2. Check that the max/min values' filter is correctly applied
- Max filter:
May-10-2023 13-45-04

- Min filter:
May-10-2023 13-53-07

3. Check that if there is one feature that has the same value as the max filter, the feature is not filtered
May-10-2023 13-46-42

4. Perform the same check with the min value
May-10-2023 13-53-57

# Basic checklist

- [x] Good PR name
- [x] Shortcut link
- [x] Changelog entry
- [x] Just one issue per PR
- [x] GitHub labels
- [ ] Tests
- [ ] Documentation


VictorVelarde (approved):
👍🏻

PR #674, opened 2023-05-18
Bar & Histogram & Formula  & ComparativeFormula Widgets: Add a skeleton for loading state

vmilan:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/314283/new-loader-ui-for-widgets
[sc-314283]


packages/react-ui/src/theme/sections/components/feedback.js
```diff
@@ -7,5 +10,30 @@ export const feedbackOverrides = {
         horizontal: 'center'
       }
     }
+  },
+
+  // Skeleton
+  MuiSkeleton: {
+    defaultProps: {
```
VictorVelarde, line 17: Ok, good to know we have some basics in common

packages/react-widgets/src/widgets/BarWidget.js
```diff
@@ -156,6 +156,7 @@ function BarWidget({
             height={height}
             animation={animation}
             filterable={filterable}
+            isLoading={isLoading}
```
VictorVelarde, line 159: Why is this injected externally? Shouldn't this be just a prop that the 'widget' launching the query, passes down to the UI?

vmilan: This is what I'm doing here, isn't it? passing down the prop from the parent to the UI, as in line 137 for the WrapperWidgetUI.

The logic was already there, but the UI wasn't receiving it.

vmilan:
@VictorVelarde added also the HistogramWidget

packages/react-widgets/src/widgets/BarWidget.js
VictorVelarde: Yes, misunderstanding on my side. It's fine

PR #680, opened 2023-05-22
Normalization functions for mask/viewports

ragazzojp:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/314403/formula-and-histogram-widgets-viewport-mode-is-not-displaying-data-with-zoom-0-bq-check-this-one

This normalizes the spatial filter before sending it to the API for remote calculation

### DIFF
--- a/packages/react-core/src/utils/geo.js
+++ b/packages/react-core/src/utils/geo.js
@@ -1,4 +1,8 @@
+import bboxClip from '@turf/bbox-clip';
 import bboxPolygon from '@turf/bbox-polygon';
+import union from '@turf/union';
+import { getType } from '@turf/invariant';
+import { polygon, multiPolygon } from '@turf/helpers';
 
 /**
  * Select the geometry to use for widget calculation and data filtering.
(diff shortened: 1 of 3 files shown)


packages/react-core/src/utils/geo.js
```diff
@@ -1,12 +1,16 @@
+import bboxClip from '@turf/bbox-clip';
 import bboxPolygon from '@turf/bbox-polygon';
+import union from '@turf/union';
+import { getType } from '@turf/invariant';
+import { polygon, multiPolygon } from '@turf/helpers';
```
VictorVelarde, line 5: Ok, it sounds like we should add those as peerDependencies in react-core, right?

ragazzojp: Don't we have them already?

VictorVelarde (commented):
Overall LGTM, just drop a line into CHANGELOG and manage new deps, and we'll go ahead with it @ragazzojp

CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 
 ## Not released
 
+- Fix widget calculation with very large viewports/masks [#680](https://github.com/CartoDB/carto-react/pull/680)
+
 ## 2.0
 
-## 2.0.4 (2023-05-19)
+### 2.0.4 (2023-05-19)
```
VictorVelarde, line 9: Thx for the fix

VictorVelarde (approved):
👍🏻

PR #683, opened 2023-05-24
Support remote table widget

ragazzojp:
https://app.shortcut.com/cartoteam/story/314307/support-for-table-widget


packages/react-widgets/src/models/TableModel.js
```diff
@@ -19,3 +24,32 @@ function fromLocal(props) {
     sortByColumnType
   });
 }
+
+// @todo we need to cache res somehow otherwise we'll hit the API
```
VictorVelarde, line 28: This could be probably well managed adding react-query as a new big but useful dep of the package to cache in the frontend. We can have that discussion in other moment

VictorVelarde: https://tanstack.com/query/v4/docs/react/examples/react/pagination

VictorVelarde (dismissed):
👍🏻 Go ahead whenever you want, just please add an entry in changelog

packages/react-widgets/src/models/TableModel.js
```diff
@@ -1,8 +1,12 @@
+import { _executeModel } from '@carto/react-api';
 import { Methods, executeTask } from '@carto/react-workers';
-import { wrapModelCall } from './utils';
+import { normalizeObjectKeys, wrapModelCall } from './utils';
+
+// Make sure this is sync with the same constant in cloud-native/maps-api
+const HARD_LIMIT = 100;
```
VictorVelarde, line 6: Ana PM is aware of this limit, right?

ragazzojp: Yes, she's fine with it.

VictorVelarde:
Checked last changes and LGTM 👍🏻

PR #685, opened 2023-05-24
Avoid reset of table widget to page 0 when not necessary

ragazzojp:
This reduces the number of spurious reset of a table widget to page 0.
Page is reset to 0 only when the total number of pages change.
Before, page was reset to 0 at every change of mask, viewport (scrolling) or setting of other widgets.

I think this improves the UX but the true need behind this change is that with remote table widget support the `isSourceReady` variable is always `true` and it's impossible to reset the page to 0 when, for example, the viewport change and the number of pages decreases, leaving the widget in an inconsistent, _no data_ state.

It could even be improved by resetting to page 0 only in case the total number of pages decreases below the current page, but I was not able to implement it. 😅


packages/react-widgets/src/widgets/TableWidget.js
```diff
@@ -34,7 +34,6 @@ function TableWidget({
   onError,
   initialPageSize = 10,
   onPageSizeChange,
-  global,
```
VictorVelarde, line 37: I guess removing global option is a product decision, already discussed, right?. From Builder it's not used, so ok from that side, but not sure if we might have other uses cases outside... @aaranadev ?

ragazzojp: Global mode is not supported for table widgets, according to both documentation and actual way it's integrated in Builder, where there's no global radio button.

https://docs.carto.com/carto-for-developers/carto-for-react/library-reference/widgets#tablewidget

ragazzojp: If instead we want to have it, it should be documented, and then I'll revert and add documentation.

ragazzojp: I'll revert this to be on the safe side.

VictorVelarde (approved):
👍🏻 thx!

PR #721, opened 2023-06-23
Fix case sensitive for cell values

Josmorsot:
# Description

Shortcut: [(link)](https://app.shortcut.com/cartoteam/story/324740/c4r-sql-api-for-widgets-clientid)

## Type of change
- Feature

### DIFF
--- a/packages/react-ui/__tests__/widgets/TableWidgetUI.test.js
+++ b/packages/react-ui/__tests__/widgets/TableWidgetUI.test.js
@@ -83,7 +83,7 @@ describe('TableWidgetUI', () => {
       render(<Widget columns={filteredColumns} />);
 
       const row = rows[1];
-      screen.getByText(row.address);
+      expect(screen.queryByText(row.address)).toBeInTheDocument();
     });
   });
 });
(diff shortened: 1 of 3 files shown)


packages/react-ui/__tests__/widgets/TableWidgetUI.test.js
```diff
@@ -73,4 +73,17 @@ describe('TableWidgetUI', () => {
...
+  describe('case insensitive', () => {
+    test('should render rows properly', () => {
+      const keptFields = ['address', 'city'];
+      const filteredColumns = columns
+        .filter((col) => keptFields.includes(col.field))
+        .map((c) => ({ ...c, field: c.field.toUpperCase() }));
+      render(<Widget columns={filteredColumns} />);
+
+      const row = rows[1];
+      screen.getByText(row.address);
```
bbecquet, line 86: This test is lacking an `expect`. I know it will fail if the `getByText` doesn't work, but it's better to be explicit about what we want and to have proper error messages.

Josmorsot: ups, you're right. Copy-paste

Josmorsot: fixed

bbecquet (approved):

PR #725, opened 2023-06-29
Add TableCell styles for padding prop

vmilan:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/325900/unify-table-ui-in-settings
[sc-325900]

Add TableCell styles for padding prop to be able to adapt the padding to different use cases based on DS specs.


VictorVelarde (approved):
👍🏻

PR #727, opened 2023-06-29
Localization in C4R

menusal:
# Description

image

Basic setup PoC (alpha): initialization + 1 widget (CategoryWidget)

## Type of change

- Feature

## TODOs

[-] Implementation
[ ] Test
[ ] Docuentation
[ ] Storybook

# Acceptance

1. dispatch a new locale with `dispatch(setLocale('es-ES'))` in your C4R app.
2. Check if the messages are in dispatched locale.

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation


packages/react-ui/src/localization/en.js
```diff
@@ -0,0 +1,7 @@
+import c4r from './en/c4r.js';
+
+const locales = {
+  c4r
```
VictorVelarde, line 4: Probably at this point we can just put the c4r.js content here. And if we eventually need to split this into several files, we could have one file per package (so one file per `widgets`, other for `ui`, other fore `basemaps`, etc)

packages/react-ui/src/localization/index.js
```diff
@@ -0,0 +1,11 @@
+import en from './en';
+
+import { flattenMessages } from './localeUtils';
+
+const enFlat = flattenMessages(en);
+
+export const messages = {
```
VictorVelarde, line 7: As we might eventually add localization in other parts (as error messages or similar), I feel this could probably go into core, what do you think?

VictorVelarde: cc/ @zbigg

packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
```diff
@@ -45,7 +46,8 @@ function CategoryWidgetUI(props) {
     animation,
     filterable,
     searchable,
-    isLoading
+    isLoading,
+    localizationMessages
```
VictorVelarde, line 50: Why the need to pass localizationMessages as prop? I think they easiest approach here would be to read it, in the widgets, from a simple

```
import { useIntl } from 'react-intl';

function MyComponent() {
  const intl = useIntl();

  return (
    <div>
      <p>{intl.formatMessage({ id: 'greeting' })}</p>
      {/* Other components */}
    </div>
  );
}
```

packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
```diff
@@ -476,7 +502,9 @@ CategoryWidgetUI.propTypes = {
   animation: PropTypes.bool,
   filterable: PropTypes.bool,
   searchable: PropTypes.bool,
-  isLoading: PropTypes.bool
+  isLoading: PropTypes.bool,
+  localizationMessages: PropTypes.object,
```
VictorVelarde, line 506: I wouldn't recommend this approach... Is this a must for you guys?

packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
VictorVelarde: That leaves us with probably the main point here, which is lib initialization with the correct locale... Probably there are several options...

VictorVelarde: I'm not totally sure, but maybe there is an option if we set it up upper, in the widgets (with access to the redux). The widgets could wrap up the UI, with a provider around them, and the provider could ideally get the locale from redux (just get / set locale in redux, not messages). This is pure theory, haven't tested id

packages/react-ui/src/hooks/useLocale.js
```diff
@@ -0,0 +1,25 @@
...
+import { messages } from '../localization';
+import { useMemo } from 'react';
+
+const DEFAULT_LANGUAGE = 'en';
+/**
+ * React hook to handle custom locale messages to override default ones
+ * @param {object} localizationMessages
+ */
+export default function useLocale(localizationMessages) {
+  const cache = createIntlCache();
```
VictorVelarde, line 11: I'm not sure if the cache-thing should leave here, in the hook. Ideally it should live upper, right?

menusal: right 👍

packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
menusal: It's a first approach. We can discuss

packages/react-ui/src/localization/en.js
menusal: Ok 👍 , that has all the sense to me

packages/react-ui/src/hooks/useLocale.js
menusal: Done ✅

zbigg:
Yes, we do have code in our products like this:

```
 <WrapperWidgetUI {...props}>
     ...
        <FormattedMessage id="app.specific.message" />
      ...
    </WrapperWidgetUI>
```

so, if `WrapperWidgetUI` would need "c4r" intl provider, then it should _reset_ intl to "app" when rendering children ... or somehow merge intl instances.

packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
```diff
@@ -57,10 +58,21 @@ function TableWidgetUI({
...
-  lastPageTooltip
+  lastPageTooltip,
+  locale
 }) {
   const paginationRef = useRef(null);
 
+  const intl = useImperativeIntl(locale);
+
+  const defaultLabelDisplayedRows = ({ from, to, count }) => {
+    return `${from}–${to} ${intl.formatMessage({ id: 'c4r.widgets.table.of' })} ${
```
zbigg, line 69: please create _one_ localized string which is `{from}-{to} of {count} more than {to}` and pass values using `values` mechanism of formatjs/react-intl (used in other parts of PR)

packages/react-ui/src/widgets/FeatureSelectionWidgetUI/FeatureSelectionWidgetUI.js
```diff
@@ -51,25 +53,68 @@ function FeatureSelectionWidgetUI({
...
+  locale
 }) {
+  const intl = useImperativeIntl(locale);
+
+  const selectionModeWithLocales = useMemo(() => {
+    return selectionModes.map((mode) => ({
+      ...mode,
+      label: intl.formatMessage({
+        id: `c4r.widgets.featureSelection.${
+          mode.id.charAt(0).toLowerCase() + mode.id.slice(1)
```
zbigg, line 66: please create util function for this logic

packages/react-ui/src/widgets/legend/LegendWrapper.js
```diff
@@ -79,14 +83,15 @@ export default function LegendWrapper({
...
+        locale={locale}
       />
       {hasChildren && !!children && (
         <Collapse ref={wrapper} in={expanded} timeout='auto' unmountOnExit>
           <Box sx={{ pt: '2px', pl: 3, pr: 2, pb: 2 }}>
             <Grid container direction='column' spacing={1}>
               {attr && (
                 <Typography xs mb={1} variant='caption'>
-                  By {attr}
+                  {intl.formatMessage({ id: 'c4r.widgets.legend.by' })} {attr}
```
zbigg, line 94: you don't know grammar of other languages, create message `by {attr}` and pass `attr` in `values`

packages/react-redux/src/slices/cartoSlice.d.ts
```diff
@@ -176,3 +177,8 @@ export function selectSpatialFilter(
 ): Feature<Polygon | MultiPolygon> | null;
 
 export function selectFeatureSelectionMode(state: any): string | null;
+
+export function setLocale(localization: any): {
+  type: CartoActions.SET_LOCALE;
+  payload: any;
```
zbigg, line 183: any -> string

zbigg (changes requested):
i still don't see a wa for user to pass his own localization strings for other languages "polish" isn't this in scope scope?

> - yarn.lock
why yarn.lock was deleted?

packages/react-ui/src/hooks/useImperativeIntl.js
```diff
@@ -0,0 +1,24 @@
+import { createIntl, createIntlCache } from 'react-intl';
+import { messages } from '../localization';
+
+// This is optional but highly recommended
+// since it prevents memory leak
```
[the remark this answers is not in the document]
menusal: Ok, it's from official intl website, removing comment.

packages/react-ui/src/hooks/useImperativeIntl.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * The hook returns an instance of the Intl object with the specified locale
+ * and messages using imperative API.
+ * @see https://formatjs.io/docs/react-intl/api/#createintl
+ * @param {string} locale - The locale to use.
+ * @returns the `intl` object.
+ */
+export default function useImperativeIntl({ locale }) {
+  const intl = createIntl(
```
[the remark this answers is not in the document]
menusal: Done

packages/react-redux/src/slices/cartoSlice.d.ts
menusal: Done

packages/react-ui/src/widgets/FeatureSelectionWidgetUI/FeatureSelectionWidgetUI.js
menusal: Done

packages/react-ui/src/widgets/legend/LegendWrapper.js
menusal: Done

packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
menusal: Done

menusal:
> i still don't see a wa for user to pass his own localization strings for other languages "polish" isn't this in scope scope?
> 
> > * yarn.lock
> >   why yarn.lock was deleted?

About `yarn.lock`, I'll restore
About `the user passing his own localization strings, it's a feature for another iteration and depends on future demands from PS customers cc: @VictorVelarde

packages/react-ui/src/hooks/useImperativeIntl.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * The hook returns an instance of the Intl object with the specified locale
+ * and messages using imperative API.
+ * @see https://formatjs.io/docs/react-intl/api/#createintl
+ * @param {string} locale - The locale to use.
+ * @returns the `intl` object.
+ */
+export default function useImperativeIntl(locale) {
+  // NOTE: We need to remove the dash from the locale to match the messages file keys, ex: 'es-ES' -> 'esES'
```
VictorVelarde, line 16: Why not just renaming the localization files?

packages/react-ui/src/localization/idID.js
```diff
@@ -0,0 +1,77 @@
...
+        minute: 'Menit'
+      },
+      featureSelection: {
+        drawPolygonMode: 'Poligon',
+        drawRectangleMode: 'Persegi panjang',
+        drawPolygonByDraggingMode: 'Alat lasso',
+        drawCircleFromCenterMode: 'Lingkaran',
+        edit: 'Edit',
+        clickToEdit: 'Klik pada masker untuk mengeditnya',
+        clickToCreate: 'Klik pada peta untuk membuat sebuah masker',
```
VictorVelarde, line 66: Amazing 😛

packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
```diff
@@ -195,7 +205,8 @@ TableWidgetUI.defaultProps = {
   pagination: false,
   rowsPerPage: 10,
   rowsPerPageOptions: [5, 10, 25],
-  dense: false
+  dense: false,
+  locale: 'en-ES'
```
VictorVelarde, line 209: You mean 'en-US' right?

packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -190,6 +191,9 @@ export const createCartoSlice = (initialState) => {
       },
       setFeatureSelectionEnabled: (state, action) => {
         state.featureSelectionEnabled = action.payload;
+      },
+      setLocale: (state, action) => {
+        state.locale = action.payload;
```
VictorVelarde, line 196: Maybe it's fine to have it in redux, but the problem is that even having it inside, we're not able to get it from any part of the lib (as the UI), and we must pass as prop to widgets. 

This is due the current split in several modules of the lib, maybe we should reconsider this in the future, and things like this would be much easier.

packages/react-ui/src/hooks/useImperativeIntl.js
menusal: Do you mean es-ES.js for example, with dash?

packages/react-ui/src/widgets/TableWidgetUI/TableWidgetUI.js
menusal: Right

packages/react-ui/src/hooks/useImperativeIntl.js
menusal: Done

packages/react-ui/src/widgets/BarWidgetUI/BarWidgetUI.js
```diff
@@ -44,13 +45,16 @@ function BarWidgetUI(props) {
     height,
     filterable,
     animation,
-    isLoading
+    isLoading,
+    locale
```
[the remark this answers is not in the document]
menusal: If you use the UI it could be useful to keep locale prop.
Perhaps I need more context from @Josmorsot approach to apply with locale, redux. 
¿Do you have an example using setClient?

packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
```diff
@@ -456,7 +472,8 @@ CategoryWidgetUI.defaultProps = {
   selectedCategories: [],
   animation: true,
   filterable: true,
-  searchable: true
+  searchable: true,
+  locale: 'en-US'
```
aaranadev, line 476: Maybe it's better to use `DEFAULT_LOCALE` from the `packages/react-ui/src/hooks/useImperativeIntl.js` file

packages/react-ui/src/hooks/useImperativeIntl.js
```diff
@@ -0,0 +1,33 @@
...
+ * The hook returns an instance of the Intl object with the specified locale
+ * and messages using imperative API.
+ * @see https://formatjs.io/docs/react-intl/api/#createintl
+ * @param {string} locale - The locale to use.
+ * @returns the `intl` object.
+ */
+export default function useImperativeIntl(locale) {
+  // Check if the locale is valid ISO 5 letters code (ex: 'en-US', 'es-ES', 'id-ID')
+  const testLocale = /^[a-z]{2}-[A-Z]{2}$/;
+  const validLocale = testLocale.test(locale) ? locale : DEFAULT_LOCALE;
```
aaranadev, line 18: Is it possible to check if the locale is the shortest and add the next part by default? If the locale is `en`, the full location is `en-EN` so we can build it.

menusal: In my opinion it is better to pass the complete ISO since it is required for the formatting of dates and numbers, and the same format is not applied to es-ES as to es-MX or en-EN to en-US for example

packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
menusal: yes, you right, I'll change it

menusal: Done

packages/react-ui/src/widgets/BarWidgetUI/BarWidgetUI.js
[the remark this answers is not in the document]
zbigg: If we want to have strict control over locale, yes, we need to pass it somehow using react mechanism.

`client` approach is just a global variable, that doesn't have responsivness.

Now, `dispatch(setLocale())` will cause proper rerender of affected components. Pure `setLocale` (global setter) will not, unless we implement complex, unnecessary notification mechanism ... aka like react context.

I already proposed to use `IntlContext` [here](https://github.com/CartoDB/carto-react/pull/727#pullrequestreview-1512604592) to get locale provided by `ReactIntlProvider` and ... fallback to default if it's missing. This API would be most robust and would be compatible with "react-intl" way of doing things.

Now, in this situation we can only get `locale` from externally provided object like this:

```
const useOurIntl = () => {
  const externalIntl = useContext(IntlContext)
  const locale = externalIntl?.locale || DEFAULT_LOCALE
  const messagesLocale = findMatchingMessagesLocale(locale)
  return createIntl({locale: externalIntl?.locale, messages: messages[messagesLocale]) // ofc memo and cache and stuff
}
```

now, if app doesn't have ReactIntlProvider, c4r would fallback to `en-EN`
else we would use what user has specified in app

my subjective evaluation
 - no redux,
 - no prop drilling
 - responsive
 - simple enough imo
 - ReactIntl-way

@VictorVelarde @menusal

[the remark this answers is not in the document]
zbigg: > `fingMatchingMessagesLocale`

there is not-yet-standard api for this:
 * https://github.com/tc39/proposal-intl-localematcher
 * poly/ponyfill: https://github.com/tc39/proposal-intl-localematcher

```
> const { match } = require('@formatjs/intl-localematcher')
> match(['en'], ['es-ES', 'es-MX', , 'en-US'], 'en')
'en-US'
> match(['es'], ['es-ES', 'es-MX', 'en-US'], 'en')
'es-ES'
> match(['es-MX'], ['es-ES', 'es-MX', 'en-US'], 'en')
'es-MX'
> match(['es-XX'], ['es-ES', 'es-MX', 'en-US'], 'en')
'es-ES'
```

packages/react-redux/src/slices/cartoSlice.js
VictorVelarde: As formatters were out of the scope initially, I didn't consider those here. In our current codebase, formatters can be passed down to widgets as props

packages/react-ui/src/widgets/BarWidgetUI/BarWidgetUI.js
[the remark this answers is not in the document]
VictorVelarde: To me one key point is avoid prop drilling.
I think zbigg's proposal is the simplest and keeps that in control, and aligned to the library model of work. This would be my vote.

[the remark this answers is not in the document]
menusal: @VictorVelarde @zbigg  
In this case:

```
now, if the app doesn't have ReactIntlProvider, c4r would fallback to en-US
else we would use what user has specified in app
```

If the app doesn't have ReactIntlProvider. Will you be unable to change the language?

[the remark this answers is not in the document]
VictorVelarde: Correct

menusal:
Closed, new implementation in `feature/menusal/locale-by-provider`

PR #729, opened 2023-07-03
Passing client through widgets without component attribute

Josmorsot:


VictorVelarde:
I would like to see this first working from an alpha, in a dedicated from cloud-native adapted well to it... Is that possible please @Josmorsot ?

Josmorsot:
> I would like to see this first working from an alpha, in a dedicated from cloud-native adapted well to it... Is that possible please @Josmorsot ?

hey @VictorVelarde, I've just published an alpha. **2.1.5-alpha-widget-client.0**. You can see it here: https://github.com/CartoDB/cloud-native/pull/12881

PR #742, opened 2023-07-14
Improve regex to detect coordinate pairs on search location widget

Shylpx:
# Description

[Shortcut story](https://app.shortcut.com/cartoteam/story/330496/support-pinea-geocode-by-lat-lon-in-builder-uses-lds-api-if-coordinate-pair-is-long)

The scope of this task would be to improve the regex we use to detect a coordinate pair on the search location widget:

- We are not supporting more than 9 decimal digits (for example, when we copy a coordinate from Google Maps, it is using ~15 decimal digits).
- We are allowing some invalid coordinates, like latitudes or longitudes out of range.

## Type of change

- Fix

# Acceptance

- Use a long coordinate pair on the search location widget:

search-location

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Not released
 
 - Fix to build more accurate ranges for time filters [#655](https://github.com/CartoDB/carto-react/pull/655)
+- Improve regex to detect coordinate pairs on search location widget [#742](https://github.com/CartoDB/carto-react/pull/742)
 
 ## 2.1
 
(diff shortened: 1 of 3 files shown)


Shylpx:
> About **precision**: we decided with PM what a 'reasonable precision for coordinate' would look like, that's why we kept it smaller. But if adding a bigger size helps that copy & paste with gmaps I see no objection.
> 
> Regarding **validation**, we decided to manage it outside of the coord. detection. I think that way the code is simpler.
> 
> What do you think @eamador ?

Thank you! Just a clarification regarding **validation**: I didn't change the current behavior, just adjusting a bit what we consider a "coordinate" for consistency. I mean, for example we were already saying that `100 0` is NOT a coordinate (because latitude is out of range), but the old regex was not taking into account other cases like `91.5 0`. Let me know what you thinkg @eamador :bow:

eamador (approved):

PR #744, opened 2023-07-19
Fix size of image placed inside a Chip 

vmilan:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/332632/sharing-with-groups-avatar-on-chips-are-not-rounded
[sc-332632]

The image placed inside a chip is not displayed properly if there is not enough space, so we're making sure always it has the needed width.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Not released
 
+- Fix size of image placed inside a Chip [#744](https://github.com/CartoDB/carto-react/pull/744)
+
 ## 2.1
 
 ## 2.1.8 (2023-07-07)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -54,6 +54,7 @@
 - Mask, when set, is applied to global widgets as well as to viewport-based widgets [#704](https://github.com/CartoDB/carto-react/pull/704)
 - Support for remote scatter plot widget [#704](https://github.com/CartoDB/carto-react/pull/704)
 - Breaking change for styles: sx / classname props removal [#715](https://github.com/CartoDB/carto-react/pull/715)
+- Fix size of image placed inside a Chip [#744](https://github.com/CartoDB/carto-react/pull/744)
```
VictorVelarde, line 57: Wrong place, move to "Not released"

VictorVelarde (approved):
👍🏻

PR #750, opened 2023-07-25
Add namespace as optional param

Josmorsot:
# Description

Shortcut: [(link?)](https://app.shortcut.com/cartoteam/story/)

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,5 @@
   ],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "2.2.2-cra-sso-support.2"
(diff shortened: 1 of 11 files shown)


mdeividas (approved):

lerna.json
```diff
@@ -4,5 +4,5 @@
   ],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "2.2.1"
+  "version": "2.2.2-cra-sso-support.3"
```
alasarr, line 7: I'm not following why are we binding to a specific version

Josmorsot: it's fixed in main

PR #767, opened 2023-09-06
TimeSeriesWidget support for multiple time series.

zbigg:
# Description

Shortcut: 
 * https://app.shortcut.com/cartoteam/story/336297/c4r-support-new-fields-model-for-timeseries-with-multiple-agg-splitbycategory-fidelity
 * https://app.shortcut.com/cartoteam/story/336298/c4r-implement-ui-for-timeseries-with-multiple-agg-splitbycategory

TimeSeriesWidget support for multiple time series. IN particular
 * support rendering of multiple series
 * support rendering for legend
 * support for new `splitByCategory` mode
 * support for filtering by category in `splitByCategory` mode
 * improved skeleton to better match sizing/placement of items with original (also in TableWidget)
 * new visuals in storybook
 

image

## Type of change

- Feature

# Acceptance

Please describe how to validate the feature or fix

1. go to X
2. test Y
3. assert Z

If feature deals with theme / UI or internal elements used also in CARTO 3, please also add a note on how to do acceptance on that part.

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

PR #777, opened 2023-09-22
Pie Widget:  New ChartLegend component

vmilan:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/347753/legend-must-be-interactive
[sc-347753]

When user select over a legend item or more, those slices are highlighted while the other slices are kept greyed as well as the rest of the legend item.

The same approach as new Multi TimeSeries

### DIFF
--- a/packages/react-ui/src/widgets/ChartLegend.js
+++ b/packages/react-ui/src/widgets/ChartLegend.js
@@ -24,7 +24,8 @@ const Item = styled(Box)(({ theme }) => ({
   gap: theme.spacing(1),
   alignItems: 'center',
   textOverflow: 'ellipsis',
-  whiteSpace: 'nowrap'
+  whiteSpace: 'nowrap',
+  cursor: 'pointer'
 }));
 
 const OverflowVeil = styled(Box)(({ theme }) => ({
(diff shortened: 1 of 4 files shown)


packages/react-ui/src/widgets/ChartLegend.js
```diff
@@ -0,0 +1,169 @@
+import React, { useState, useRef, useEffect, useCallback } from 'react';
```
vmilan, line 1: Mostly the same component copy / pasted, but adapted to PieWidget

https://github.com/CartoDB/carto-react/pull/767/files#diff-25c0014f515956ba1c9b2abf63cf9aea3ce7458ae32a195710015f038309ce18

zbigg (approved):

packages/react-ui/src/widgets/ChartLegend.js
mdeividas: would be possible to re-use those parts instead of copy&paste?

mdeividas (approved):

packages/react-ui/src/widgets/ChartLegend.js
vmilan: yes, @zbigg already did it in his branch :) we'll only have one file (ChartLegend) at the end of both projects

PR #779, opened 2023-09-28
PieWidgetUI: Refactor & New CentralText component

vmilan:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/346963/behaviour-of-central-text
[sc-346963] [sc-351761]

Echarts doesn't provide a proper way to extend the functionality of `Labels` or `Title`, so we need to create a new component to replace the text in the center of the chart to accomplish this product requirement: 

- The central text represents the most frequent category of the selection: whether they are all or just specific categories selected. Don't take into account deactivated categories.
- Don't update the central text and keep it as the most frequent category - Don't change the category on hover
- Add color indicator next to the category name

https://github.com/CartoDB/carto-react/assets/1934144/3cd56b22-faed-48af-877a-84c4ebeca83b

### DIFF
--- a/packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
+++ b/packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
@@ -98,45 +98,6 @@ function PieWidgetUI({
     [formatter, theme.palette.common.white, theme.palette.black, tooltipFormatter]
   );
 
-  // Legend
-  const legendOptions = useMemo(
-    () => ({
-      //selectedMode: false,
-      type: 'scroll',
-      left: theme.spacingValue,
-      bottom: theme.spacingValue * -0.5,
-      itemGap: theme.spacingValue * 3,
-      icon: 'circle',
-      itemWidth: theme.spacingValue,
-      itemHeight: theme.spacingValue,
-      // TODO: as prop?
-      formatter: (name) => name.toUpperCase(),
-      textStyle: {
-        ...theme.typography.overlineDelicate,
-        color: theme.palette.text.primary,
-        lineHeight: 1,
-        verticalAlign: 'bottom',
-        padding: [0, 0, 0, theme.spacingValue * 0.5]
-      },
-      inactiveColor: theme.palette.text.disabled,
-      pageIcons: {
-        horizontal: [
-          'path://M15.41 7.41 14 6 8 12 14 18 15.41 16.59 10.83 12z',
-          'path://M9 16.59 13.3265857 12 9 7.41 10.3319838 6 16 12 10.3319838 18z'
-        ]
-      },
-      pageIconSize: theme.spacingValue * 1.5,
-      pageIconColor: theme.palette.text.secondary,
-      pageIconInactiveColor: theme.palette.text.disabled,
-      pageTextStyle: {
-        fontFamily: theme.typography.overlineDelicate.fontFamily,
-        fontSize: 10,
-        color: theme.palette.text.primary
-      }
-    }),
-    [theme]
-  );
-
   // Series
   const labelOptions = useMemo(
     () => ({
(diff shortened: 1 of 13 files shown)


packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
```diff
@@ -98,38 +80,52 @@ function PieWidgetUI({
     [formatter, theme.palette.common.white, theme.palette.black, tooltipFormatter]
   );
 
-  // Series
```
vmilan, line 101: No longer needed, we use `CentralText` from now on

packages/react-ui/src/widgets/PieWidgetUI/hooks/usePieCategories.js
```diff
@@ -0,0 +1,77 @@
+import { useMemo, useRef, useEffect } from 'react';
+import { useTheme } from '@mui/material';
+import { OTHERS_CATEGORY_NAME, ORDER_TYPES } from '../../utils/chartConstants';
+import { sortDataDescending } from '../../utils/chartUtils';
+
+function usePieCategories(data, order, maxItems, colors) {
+  const theme = useTheme();
+  const colorByCategory = useRef({});
```
zbigg, line 8: I know it's old, but just raising awareness.

I wonder if keeping color mapping in ref makes sense. I would say, not.

This makes color mapping stable across data changes (apart from unmounting/remounting that resets ref, but that's other story)

I believe, If we change data we should always apply first color to first (or most dominant) category in _current_ data, not try to keep colors stable.

This is at least how i did it in TS widget.

What do you think?

So

packages/react-ui/src/widgets/PieWidgetUI/hooks/usePieCategories.js
```diff
@@ -0,0 +1,77 @@
...
+  }, [maxItems, orderedData]);
+
+  // Add a color to each category
+  const dataWithColor = useMemo(() => {
+    return groupedData.map(processDataItem(colorByCategory, colors, theme));
+  }, [groupedData, colors, theme]);
+
+  // Reset colorByCategory when colors and categories change
+  useEffect(() => {
+    colorByCategory.current = {};
```
zbigg, line 54: and this one is actually resetting `colorByCategory` on each change of data, so using ref is actually no-op

`colorByCategory` should be local to `dataWithColor` calculation

zbigg (approved):
LGTM, apart from weird and extra old logic which i believe can be fixed and some old code can be removed

packages/react-ui/src/widgets/PieWidgetUI/hooks/usePieCategories.js
vmilan: Good catch, fixed here https://github.com/CartoDB/carto-react/pull/779/commits/505f6251f853852372ab87c5f83009bb04a8e0ad

packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
```diff
@@ -98,38 +80,52 @@ function PieWidgetUI({
...
+    let sumValue = 0;
+    for (const category of processedData) {
+      sumValue += category.value;
+    }
+
+    const percentage = calculatePercentage(category.value, sumValue);
+    category.percentage = percentage;
+
+    return category;
+  }, [processedData, selectedCategories]);
```
VictorVelarde, line 114: Didn't you consider putting this hook outside, as a useTopSelectedCategory?

VictorVelarde (approved):
LGTM

packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
vmilan: I moved to the `CentralText` component

PR #780, opened 2023-09-29
cleanup: distribute only src and dist files

zbigg:
# Description

c4r publishes packages may contain lots of random files, are there reasons not to skip them

examples:
* react-ui: https://unpkg.com/browse/@carto/react-ui@2.2.7/
  * storybook-static/0.3e376249.iframe.bundle.js.LICENSE.txt …
  * storybook/*
  * firebase.json !!!
  * .firebase/… some cache …
* other packages:
  * __tests__
  * babel.config.js
   * jest.config.js
   * webpack.config.js 

I don’t think they are needed by common user, they are not going to transpile stuff or run tests against compiled base
package is not place for those

re> what others do
[deck.gl](http://deck.gl/) doesn’t package any “internal” stuff like tests or configs used to build bundles

Yes, some files are randomly picked from non-clean workspaces, but my point here is that we should only package files that we know are needed, rest should be ignored

## Type of change

- Fix

# Acceptance

Distributed packages shouldn't contain internal files. They should contain only:
 * `src/*`
 * `dist/*`
 * `package.json`
 * `LICENSE.*`
 * `README.*`
 
# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Not released
 
+- (chore) Include only required files in published packages [#780](https://github.com/CartoDB/carto-react/pull/780)
+
 ## 2.2
 
 ### 2.2.7 (2023-09-13)
--- a/packages/react-api/package.json
+++ b/packages/react-api/package.json
@@ -14,6 +14,10 @@
   },
   "license": "BSD-3-Clause",
   "main": "dist/index.js",
+  "files": [
+    "dist",
+    "src"
+  ],
   "directories": {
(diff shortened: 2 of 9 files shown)


vmilan (approved):

VictorVelarde (approved):
LGTM. Let's have an ok from PS and we can move on

aaranadev (approved):

PR #786, opened 2023-10-09
WrapperWidgetUI: add new headerItems prop to be more flexible the elements we can use

vmilan:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/353577/fix-cr4-builder-widgets-issue-when-2-lines-is-exceded
[sc-353577]

We needed a new prop to accept more variety of elements in the header, so we can remove custom code in Builder that is causing a UI bug.

### DIFF
--- a/packages/react-ui/src/widgets/WrapperWidgetUI.d.ts
+++ b/packages/react-ui/src/widgets/WrapperWidgetUI.d.ts
@@ -34,6 +34,7 @@ export type WrapperWidgetUIProps = {
   isLoading?: boolean;
   disabled?: boolean;
 
+  headerItems?: ReactNode;
   actions?: WrapperWidgetAction[];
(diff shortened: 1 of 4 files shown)


packages/react-ui/src/widgets/WrapperWidgetUI.d.ts
```diff
@@ -34,6 +34,7 @@ export type WrapperWidgetUIProps = {
   isLoading?: boolean;
   disabled?: boolean;
 
+  headerItems?: ReactNode;
   actions?: WrapperWidgetAction[];
```
VictorVelarde, line 38: @aaranadev @menusal We think actions / options props are too prescriptive. We would like to consider deprecation of those in the future, replaced by the use of this new 'headerItems' prop

PR #797, opened 2023-11-09
Server-side map filtering with widgets

padawannn:
https://app.shortcut.com/cartoteam/story/358934/server-side-filtering-in-c4r

The idea is to have a function that transforms a source (table or query) into a query taking into account the widget filters

### DIFF
--- a/packages/react-widgets/src/models/fqn.d.ts
+++ b/packages/react-widgets/src/models/fqn.d.ts
@@ -1,0 +1,69 @@
+import { Provider } from "@carto/react-core"
+
+export type Fragment = {
+  name: string,
+  quoted: boolean
+}
+export type GetIdentifierOptions = {
+  quoted?: boolean,
+  safe?: boolean
+}
+
+export type GetObjectIdentifierOptions = GetIdentifierOptions & {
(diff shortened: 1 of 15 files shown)


packages/react-widgets/src/models/fqn.d.ts
```diff
@@ -0,0 +1,69 @@
+import { Provider } from "@carto/react-core"
```
padawannn, line 1: All the code has been copied from https://github.com/CartoDB/cloud-native/blob/main/commons/lib/fqn.ts
The idea is to move it to an external module at some point.

packages/react-widgets/src/models/fqn.js
```diff
@@ -0,0 +1,317 @@
+import { Provider } from '@carto/react-core';
```
padawannn, line 1: All the code has been copied from https://github.com/CartoDB/cloud-native/blob/main/commons/lib/fqn.ts
The idea is to move it to an external module at some point.

mdeividas (approved):

VictorVelarde (approved):
LGTM

PR #804, opened 2023-11-24
Freeze widgets during time-series animation

padawannn:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/362678/timeseries-adaptation

## Type of change

- Feature


packages/react-widgets/src/widgets/TimeSeriesWidget.js
```diff
@@ -145,6 +148,7 @@ function TimeSeriesWidget({
   const dispatch = useDispatch();
 
   const [selectedStepSize, setSelectedStepSize] = useState(stepSize);
+  const [isRunningAnimation, setIsRunningAnimation] = useState(false);
```
zbigg, line 151: maybe we can use `isPlaying` and save one state variable

packages/react-redux/src/slices/cartoSlice.d.ts
```diff
@@ -18,6 +18,7 @@ type FilterBasic = {
   values: FilterValues;
   owner?: string;
   params?: Record<string, unknown>;
+  ignore?: boolean;
```
zbigg, line 21: this flag has very specific use case, maybe more verbose name would be better like `ignoreByNonOwners` or `onlyApplyToOwner` or `noCrossWidgetFiltering` ?

ignore seems too broad and misleading (it's still taken into account, but only by owner widget)

zbigg (approved):
minor comments, otherwise looks ok

packages/react-widgets/src/widgets/TimeSeriesWidget.js
zbigg: nvm, it's only used when animation is controlled externally

PR #807, opened 2023-11-28
Fix dataSources store type

padawannn:
# Description

In the redux package, the dataSources type was `dataSources: { [key: string]: string }` that obviously is wrong and must be `dataSources: { [key: string]: Source },`

## Type of change

- Fix


VictorVelarde (approved):
LGTM

PR #808, opened 2023-11-30
Added filterable prop to TimeSeriesWidget

padawannn:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/368002/add-option-in-the-widget-configurator-to-disable-filtering

It's the same property that currently we have in other widgets like the CategoryWidget

## Type of change

- Feature

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Not released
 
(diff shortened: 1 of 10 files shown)


VictorVelarde (approved):
LGTM @padawannn

PR #813, opened 2023-12-14
Alert component: UI fixes

vmilan:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/338746/implement-alert-ui-fixes-in-c4r
[sc-338746]

Alert was implemented a few months ago as a new piece of the DS, but due to deadlines, we couldn't apply the UI fixes that design team detected in the QA process. This task is to tackle those fixes.

### DIFF
--- a/packages/react-ui/src/components/molecules/Alert.js
+++ b/packages/react-ui/src/components/molecules/Alert.js
@@ -5,9 +5,12 @@ import Typography from '../atoms/Typography';
 import { ICON_SIZE_MEDIUM } from '../../theme/themeConstants';
 
 const StyledAlert = styled(MuiAlert, {
-  shouldForwardProp: (prop) => !['isNeutral', 'content', 'hasCloseButton'].includes(prop)
-})(({ isNeutral, content, hasCloseButton, theme }) => ({
-  paddingLeft: theme.spacing(1.5),
+  shouldForwardProp: (prop) =>
+    !['isNeutral', 'content', 'hasCloseButton', 'hasTitle', 'sticky'].includes(prop)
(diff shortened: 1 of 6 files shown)


packages/react-ui/src/components/molecules/Alert.js
```diff
@@ -1,106 +1,155 @@
...
+    gridTemplateColumns: hasCloseButton
+      ? `${ICON_SIZE_MEDIUM} 1fr ${theme.spacing(3)}`
+      : `${ICON_SIZE_MEDIUM}`
+  }),
+  ...(isNeutral && {
+    backgroundColor: theme.palette.default.background,
+    color: theme.palette.text.primary
+  }),
+  ...(isSticky && {
+    borderRadius: 0
```
mdeividas, line 40: It's a bit confusing that the sticky property changes the layout of the alert. Maybe we should introduce additional prop for the borderRadius change?

mdeividas (approved):

packages/react-ui/src/components/molecules/Alert.js
zbigg: I think sticky is term from design system, so it's proper term that abstracts implementation detail - border in this case.

https://www.figma.com/file/nmaoLeo69xBJCHm9nc6lEV/C4R-Components?type=design&node-id=9586-188422&mode=design&t=yujtHTpyoyJwKeT6-0

zbigg (approved):

packages/react-ui/src/components/molecules/Alert.js
vmilan: Yes, exactly. Is an abstraction affecting just a border property in this case, but it could be another different think or a set of properties, so I would keep it as `sticky`.

PR #817, opened 2023-12-29
UploadField component: UI fixes and extract pure UI components

vmilan:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/352077/cn-replace-old-code-with-uploadfield-component
[sc-352077]

Pure UI components are extracted and isolated from hooks to avoid collisions with cloud-native components behavior. I keep UploadField component to use without further development


packages/react-ui/src/components/molecules/UploadField/UploadField.d.ts
```diff
@@ -1,11 +1,17 @@
...
+  name?: string;
   buttonText?: string;
-  accept?: string[];
+  accept?: string[] | string | null;
   files?: [];
   multiple?: boolean;
-  onChange: (file?: File | null) => void;
+  onChange?: (file?: File | null) => void;
+  inProgress?: boolean;
+  inputProps?: Partial<InputProps>;
```
zbigg, line 12: For sake of consistency and understandability i would call these `InputProps` (capitalized) as in mui there is some (big and unexplored ;) difference between `input` and `Input` props ...

packages/react-ui/src/components/molecules/UploadField/UploadFieldBase.d.ts
```diff
@@ -0,0 +1,22 @@
...
+import { TextFieldProps } from '@mui/material/TextField';
+import { InputProps } from '@mui/material';
+
+export type UploadFieldBaseProps = Omit<TextFieldProps, 'placeholder' | 'error'> & {
+  name?: string;
+  multiple?: boolean;
+  handleReset?: () => void;
+  handleOpen?: () => void;
+  dragOver?: boolean;
+  error?: string | React.ReactNode;
```
zbigg, line 11: This field looks like was passed to TextField and thus previously was boolean.

In MUI textfields there is convention that `error` is "only" a boolean saying whether to render it as red. If text or node is needed it has to land in `helperText`. I think it's good to follow convention and make it back `boolean`

(yes, this is internal class, but it "defined" implementation of `error` handling in `UploadField`)

zbigg: Also not sure what if that is expected when both `error` and `helperText` are provided:

image

packages/react-ui/src/components/molecules/UploadField/StyledUploadField.js
```diff
@@ -23,7 +23,7 @@ const StyledUploadField = styled(TextField, {
     }
   },
   '& .MuiFormLabel-root': {
-    cursor: 'pointer',
+    cursor: cursor,
 
     '&.Mui-disabled': {
       pointerEvents: 'none'
```
zbigg, line 29: Really wondering if whole `...(focused && {` section below is needed.

Looks like this is really reimplementation of native behavior of TextField which ... already has `focused` boolean available ...

vmilan: I double-checked and is not needed, so removed, thanks!

packages/react-ui/src/components/molecules/UploadField/UploadField.d.ts
vmilan: Ok, changed

VictorVelarde (approved):
👍🏻

packages/react-ui/src/components/molecules/UploadField/UploadFieldBase.d.ts
vmilan: I get your point, I've reverted the change to keep `error` as boolean, if we need double `helperText` we'll manage per use case.

PR #826, opened 2024-01-22
Multiple SelectField component: fix counter version

vmilan:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/368684/core-multipleselect-fix-counter-version
[sc-368684]

### DIFF
--- a/packages/react-ui/src/components/atoms/MultipleSelectField.d.ts
+++ b/packages/react-ui/src/components/atoms/MultipleSelectField.d.ts
@@ -1,14 +1,23 @@
 import React from 'react';
 import { SelectFieldProps } from './SelectField';
 
-type MultipleSelectFieldItem = {
+type MultipleSelectFieldOption = {
   label: string | React.ReactNode;
   value: string | number;
+  disabled?: boolean;
+  tooltip?: string | false;
 };
 
-export type MultipleSelectFieldProps = SelectFieldProps & {
-  items: MultipleSelectFieldItem[];
-  itemChecked: Boolean[];
+export type MultipleSelectFieldProps = Omit<
+  SelectFieldProps,
+  'onChange' | 'defaultValue' | 'value'
+> & {
+  options: MultipleSelectFieldOption[];
+  selectedOptions?: string[];
+  selectAllDisabled?: boolean;
+  onChange: (options: string[]) => void;
+  showCounter?: boolean;
+  showFilters?: boolean;
 };
 
 declare const MultipleSelectField: (props: MultipleSelectFieldProps) => JSX.Element;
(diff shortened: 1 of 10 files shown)


packages/react-ui/src/components/molecules/MultipleSelectField.js
```diff
@@ -0,0 +1,250 @@
...
+              title={option.tooltip}
+              placement={tooltipPlacement}
+            >
+              {item}
+            </Tooltip>
+          ) : (
+            item
+          );
+          return content;
+        })}
```
VictorVelarde, line 222: We could consider extracting this into its own smaller component, to reduce a bit the size / complexity here

packages/react-ui/storybook/stories/molecules/MiltipleSelectField.stories.js
```diff
@@ -0,0 +1,225 @@
...
+    </DocContainer>
+  );
+};
+
+const DisabledWithTooltipTemplate = ({ ...rest }) => {
+  const [selectedLayerNames, setLayerNames] = useState([]);
+
+  const menuItems = [
+    {
+      label: 'table_openstreetmapt',
```
VictorVelarde, line 153: typo table_openstreetmap

packages/react-ui/src/components/molecules/MultipleSelectField.js
vmilan: I extracted the Filters and the logic into a hook, but this part needs more research, when I extract it into a component, it stops working... 

I prefer to do it in other PR, because I need to jump to the next task and I already did extra work here, like adding the filters themselves to be able to replace/remove code in cloud-native (which btw I also did and is not mandatory for the PFB).

So if you can approve the PR, I'll merge the branch and create the patch version 🙏

vmilan: Here are the changes: https://github.com/CartoDB/carto-react/pull/826/commits/c53155ba23fc3416d2a4eecc3fa28681b80f0fc7I extracted the Filters and the logic into a hook, but this part needs more research, when I extract it into a component, it stops working... 

I prefer to do it in other PR, because I need to jump to the next task and I already did extra work here, like adding the filters themselves to be able to replace/remove code in cloud-native (which btw I also did and is not mandatory for the PFB).

So if you can approve the PR, I'll merge the branch and create the patch version 🙏

Josmorsot (approved):

packages/react-ui/src/components/molecules/MultipleSelectField/useMultipleSelectField.js
```diff
@@ -0,0 +1,51 @@
...
+
+  useEffect(() => {
+    if (currentOptions !== selectedOptions) {
+      setCurrentOptions(currentOptions);
+    }
+    // intentionally ignore currentOptions as we only want to trigger on external updates
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedOptions]);
+
+  const handleChange = (event) => {
```
VictorVelarde, line 17: Nice refactor moving out the logic & handlers

VictorVelarde (approved):
LGTM, thx for taking care of the refactor

PR #830, opened 2024-01-26
[react-redux, react-widgets] Add 'GeocoderResult' types, add .passive property

donmccurdy:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/378094/search-location
[sc-378094]

We need the option of setting the geocoder result state without moving the map viewport. This PR creates type definitions for the `GeocoderResult` interface, and adds a new `.passive` property used to indicate when a result must not change the viewport.

## Type of change

- [x] Feature


zbigg (approved):
lgtm

PR #843, opened 2024-02-28
Add formatter to central text in PieWidgetUI

jantolg:
# Description
Add to central text in PieChartUI the same text formatter

- [Slack thread](https://cartoteam.slack.com/archives/C01B3D52HJB/p1709021888019739?thread_ts=1709021646.201919&cid=C01B3D52HJB
)
- Feature

# Acceptance

1. Go to PieWidgetUI
2. Use a custom formatter
3. Check the central text as the same format as you set at formatter func

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # CHANGELOG
 
 ## Not released
-
+- Feat PieWidgetUI: Add formatter to central text component and include a default formatted based on useIntl.formatNumber [#843](https://github.com/CartoDB/carto-react/pull/843)
+  
 ## 2.3
 
 ### 2.3.12 (2024-02-19)
--- a/packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
+++ b/packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
@@ -67,7 +67,10 @@ function PieWidgetUI({
   const { showSkeleton } = useSkeleton(isLoading);
   const intl = useIntl();
   const intlConfig = useImperativeIntl(intl);
-
+  let _formatter = formatter;
+  if (!_formatter) {
+    _formatter = (val) => `${intl.formatNumber(val, { maximumFractionDigits: 2 })}%`;
+  }
   // Tooltip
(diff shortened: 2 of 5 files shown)


VictorVelarde (approved):
Have a look at it please @vmilan, LGTM

vmilan (commented):
Could you add an example of use in [storybook](https://storybook-react.carto.com/?path=/docs/widgets-piewidgetui--default) and a new entry in the changelog? 🙏

VictorVelarde:
The approach LGTM, just 2 caveats:
- The new option to add a formatter deserves a story in storybook, don't you think so? That way it will show it in real action.
- Default behaviour: I think we should preserve current behaviour, so keep old formatter if non-formatter is passed down. Otherwise users will get something like https://cartodb-fb-storybook-react-dev--pr843-feat-add-central-nmfupptg.web.app/?path=/docs/widgets-piewidgetui--default 
image


Thx for the help here @jantolg

jantolg:
I've just to add an example of storybook, a new changelog entry, and a default behaviour of formatter based on useIntl.formatNumber

vmilan (commented):
I've seen a weird behavior in the storybook generated with your changes:
https://cartodb-fb-storybook-react-dev--pr843-feat-add-central-nmfupptg.web.app/?path=/docs/widgets-piewidgetui--default

master:
https://storybook-react.carto.com/?path=/docs/widgets-piewidgetui--default

At least in Storybook something is broken.

https://github.com/CartoDB/carto-react/assets/1934144/923ad565-dffb-4239-a40e-f6a95872beea

Please check the widget is still working properly, we use it in Builder.

jantolg:
I close this PR.

Upon reviewing the behavior, maybe using the same formatter for units and percentage values is not a good approach, as it represents two different types of values. After discussing this issue with @aaranadev and @menusal, we believe it would be better to have a separate formatter for each type of value.

I have created a new PR with these changes to avoid mixing different approaches.

NOTE: It might be worth considering a standardized approach for using formatters, perhaps by unifying all formatters into a single property or something similar. This would be a good topic to open a discussion about for future C4R versions

Continue in this new PR #844

PR #844, opened 2024-03-04
feat: add percentFormatter prop to PieWidgetUI

jantolg:
# Description
Add to central text in PieChartUI the same text formatter

- [Slack thread](https://cartoteam.slack.com/archives/C01B3D52HJB/p1709021888019739?thread_ts=1709021646.201919&cid=C01B3D52HJB)
- Feature

# Acceptance

1. Go to PieWidgetUI
2. Add percentFormatter prop
3. Check the central text and tootltip have the same format as you set

PR #850, opened 2024-03-06
fix: keep always two digits at PieWidgetUI default percentFormatter

jantolg:
# Description
Related by: https://github.com/CartoDB/carto-react/pull/844

Keep two decimals always in default percent formatters in PieWidgetUI

## Type of change
- Fix

# Acceptance

1. go to PieWidgetUI
2. Check percents format should have two decimals always, if you didn't defined a different percentFormatter

### DIFF
--- a/packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
+++ b/packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
@@ -71,7 +71,11 @@ function PieWidgetUI({
   const _percentFormatter = useMemo(
     () =>
       percentFormatter ||
-      ((v) => `${intl.formatNumber(v, { maximumFractionDigits: 2 })}%`),
+      ((v) =>
+        `${intl.formatNumber(v, {
+          maximumFractionDigits: 2,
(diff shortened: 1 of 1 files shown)


packages/react-ui/src/widgets/PieWidgetUI/PieWidgetUI.js
```diff
@@ -71,7 +71,11 @@ function PieWidgetUI({
   const _percentFormatter = useMemo(
     () =>
       percentFormatter ||
-      ((v) => `${intl.formatNumber(v, { maximumFractionDigits: 2 })}%`),
+      ((v) =>
```
vmilan, line 74: As this has caused problems with e2e, please double-check the case when `total === 0`, it should returns `0,00%`

It was also modified in your last PR -> https://github.com/CartoDB/carto-react/pull/844/files#diff-aef7c79c208e732a247556c69887e48f3f8b2076d3bf28aeeb9abde1419806adL98-L99

padawannn: Checked

padawannn (approved):

PR #865, opened 2024-04-19
TimeSeriesWidget: fix echarts props update, to keep state of control when clicking

zbigg:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/404593/builder-timeserieswidget-tooltip-disappears-when-clicking-on-series

In `TimeseriesWidget`, fix way we update props in `echarts` instance so 
 * series are effectively removed
 * we don't lose state of control on props update

(note, data update still resets control animation/filter state)

Fixes regression from #863 
## Type of change

- Fix

# Acceptance

1. go to X
2. test Y
3. assert Z

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-ui/src/custom-components/echarts-for-react/index.js
+++ b/packages/react-ui/src/custom-components/echarts-for-react/index.js
@@ -49,6 +49,20 @@ export default class ReactEcharts extends _ReactEcharts {
     }
   }
 
+  updateEChartsOption() {
+    const { option, lazyUpdate, showLoading, loadingOption = null } = this.props;
+    const echartInstance = this.getEchartsInstance();
+
+    echartInstance.setOption(option, { replaceMerge: ['series'], lazyUpdate });
+
+    if (showLoading) {
+      echartInstance.showLoading(loadingOption);
+    } else {
+      echartInstance.hideLoading();
+    }
+    return echartInstance;
+  }
+
   offEvents(instance, events) {
     if (!events) return;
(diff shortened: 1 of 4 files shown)


packages/react-ui/src/widgets/TimeSeriesWidgetUI/components/TimeSeriesChart.js
```diff
@@ -197,6 +197,7 @@ export default function TimeSeriesChart({
             : theme.palette.action.disabledBackground;
 
         return {
+          id: name,
```
zbigg, line 200: This ensures that echarts keeps track of series when we update data.

packages/react-ui/src/custom-components/echarts-for-react/index.js
```diff
@@ -49,6 +49,20 @@ export default class ReactEcharts extends _ReactEcharts {
     }
   }
 
+  updateEChartsOption() {
+    const { option, lazyUpdate, showLoading, loadingOption = null } = this.props;
+    const echartInstance = this.getEchartsInstance();
+
+    echartInstance.setOption(option, { replaceMerge: ['series'], lazyUpdate });
```
zbigg, line 56: This ensures that merges `series` objects as list identified by `id`

PR #877, opened 2024-06-11
fix(react-ui): Fix LegendProportion radius scale

donmccurdy:
# Description

Shortcut: Story 415171, URL redacted.

LegendProportion displays the legend for point scaling, and sometimes calculates the scale incorrectly. In the image below, note that the 'max' is smaller than the second number, and the 'min' is larger than the third.

The cause is a typo, we should be using `max - min` instead of `max + min` to calculate the interval over which the two middle steps are interpolated. With this bug, the steps were calculated correctly only when `min` was zero.

Additionally, the visual sizes of the circles in the legend panel are hard-coded to `[3, 6, 9, 12]`, representing two evenly-spaced steps between the min and max. However, our implementation would have generated different steps given `min=3, max=12` as input, because it divided the interval by 4, not 3.

| before | after |
|---|---|
| legend_range_bug | Screenshot 2024-06-11 at 12 39 00 PM |

## Type of change

- [x] Fix

# Acceptance

1. Link C4R into Builder
2. Create a new map
3. Add SQL query

    ```sql
    SELECT * FROM `carto-demo-data.demo_tables.cell_towers_worldwide` AS table
    WHERE (table.lat - 36.744874) < 5 AND (table.lon - 3.195371) < 5;
    ```

4. Set point layer to style radius by `lat` column
5. Open legend, observe steps are correctly spaced

### DIFF
--- a/packages/react-ui/src/widgets/legend/legend-types/LegendProportion.js
+++ b/packages/react-ui/src/widgets/legend/legend-types/LegendProportion.js
@@ -149,10 +149,11 @@ function calculateRange(legend) {
   return { min, max, error };
 }
 
+/**
+ * Calculates two evenly-spaced steps, linearly interpolated between a given
+ * min and max. For example, `calculateSteps(3, 12)` gives `[6, 9]`.
+ */
 function calculateSteps(min, max) {
-  const gap = (max + min) / 4;
-  const step1 = min + gap;
-  const step2 = max - gap;
-
-  return [step1, step2];
+  const step = (max - min) / 3;
+  return [min + step, max - step];
 }
--- a/packages/react-ui/src/widgets/legend/legend-types/LegendProportion.js
+++ b/packages/react-ui/src/widgets/legend/legend-types/LegendProportion.js
@@ -5,6 +5,8 @@ import Typography from '../../../components/atoms/Typography';
 import { useIntl } from 'react-intl';
 import useImperativeIntl from '../../../hooks/useImperativeIntl';
 
+const MAX_DIGITS = 6;
+
 const sizes = {
   0: 12,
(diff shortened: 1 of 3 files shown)


packages/react-ui/src/widgets/legend/legend-types/LegendProportion.js
```diff
@@ -91,10 +91,10 @@ function LegendProportion({ legend }) {
                 : max}
             </Typography>
             <Typography variant='overline' color='textSecondary'>
-              {step2}
+              {intl.formatNumber(step2, { maximumFractionDigits: 6 })}
             </Typography>
             <Typography variant='overline' color='textSecondary'>
-              {step1}
+              {intl.formatNumber(step1, { maximumFractionDigits: 6 })}
```
donmccurdy, line 97: Note – I chose 6 digits since that's what the min/max passed into this component were already rounded to, but I'm not sure where/when they are rounded. Open to other suggestions.

zbigg: I'm strict about these, i try to use max _significant_ digits.

It would be great to differentiate between `0.0000001` and `0.0000003` and i'm not sure what this particular API will to with these numbers.

Anyway, so number with so many leading/trailing zeros without units or scientific notation are unusable anyway,

zbigg (approved):
lgtm

packages/react-ui/src/widgets/legend/legend-types/LegendProportion.js
donmccurdy: Good call! Agreed that max significant digits should be a better choice here, I've added it.

PR #879, opened 2024-06-13
fix(core): Add getUTCMonday, fix tests in U.S. timezone

donmccurdy:
# Description

Currently the tests do not pass when running locally from a U.S. time zone, fixed in this PR. Existing code in `groupByDate.js` used UTC time for all groupings, and `TimeSeriesWidget` uses local time, but both call into `getMonday` which cannot do both ... so I think it's likely that a UTC-specific version of `getMonday` is required.

Changes:

- **getMonday:** Returns midnight (local time) on the Monday preceeding a given date, in milliseconds since the UNIX epoch.
- **getUTCMonday:** Returns midnight (UTC) on the Monday preceeding a given date, in milliseconds since the UNIX epoch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Not released
 
+- Fix time zone handling in week counts, separate getMonday and getUTCMonday utilities [#879](https://github.com/CartoDB/carto-react/pull/879)
+
 ## 3.0.0
 
 ### 3.0.0-alpha.12 (2024-06-14)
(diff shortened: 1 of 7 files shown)


vmilan (commented):
Please, add an entry to the changelog on every PR 🙏 

I added for you the last one that was missing, here: https://github.com/CartoDB/carto-react/pull/881/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

jmgaya (approved):
May the time be with you, now... 🕐

donmccurdy:
@vmilan changelog added, thank you!

PR #894, opened 2024-07-26
Cache Intl instances to improve performance

zbigg:
# Description

Cache intl instances globally.

`flattenMessages` over _c4r_ and app could take up to 2-4 milliseconds (on fast machine). And with current, `useMemo` approach it was executed for each component mounted .. and lost when unmounted.

Real app can mount even 10+ c4r+intl-based components a time, causing some mounts to miss sane deadlines and cause warnings like:

> [Violation] 'click' handler took 106ms

(it's not that this call is only one slow in big components, but this can be easily avoided as `intlConfig` instance is generally stable)

We also don't want to waste memory to keep next 10+ copies of all messages (3.5k keys)

## Type of change

- Fix
- Refactor

# Acceptance

N/A performance issue

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-ui/src/hooks/useImperativeIntl.js
+++ b/packages/react-ui/src/hooks/useImperativeIntl.js
@@ -8,22 +8,40 @@ import {
 } from '../localization/localeUtils';
 
 const cache = createIntlCache();
+const intlInstanceCache = new WeakMap();
+
+const createIntlInstance = (intlConfig) => {
+  const locale = intlConfig?.locale || DEFAULT_LOCALE;
+  const messagesLocale = findMatchingMessagesLocale(locale, messages);
+  const intMessages = {
+    ...(messages[messagesLocale] || {}),
+    ...(intlConfig?.messages || {})
+  };
+
+  const combinedMessages = flattenMessages(intMessages);
+  return createIntl(
+    {
+      locale,
+      messages: combinedMessages
+    },
+    cache
+  );
+};
+
+const getGloballyCachedIntl = (intlConfig) => {
+  // This is very simple cache exploits fact that Intl instance is actually same for most of time
+  // so we can reuse those maps across several instances of same components
+  // note, useMemo can't cache accross many that globally and flattenMessages over _app_ and c4r messages is quite costly
+  // and would be paid for every c4r component mounted.
+  let cachedInstance = intlInstanceCache.get(intlConfig);
+  if (cachedInstance) {
+    return cachedInstance;
+  }
+  const newInstance = createIntlInstance(intlConfig);
+  intlInstanceCache.set(intlConfig, newInstance);
+  return newInstance;
+};
 
 export default function useImperativeIntl(intlConfig) {
-  return useMemo(() => {
-    const locale = intlConfig?.locale || DEFAULT_LOCALE;
-    const messagesLocale = findMatchingMessagesLocale(locale, messages);
-    const intMessages = {
-      ...(messages[messagesLocale] || {}),
-      ...(intlConfig?.messages || {})
-    };
-
-    return createIntl(
-      {
-        locale,
-        messages: flattenMessages(intMessages)
-      },
-      cache
-    );
-  }, [intlConfig]);
+  return getGloballyCachedIntl(intlConfig);
 }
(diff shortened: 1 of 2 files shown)


packages/react-ui/src/hooks/useImperativeIntl.js
```diff
@@ -9,21 +8,31 @@ import {
...
+let lastIntlConfig;
+let cachedC4rIntl;
+
 export default function useImperativeIntl(intlConfig) {
-  return useMemo(() => {
+  if (!cachedC4rIntl || lastIntlConfig !== intlConfig) {
+    // this is very naive cache that bases on fact that Intl instance is actually same for most of time
+    // so we can reuse those maps across several instances of same components
+    // note, useMemo can't do that globally and flattenMessages over _app_ and c4r is quite costly and would
+    // be paid for evey c4r component mounted
```
zbigg, line 19: This cache assumes we have 1 active Intl instance in app which is usually the case.

It _can_ be changed to WeakMap<IntlConfig, IntlShape> if we have use case where one app use many languages at once, not sure if that's the case

packages/react-ui/src/hooks/useImperativeIntl.js
```diff
@@ -9,21 +9,36 @@ import {
 
 const cache = createIntlCache();
 
-export default function useImperativeIntl(intlConfig) {
-  return useMemo(() => {
+let lastIntlConfig;
+let cachedC4rIntl;
```
donmccurdy, line 13: Optional — For me this would be a little more readable if we omit the "C4r" abbreviation and align the two names. Perhaps:

```suggestion
let cachedIntlConfig;
let cachedIntl;
```

packages/react-ui/src/hooks/useImperativeIntl.js
```diff
@@ -9,21 +9,36 @@ import {
...
     );
-  }, [intlConfig]);
+    lastIntlConfig = intlConfig;
+  }
+  return cachedC4rIntl;
+};
+
+export default function useImperativeIntl(intlConfig) {
+  // second level cache, in components is just to avoid re-creating the Intl instance if user is rendering many languages in one app
+  return useMemo(() => getGloballyCachedIntl(intlConfig), [intlConfig]);
```
donmccurdy, line 43: The second-level cache only matters if we're rendering multiple languages _simultaneously_, right? If we just have a language dropdown, and switch from one language to another, the primary cache should be enough? 

I wonder if a single global cache of `WeakMap<IntlConfig, Intl>` would be simpler than two caches? I'm happy with either, though. I assume using multiple languages simultaneously is rare (maybe just SSR?).

donmccurdy (approved):
LGTM!

packages/react-ui/src/hooks/useImperativeIntl.js
zbigg: I've backed off to WeakMap based solution.
This weakmap will mostly have one entry so shouldn't be slower than useMemo.

@donmccurdy Can you take another view?

packages/react-ui/src/hooks/useImperativeIntl.js
zbigg: Changed to WeakMap, so we're on safe side.

packages/react-ui/src/hooks/useImperativeIntl.js
```diff
@@ -8,22 +8,40 @@ import {
...
-      },
-      cache
-    );
-  }, [intlConfig]);
+const getGloballyCachedIntl = (intlConfig) => {
+  // This is very simple cache exploits fact that Intl instance is actually same for most of time
+  // so we can reuse those maps across several instances of same components
+  // note, useMemo can't cache accross many that globally and flattenMessages over _app_ and c4r messages is quite costly
+  // and would be paid for every c4r component mounted.
+  let cachedInstance = intlInstanceCache.get(intlConfig);
```
donmccurdy, line 36:
```suggestion
  const cachedInstance = intlInstanceCache.get(intlConfig);
```

donmccurdy (approved):

PR #898, opened 2024-07-31
Remote widget calculation for  spatial indexed sources

zbigg:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/425306

1. Spatial index sources use remote widgets calculation.
2. addSource: Support for `spatialDataType`, `spatialIndexColumn` instead of old `geoColumn`, at least for remote widgets
3. remote widgets: support for new parameters `spatialDataType`, `spatialIndexColumn`, `spatialFiltersResolution` and `spatialFiltersMode`

## Type of change

- Feature

# Acceptance

...

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-api/src/api/model.js
+++ b/packages/react-api/src/api/model.js
@@ -71,22 +71,24 @@ export function executeModel(props) {
     let spatialDataType = source.spatialDataType;
     let spatialDataColumn = source.spatialDataColumn;
 
-    if ((!spatialDataType || !spatialDataColumn) && source.geoColumn) {
-      const parsedGeoColumn = source.geoColumn ? source.geoColumn.split('.') : [];
-      if (parsedGeoColumn.length === 2) {
-        spatialDataType = parsedGeoColumn[0];
-        spatialDataColumn = parsedGeoColumn[1];
-      } else if (parsedGeoColumn.length === 1) {
-        spatialDataColumn = parsedGeoColumn[0] || DEFAULT_GEO_COLUMN;
+    if (!spatialDataType || !spatialDataColumn) {
+      if (source.geoColumn) {
+        const parsedGeoColumn = source.geoColumn ? source.geoColumn.split(':') : [];
+        if (parsedGeoColumn.length === 2) {
+          spatialDataType = parsedGeoColumn[0];
+          spatialDataColumn = parsedGeoColumn[1];
+        } else if (parsedGeoColumn.length === 1) {
+          spatialDataColumn = parsedGeoColumn[0] || DEFAULT_GEO_COLUMN;
+          spatialDataType = 'geo';
+        }
+        if (spatialDataType === 'geom') {
+          // fallback if for some reason someone provided old `geom:whatever`
+          spatialDataType = 'geo';
+        }
+      } else {
         spatialDataType = 'geo';
+        spatialDataColumn = DEFAULT_GEO_COLUMN;
       }
-      if (spatialDataType === 'geom') {
-        // fallback if for some reason someone provided old `geom:whatever`
-        spatialDataType = 'geo';
-      }
-    } else {
-      spatialDataType = 'geo';
-      spatialDataColumn = DEFAULT_GEO_COLUMN;
     }
 
     // API supports multiple filters, we apply it only to geometry column or spatialDataColumn
@@ -98,10 +100,12 @@ export function executeModel(props) {
 
     queryParams.spatialFilters = JSON.stringify(spatialFilters);
     queryParams.spatialDataType = spatialDataType;
-    queryParams.spatialFiltersMode = source.spatialFiltersMode;
     if (spatialDataType !== 'geo') {
       // TODO: any sane default (?)
-      queryParams.spatialFiltersResolution = source.spatialFiltersResolution || 10;
+      if (source.spatialFiltersResolution !== undefined) {
+        queryParams.spatialFiltersResolution = source.spatialFiltersResolution;
+      }
+      queryParams.spatialFiltersMode = source.spatialFiltersMode || 'intersects';
     }
   }
 
--- a/packages/react-api/src/api/model.js
+++ b/packages/react-api/src/api/model.js
@@ -101,7 +101,6 @@ export function executeModel(props) {
     queryParams.spatialFilters = JSON.stringify(spatialFilters);
     queryParams.spatialDataType = spatialDataType;
     if (spatialDataType !== 'geo') {
-      // TODO: any sane default (?)
       if (source.spatialFiltersResolution !== undefined) {
         queryParams.spatialFiltersResolution = source.spatialFiltersResolution;
       }
--- a/packages/react-api/src/api/model.js
+++ b/packages/react-api/src/api/model.js
@@ -73,7 +73,7 @@ export function executeModel(props) {
 
     if (!spatialDataType || !spatialDataColumn) {
       if (source.geoColumn) {
-        const parsedGeoColumn = source.geoColumn ? source.geoColumn.split(':') : [];
+        const parsedGeoColumn = source.geoColumn.split(':');
         if (parsedGeoColumn.length === 2) {
           spatialDataType = parsedGeoColumn[0];
           spatialDataColumn = parsedGeoColumn[1];
@@ -82,7 +82,7 @@ export function executeModel(props) {
           spatialDataType = 'geo';
         }
         if (spatialDataType === 'geom') {
-          // fallback if for some reason someone provided old `geom:whatever`
+          // fallback if for some reason someone provided old `geom:$column`
           spatialDataType = 'geo';
         }
       } else {
--- a/packages/react-api/src/api/model.js
+++ b/packages/react-api/src/api/model.js
@@ -66,15 +66,43 @@ export function executeModel(props) {
     filtersLogicalOperator
   };
 
-  // API supports multiple filters, we apply it only to geoColumn
-  const spatialFilters = spatialFilter
-    ? {
-        [source.geoColumn ? source.geoColumn : DEFAULT_GEO_COLUMN]: spatialFilter
-      }
-    : undefined;
+  let spatialFilters;
+  if (spatialFilter) {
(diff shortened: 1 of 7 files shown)


packages/react-api/src/api/model.js
```diff
@@ -66,15 +66,46 @@ export function executeModel(props) {
...
+    spatialFilters = spatialFilter
+      ? {
+          [spatialDataColumn]: spatialFilter
+        }
+      : undefined;
 
-  if (spatialFilters) {
     queryParams.spatialFilters = JSON.stringify(spatialFilters);
+    queryParams.spatialDataType = spatialDataType;
+    if (spatialDataType !== 'geo') {
```
jmgaya, line 103: I see `geo` repeated a few times in this file, so what do you think about extracting it somewhere and naming it as something like `const DEFAULT_SPATIAL_DATA_TYPE`?

packages/react-api/src/api/model.js
```diff
@@ -66,15 +66,46 @@ export function executeModel(props) {
...
+      if (source.spatialFiltersResolution !== undefined) {
```
jmgaya, line 104: Out of curiosity: Could `source.spatialFiltersResolution` be `null` at some point? If so, maybe we can double negate this expression to prevent this case, or is there any edge case related to a 0 resolution?

packages/react-api/src/types.d.ts
```diff
@@ -43,6 +43,10 @@ export type SourceProps = {
   type: MapTypesType['QUERY'] | MapTypesType['TABLE'] | MapTypesType['TILESET'];
   connection: string;
   geoColumn?: string;
+  dataResoultion?: number;
```
jmgaya, line 46: Seems like there's a typo and should be `dataResolution` instead of `dataResoultion`

packages/react-widgets/src/hooks/useWidgetFetch.js
```diff
@@ -93,6 +107,36 @@ export default function useWidgetFetch(
     [global, viewport, spatialFilter]
   );
 
+  const source2 = useMemo(() => {
```
jmgaya, line 110: Probably a more descriptive name for this parameter would help our future ourselves? What about `enrichedSource`?

jmgaya (approved):
Everything seems correct and the enhancements seem correct for supporting spatial indexes, at least from my side/knowledge, great job!

packages/react-api/src/api/model.js
```diff
@@ -66,15 +66,46 @@ export function executeModel(props) {
...
-    ? {
-        [source.geoColumn ? source.geoColumn : DEFAULT_GEO_COLUMN]: spatialFilter
+  let spatialFilters;
+  if (spatialFilter) {
+    let spatialDataType = source.spatialDataType;
+    let spatialDataColumn = source.spatialDataColumn;
+
+    if (!spatialDataType || !spatialDataColumn) {
+      if (source.geoColumn) {
+        const parsedGeoColumn = source.geoColumn ? source.geoColumn.split(':') : [];
```
donmccurdy, line 76: Checking `source.geoColumn` redundantly here – probably one or the other isn't required?

packages/react-api/src/api/model.js
```diff
@@ -66,15 +66,46 @@ export function executeModel(props) {
...
+        if (parsedGeoColumn.length === 2) {
+          spatialDataType = parsedGeoColumn[0];
+          spatialDataColumn = parsedGeoColumn[1];
+        } else if (parsedGeoColumn.length === 1) {
+          spatialDataColumn = parsedGeoColumn[0] || DEFAULT_GEO_COLUMN;
+          spatialDataType = 'geo';
+        }
```
donmccurdy, line 83: Could `parsedGeoColumn` really be `[]`? If so do we need a third case for it?

packages/react-api/src/types.d.ts
```diff
@@ -43,6 +43,10 @@ export type SourceProps = {
...
+  spatialDataType?: string;
+  spatialDataColumn?: string;
+  spatialFiltersResolution?: number;
```
donmccurdy, line 49: The `spatialFilter` parameter is singular, not plural – assuming these are related, should this be `spatialFilterResolution` instead? Note that this may become a public-facing API once added to `@carto/api-client`, though I do see the note in the AD that we may not expose it.

packages/react-widgets/src/hooks/useWidgetFetch.js
```diff
@@ -57,6 +57,19 @@ export function selectGeometryToIntersect(global, viewport, spatialFilter) {
...
+  // Difference in given tile size compared to deck's internal 512px tile size,
+  // expressed as an offset to the viewport zoom.
+  const zoomOffset = Math.log2(tileSize / 512);
+  const hexagonScaleFactor = (2 / 3) * (viewport.zoom - zoomOffset);
+  const latitudeScaleFactor = Math.log(1 / Math.cos((Math.PI * viewport.latitude) / 180));
+
+  // Clip and bias
+  return Math.max(0, Math.floor(hexagonScaleFactor + latitudeScaleFactor - BIAS));
+}
+
```
donmccurdy, line 72: (no action required) @felixpalmer maybe this is something we'll want to pull into `@carto/api-client` to reuse in both places? Once the repository is public and can be made a dependency, that is.

packages/react-redux/src/slices/cartoSlice.js
```diff
@@ -204,6 +204,8 @@ export const createCartoSlice = (initialState) => {
  * @param {FiltersLogicalOperators=} data.filtersLogicalOperator - logical operator that defines how filters for different columns are joined together.
  * @param {import('@deck.gl/carto').QueryParameters} data.queryParameters - SQL query parameters.
  * @param {string=} data.geoColumn - (optional) name of column containing geometries or spatial index data.
+ * @param {number=} data.dataResolution - data resolution for spatial index data.
```
donmccurdy, line 207: Is this connected to `spatialDataType` and `spatialDataColumn`, and would `spatialDataResolution` be a clearer name if so? For my own learning, is more information on what the parameters do available somewhere?

packages/react-api/src/api/model.js
zbigg: I don't think so (and i've checked as i wasn't sure):

```
> "".split(':')
['']
> "a".split(':')
['a']
> "a:b".split(':')
['a', 'b']
```

packages/react-api/src/types.d.ts
zbigg: `spatialFiltersResolution` is term already used in platform maps API. Didn't invent this name

I think, creating another slightly different symbol would be confusing, so i would leave it as it is.

@thedae @josemaria-vilaplana I leave this comment to You. Maybe we should broaden reviews of new platform API that is about to be visible in many places us for long time

zbigg: Sorry, platform APIs already spell this as `spatialFilters` so naming problem is only visible in "carto4react" source API.

@donmccurdy I think we're free to fix it in new frontend packages/APIs. I would leave it as it is in c4r which - i believe - is in deprecated mode and we don't refactor things here anymore

packages/react-redux/src/slices/cartoSlice.js
zbigg: We don't understand connection "well" enogh yet.

`dataResolution` is property of tileset - namely, maximum resolution of h3 or quadbin index encountered in table.

To apply filters, we need to apply _some_ resolution that is not bigger than `dataResolution`

The idea is that we allow users to provide two:
 * `dataResolution` - so c4r is smart enough to determine some resolution from property received in map instantiation (`dataResolution` exists in tilejson response)
 * `spatialFiltersResolution` - if user (like builder) wants to control this value directly.
 
 
That's why we have this `getHexagonResolution` - it's only used when `spatialFiltersResolution` is needed (h3/quadbin) and user didn't provided it, so we derive it from `dataResolution`.

packages/react-api/src/api/model.js
zbigg: Well, this we can't protect user from themselves.

If he passed number, he's ok. If he didn't passed anything (undefined), he's ok - default is used. If he passed null or object or date, it's his own fault - i don't believe that we should check _every_ possible failure mode if we already defined that `spatialFiltersResolution` is `optional number`.

This is just API wrapper, nothing will explode - just weird stuff will be sent to.

(and btw, this is internal code - validation if required should be placed somewhere on top of widget)

packages/react-widgets/src/hooks/useWidgetFetch.js
felixpalmer: Hmm, wouldn't it be better to export it from deck? It's a key part of how `H3Tileset` works

packages/react-api/src/types.d.ts
donmccurdy: Ok — sounds good to me!

packages/react-widgets/src/hooks/useWidgetFetch.js
donmccurdy: Oh, yes – that does sound better. :)

packages/react-redux/src/slices/cartoSlice.js
donmccurdy: Ok, thanks! I guess it's a request for the API team more, but a glossary in docs would be really helpful. 😓

packages/react-api/src/api/model.js
jmgaya: I was thinking more about something like `!!source.spatialFiltersResolution`, WDYT?

jmgaya (approved):

packages/react-api/src/api/model.js
zbigg: I don't like `!!source.spatialFiltersResolution` for one reason: who said, 0 is bad value for this parameter (is it?). I only care if user provided it or not - and that's why we have `undefined` in JS.

PR #900, opened 2024-09-06
Retrieve spatialDataColumn in TableWidget for its future usage

jmgaya:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/438209/add-support-for-geocolumn-spatialdatacolumn-when-retrieving-table-widget-information?cf_workflow=500091853&ct_workflow=all&vc_group_by=day

Sending this to the API should give us back the information about the `spatialDataColumn`, which we'll use afterward for zooming into the coordinate

zoomtopoint

## Type of change

- Feature

### DIFF
--- a/packages/react-widgets/src/widgets/TableWidget.js
+++ b/packages/react-widgets/src/widgets/TableWidget.js
@@ -14,7 +14,7 @@ import { _FeatureFlags, _hasFeatureFlag } from '@carto/react-core';
  * @param  {string} props.title - Title to show in the widget header.
  * @param  {string} props.dataSource - ID of the data source to get the data from.
  * @param  {Column[]} props.columns - List of data columns to display.
- * @param  {Column[]} props.hiddenColumnsFields - List of data columns to be retrieved, but not displayed.
+ * @param  {Column[]} props.hiddenColumnFields - List of data columns to be retrieved, but not displayed.
  * @param  {Function=} props.onRowClick - Function to handle on click events on rows.
  * @param  {Function=} [props.onError] - Function to handle error messages from the widget.
  * @param  {Function=} [props.onStateChange] - Callback to handle state updates of widgets
@@ -34,7 +34,7 @@ function TableWidget({
   title,
   dataSource,
   columns,
-  hiddenColumnsFields = [],
+  hiddenColumnFields = [],
   wrapperProps,
   noDataAlertProps,
   onError,
@@ -64,7 +64,7 @@ function TableWidget({
     id,
     dataSource,
     params: {
-      columns: [...columns.map((c) => c.field), ...hiddenColumnsFields],
+      columns: [...columns.map((c) => c.field), ...hiddenColumnFields],
       sortBy,
       sortDirection,
       sortByColumnType,
@@ -144,7 +144,7 @@ TableWidget.propTypes = {
       align: PropTypes.oneOf(['left', 'right'])
     })
   ).isRequired,
-  hiddenColumnsFields: PropTypes.arrayOf(PropTypes.string),
+  hiddenColumnFields: PropTypes.arrayOf(PropTypes.string),
   onError: PropTypes.func,
   wrapperProps: PropTypes.object,
   noDataAlertProps: PropTypes.object,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Not released
 
 - Spatial Index Sources use remote widgets calculation [#898](https://github.com/CartoDB/carto-react/pull/898)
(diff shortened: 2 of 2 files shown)


DEVELOPERS.md
```diff
@@ -31,6 +31,20 @@ Then, inside the proper template folder in carto-react-template, link packages w
...
+
+```
+./copy-packages.sh ~/workspace/repositories/cloud-native/workspace-www/node_modules/@carto
+```
+
+Thus, combined with the execution of `yarn build` in the root of this repository, will copy every package into the target directory:
+
+```
+yarn build && ./copy-packages.sh ~/workspace/repositories/cloud-native/workspace-www/node_modules/@carto
+```
```
jmgaya, line 46: @zbigg I've added this, as we both sometimes copy&paste content for the seek of pragmatism when developing things on Carto4React and testing them on cloud-native 😄

packages/react-api/src/api/model.js
```diff
@@ -98,14 +105,24 @@ export function executeModel(props) {
         }
       : undefined;
 
-    queryParams.spatialFilters = JSON.stringify(spatialFilters);
-    queryParams.spatialDataType = spatialDataType;
+    if (spatialFilters) queryParams.spatialFilters = JSON.stringify(spatialFilters);
+    if (spatialDataType) queryParams.spatialDataType = spatialDataType;
+    if (spatialDataColumn) queryParams.spatialDataColumn = spatialDataColumn;
```
jmgaya, line 110: This only specifies which is the `spatialDataColumn` and shouldn't affect other Widgets, but it's used from the Backend perspective for determining which column they should "transform", for more info see: https://github.com/CartoDB/cloud-native/pull/17146

packages/react-widgets/src/widgets/TableWidget.js
```diff
@@ -121,6 +122,7 @@ function TableWidget({
             height={height}
             dense={dense}
             isLoading={isLoading}
+            onRowClick={onRowClick}
```
jmgaya, line 125: I've added support for the `onRowClick` event Handler to the TableWidget, because it was support but unused, and it's very useful for testing this on the `cloud-native` side 👍

packages/react-api/src/api/model.js
```diff
@@ -98,14 +105,24 @@ export function executeModel(props) {
...
         queryParams.spatialFiltersResolution = source.spatialFiltersResolution;
       }
       queryParams.spatialFiltersMode = source.spatialFiltersMode || 'intersects';
     }
+
+    // We should retrieve the spatialDataColumn as part of the params.column
+    if (props.model === 'table') {
+      queryParams.params = JSON.stringify({
+        ...params,
+        column: [...params.column, spatialDataColumn]
```
zbigg, line 123: Very minor.
I would prefer special parameter that either hides (`hiddenColumns`) some column or fetches extra column (`invisiableColumns`) than magic that just adds geo column to fetched set.

App will have power to use whatever "extra" column it needs - even if current use case is to add exactly same column as `spatialDataColumn`

jmgaya: I like it 🤔 

The only "problem" I see is that we'll have to expose or move somewhere the extraction of the `spatialDataColumn`, because it's the one we need to pass down as "hiddenColumns` or whatever name we decide to use, WDYT?

packages/react-widgets/src/widgets/TableWidget.js
padawannn: ok but add a new comment in the parm section please

packages/react-api/src/api/model.js
padawannn: Agree with @donmccurdy but it's something that we need to check in the server-side click initiative. Also, I'm worried if PS uses this widget and suddenly they will start receiving geometries. @jmgaya It would be possible to request the geometries as an optional parameter of the table widget?

padawannn: I think this `if` is going to be difficult to adapt when we integrate `carto-api-client`. The TableWidget has a `columns` parameter that you could use to add the `spatialDataColumn` and you would need an extra property to hide this column. I know that it requires more changes but I think that would be cleaner.

padawannn (changes requested):

packages/react-api/src/api/model.js
jmgaya: > When a unique ID is available, perhaps fetching the entire geometry into the table widget is not necessary?

Yup, I think this makes total sense! However, it's not ready yet, but we should probably re-visit this once we merge this 👍

jmgaya: > I think this if is going to be difficult to adapt when we integrate carto-api-client

Absolutely agree! I've tried to keep this as minimal as possible, but in terms of reusability, we can handle this in a different way, let me propose something else 👍

padawannn (approved):

PR #902, opened 2024-09-12
Add support for `stringSearch` to Table Widget

jmgaya:
# Description

Shortcut: (434955)

Screenshot 2024-09-12 at 15 21 37

## Type of change

- Feature


packages/react-widgets/src/widgets/TableWidget.js
```diff
@@ -46,6 +49,9 @@ function TableWidget({
   height,
   stableHeight,
   dense,
+  searchText,
+  searchColumn,
+  search,
```
zbigg, line 54: Is it needed? It seems unused.

zbigg (approved):
lgtm, one minor comment

packages/react-widgets/src/widgets/TableWidget.js
jmgaya: Unused, and removed 🤦

PR #903, opened 2024-09-19
Support local Tileset Search in react-workers

jmgaya:
# Description

Shortcut: (441123)

https://github.com/user-attachments/assets/13deb351-4c70-4693-83aa-3483783988cf

## Type of change

- Feature

### DIFF
--- a/packages/react-workers/src/workers/methods.js
+++ b/packages/react-workers/src/workers/methods.js
@@ -239,26 +239,23 @@ export function getRawFeatures({
   let hasData = false;
 
   if (currentFeatures) {
-    const filteredFeatures = applySorting(
-      getFilteredFeatures(filters, filtersLogicalOperator),
-      {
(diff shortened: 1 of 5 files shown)


packages/react-workers/src/workers/methods.js
```diff
@@ -236,11 +239,26 @@ export function getRawFeatures({
...
+    const filteredFeatures = applySorting(
+      getFilteredFeatures(filters, filtersLogicalOperator),
+      {
+        sortBy,
+        sortByDirection,
+        sortByColumnType
+      }
+    );
+
+    if (searchFilterColumn && searchFilterText) {
```
zbigg, line 251: Minor. I would apply sorting after searching.
Searching is kind-a filtering. Is it?

zbigg (commented):
one minor, otherwise LGTM

PR #910, opened 2024-09-23
Remote spatial index widgets #2

zbigg:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/425306

* [x] Spatial index sources use remote widgets calculation.
* [x] Calculate `spatialFilterResolution` for remote spatial index widgets 

## Type of change

- Feature

# Acceptance

N/A 

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-widgets/src/models/spatialFiltersResolution.js
+++ b/packages/react-widgets/src/models/spatialFiltersResolution.js
@@ -48,7 +48,7 @@ export function getSpatialFiltersResolution({ source, spatialDataType, viewState
   const dataResolution = source.dataResolution ?? Number.MAX_VALUE;
 
   const aggregationResLevel =
-    source.metadata?.aggregationResLevel ??
+    source.aggregationResLevel ??
     (spatialDataType === 'h3'
       ? DEFAULT_AGGREGATION_RES_LEVEL_H3
       : DEFAULT_AGGREGATION_RES_LEVEL_QUADBIN);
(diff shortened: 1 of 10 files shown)


packages/react-widgets/src/models/spatialFiltersResolution.js
```diff
@@ -0,0 +1,87 @@
...
+  [3, 1],
+  [2, 1],
+  [1, 0]
+];
+
+const quadBinZoomMaxOffset = 4;
+
+const DEFAULT_TILE_SIZE = 512;
+const DEFAULT_AGGREGATION_RES_LEVEL_H3 = 4;
+const DEFAULT_AGGREGATION_RES_LEVEL_QUADBIN = 6;
```
donmccurdy, line 41: I am a little nervous about implementation skew – these constants are changing with a feature flag in Builder, `getHexagonResolution` could change in `@deck.gl/carto`... but given the plans I guess it is temporary!

donmccurdy (approved):

packages/react-widgets/src/models/spatialFiltersResolution.js
zbigg: I was against exposing this functionality so early, but it was decided we want it to be available to PS, so that's why we're here.
I'm not satisfied with this too.

PR #911, opened 2024-09-23
tileFeatures returns features with geometry if asked

padawannn:
# Description

Added a new option to `tileFeatures` method to return the features with their geometry

## Type of change
- Feature


packages/react-core/src/utils/transformTileCoordsToWGS84.js
```diff
@@ -0,0 +1,66 @@
+import { lerp } from '@math.gl/core';
+import { lngLatToWorld, worldToLngLat } from '@math.gl/web-mercator';
+
+/**
+ * Transform WGS84 coordinates to tile coords.
+ * It's the inverse of deck.gl coordinate-transform (https://github.com/visgl/deck.gl/blob/master/modules/geo-layers/src/mvt-layer/coordinate-transform.js)
```
jmgaya, line 6: This link leads to a 404, or am I doing something wrong?

packages/react-core/src/filters/tileFeaturesGeometries.js
```diff
@@ -4,36 +4,84 @@ import booleanWithin from '@turf/boolean-within';
...
+      geometry = getFeatureByType(ringCoordinates, type);
+    }
+
+    // If intersection is required, check before proceeding
+    if (!geometryIntersection || intersects(geometry, geometryIntersection)) {
+      const properties = parseProperties(tileProps);
+
+      // Only save geometry if necessary
+      if (storeGeometry) {
+        properties.geom =
```
jmgaya, line 46: What would happen if the source contains a column named `geom`, could we have collisions between them?

If the answer is yes, maybe we can name this property as `__internal_tileset_geometry` (or similar), and expose it for its usage in `cloud-native` when recovering such information, as we do for example with `TILE_FORMATS`, as this column name feels less "common" than this one?

jmgaya (commented):
Despite a small concern regarding the `geom` property name, LGTM 🚢 🪄

packages/react-core/src/filters/tileFeaturesGeometries.js
```diff
@@ -4,36 +4,84 @@ import booleanWithin from '@turf/boolean-within';
...
+  bbox,
+  tileFormat,
+  uniqueIdProperty,
+  storeGeometry,
+  geometryIntersection = null
+}) {
+  const tileProps = getPropertiesFromTile(data, startIndex);
+  const uniquePropertyValue = getUniquePropertyValue(tileProps, uniqueIdProperty, map);
+
+  if (uniquePropertyValue && !map.has(uniquePropertyValue)) {
```
donmccurdy, line 30: Minor - An early exit might be more readable than nesting `if` blocks, here and below:

```javascript
if (!uniquePropertyValue || map.has(uniquePropertyValue)) {
  return;
}

...

if (geometryIntersection && !intersects(geometry, geometryIntersection)) {
  return;
}
```

donmccurdy (approved):
Only minor comments, LGTM!

packages/react-core/src/utils/transformTileCoordsToWGS84.js
padawannn: Fixed

packages/react-core/src/filters/tileFeaturesGeometries.js
padawannn: Done

packages/react-core/src/filters/tileFeaturesGeometries.js
padawannn: Yes is more readable, thank you for the suggestion, I've changed it

packages/react-core/src/filters/tileFeaturesGeometries.js
```diff
@@ -4,36 +4,87 @@ import booleanWithin from '@turf/boolean-within';
 import intersect from '@turf/intersect';
 import transformToTileCoords from '../utils/transformToTileCoords';
 import { TILE_FORMATS } from '../types';
+import transformTileCoordsToWGS84 from '../utils/transformTileCoordsToWGS84';
+
+export const GEOM_STORED_VALUE = '__geomValue';
```
jmgaya, line 9: I'm a bit concerned about this `GEOM_STORED_VALUE` because it's a bit hard to determine what it represents 🤔 

Given that, I'll suggest we replace this with something more meaningful, like`TILESET_GEOMETRY_PROPERTY`, which makes it more obvious that it's a Geometry Property from a Tileset Data Source, WDYT?

padawannn: TILESET sounds like static tileset and this code could be used for dynamic tiling. I've changed to `FEATURE_GEOM_PROPERTY `

PR #913, opened 2024-09-25
getSqlEscapedSource: support for databricksRest provider

zbigg:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/442095

`getSqlEscapedSource` & `FullyQualifiedName` to support `databricksRest` provider type - which is same as `databricks` from SQL pov
## Type of change

- Fix

# Acceptance

Already covered with UT

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-widgets/src/models/fqn.js
+++ b/packages/react-widgets/src/models/fqn.js
@@ -9,6 +9,9 @@ const bqProjectId = /^[a-z][a-z0-9-]{4,28}[a-z0-9]$/;
 const bqIdentifierRegex = '((?:[^`.]*?)|`(?:(?:[^`.])*?)`)';
 const identifierRegex = '((?:[^".]*?)|"(?:(?:[^"]|"")*?)")';
 const databricksIdentifierRegex = '((?:[^`.]*?)|`(?:(?:[^`]|``)*?)`)';
+const databricksFqnParseRegex = new RegExp(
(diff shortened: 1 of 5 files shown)


packages/react-widgets/src/models/fqn.js
```diff
@@ -26,13 +26,15 @@ const fqnParseRegex = {
     `^${databricksIdentifierRegex}(?:\\.${databricksIdentifierRegex})?(?:\\.${databricksIdentifierRegex})?$`
   )
 };
+fqnParseRegex[Provider.DatabricksRest] = fqnParseRegex[Provider.Databricks];
```
jmgaya, line 29: _Suggestion:_ I'll add this for consistency inside of `fqnParseRegex` and will extract the Regex, rather than updating the object afterward, increasing (at least, IMHO) this part's readability, e.g.

```
const databricksRegex = `^${databricksIdentifierRegex}(?:\\.${databricksIdentifierRegex})?(?:\\.${databricksIdentifierRegex})?$`
const fqnParseRegex = {
  ...
  [Provider.Databricks]: databricksRegex,
  [Provider.DatabricksRest]: databricksRegex
}
```

Same applies to `nameNeedsQuotesChecker`

jmgaya (approved):
LGTM 🚀

packages/react-widgets/src/models/fqn.js
zbigg: Applied more or less.

PR #920, opened 2024-10-10
Handle BigInt numbers from API

jmgaya:
# Description

https://app.shortcut.com/cartoteam/story/445700/handle-bigint-numbers-coming-from-the-api-properly-in-carto-react?vc_group_by=day&ct_workflow=all&cf_workflow=500193639

## Type of change

- Fix

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,3 +1,6 @@
 {
-  "extends": "react-app"
+  "extends": "react-app",
+  "env": {
+    "es2020": true
+  }
 }
--- a/packages/react-api/__tests__/api/stats.test.js
+++ b/packages/react-api/__tests__/api/stats.test.js
@@ -86,7 +86,7 @@ describe('stats', () => {
       const fetchMock = (global.fetch = jest.fn().mockImplementation(async () => {
         return {
           ok: true,
-          json: async () => TABLE_TEST.output
+          text: async () => JSON.stringify(TABLE_TEST.output)
         };
       }));
 
@@ -117,7 +117,7 @@ describe('stats', () => {
       const fetchMock = (global.fetch = jest.fn().mockImplementation(async () => {
         return {
           ok: true,
-          json: async () => QUERY_TEST.output
+          text: async () => JSON.stringify(QUERY_TEST.output)
         };
       }));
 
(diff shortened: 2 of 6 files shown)


.eslintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "react-app"
+  "extends": "react-app",
+  "env": {
+    "es2020": true
```
jmgaya, line 4: Otherwise, `BigInt` constructor gives an ESLint error 🤦

packages/react-api/__tests__/api/stats.test.js
```diff
@@ -225,5 +225,20 @@ describe('stats', () => {
...
+      const fetchMock = (global.fetch = jest.fn().mockImplementation(async () => {
+        return {
+          ok: true,
+          text: async () => jsonWithBigInt
+        };
+      }));
+
+      const res = await getStats({ source: QUERY_SOURCE, column: 'injuries' });
+
+      expect(res.rows[0].quadbin).toEqual(BigInt(5256683984082960383));
```
donmccurdy, line 241: Careful here! Writing this out as `BigInt(5256683984082960383)` will truncate the number before it's passed to the BigInt function, so these are not equal, note the `n` suffix:

```javascript
BigInt(5256683984082960383) === BigInt(5256683984082960383n) // false!
```

packages/react-api/src/api/common.js
```diff
@@ -1,5 +1,13 @@
 import { InvalidColumnError } from '@carto/react-core/';
 
+// Widgets API may return BigInt values, and JSON.parse does not handle them properly, so we need to use a reviver
+const bigIntReviver = (key, value) => {
+  if (typeof value === 'number' && value > Number.MAX_SAFE_INTEGER) {
+    return BigInt(value);
+  }
+  return value;
+};
```
donmccurdy, line 9: See above — unfortunately I think that it's necessary to receive the value as a string and not an integer. By the time the reviver receives a numeric value, I think it will be too late to recover the original integer.

Possible solution in https://stackoverflow.com/a/69644630, though maybe this is complex enough that we want a dependency to help...

donmccurdy (changes requested):

PR #923, opened 2025-02-05
Add Virtualized List to Category Widget

jmgaya:
# Description

https://app.shortcut.com/cartoteam/story/460862/qa-for-support-virtual-scroll-in-category-widgetqa-for

https://github.com/user-attachments/assets/aab2daee-3c5b-4293-bcc3-7ab2580fce80

## Type of change

- Fix

### DIFF
--- a/packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
+++ b/packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
@@ -432,7 +432,7 @@ function CategoryWidgetUI(props) {
     );
   };
 
-  const Row = useCallback(
+  const CategoryRow = useCallback(
     ({ index, style }) => {
       const d = animValues[index];
       return (
@@ -542,7 +542,7 @@ function CategoryWidgetUI(props) {
             itemCount={animValues.length}
             itemSize={48}
           >
-            {Row}
+            {CategoryRow}
           </FixedSizeList>
         ) : (
           <Box>
--- a/packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
+++ b/packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
@@ -1,4 +1,5 @@
 import React, { useCallback, useEffect, useRef, useState } from 'react';
+import { FixedSizeList } from 'react-window';
 import PropTypes from 'prop-types';
 import {
   Button,
@@ -32,6 +33,7 @@ import SearchIcon from '../../assets/icons/SearchIcon';
 import useImperativeIntl from '../../hooks/useImperativeIntl';
 import { ORDER_TYPES } from '../utils/chartConstants';
 import useSkeleton from '../useSkeleton';
+import styled from '@emotion/styled';
 
 function usePrevious(value) {
   const ref = useRef();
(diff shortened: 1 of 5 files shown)


packages/react-widgets/package.json
```diff
@@ -66,7 +66,8 @@
     "webpack-cli": "^4.5.0"
   },
   "dependencies": {
-    "@babel/runtime": "^7.13.9"
+    "@babel/runtime": "^7.13.9",
```
jmgaya, line 69: As we pretend to deprecate this package, I prefer not to dig into this too much, but I think this dependency shouldn't be here 🤦

packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
```diff
@@ -517,15 +536,14 @@ function CategoryWidgetUI(props) {
...
-            />
-          ))
+          <FixedSizeList
+            height={320}
+            width='100%'
+            itemCount={animValues.length}
+            itemSize={48}
+          >
+            {Row}
+          </FixedSizeList>
```
jmgaya, line 546: This is what is being recommended by MUI: https://mui.com/material-ui/react-list/#virtualized-list

packages/react-ui/storybook/stories/widgetsUI/CategoryWidgetUI.stories.js
```diff
@@ -60,7 +60,7 @@ const LoadingTemplate = (args) => {
   );
 };
 
-const data = [...Array(30)].map((_, idx) => ({
+const data = [...Array(2000)].map((_, idx) => ({
```
jmgaya, line 63: I've tested with up to 10K elements, and it works, but starting at that point, memory gets drained 👎 

I did some tests with `react-window`, and it supports millions of rows, so you probably know where the problem is at this point 😄

vmilan (approved):
LGTM, but just wondering if we should make this same change also in Meridian, for feature parity, since we'll be using UI widgets from there at some point ¿?

packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
```diff
@@ -517,15 +536,14 @@ function CategoryWidgetUI(props) {
...
-            <CategoryItem
-              key={i}
-              data={d}
-              onCategoryClick={() =>
-                showAll ? handleCategoryBlocked(d.name) : handleCategorySelected(d.name)
-              }
-            />
-          ))
+          <FixedSizeList
+            height={320}
```
padawannn, line 540: Use theme.spacing rather than 320 and 48

padawannn (changes requested):

packages/react-ui/storybook/stories/widgetsUI/CategoryWidgetUI.stories.js
```diff
@@ -42,63 +19,15 @@ const Template = (args) => {
...
   name: `Category ${idx + 1}`,
   value: idx * 100
 }));
 
-const filteredData = data.slice(0, 4).map((cat, idx) => {
-  return { ...cat, value: idx % 2 === 0 ? null : cat.value };
-});
-
 export const Default = Template.bind({});
-const DefaultProps = { data };
```
vmilan, line 73: This simplification is against use cases visibility... Basically you're removing most of the stories to display this new case.... Please, let's think in a different way.

vmilan (changes requested):

packages/react-ui/storybook/stories/widgetsUI/CategoryWidgetUI.stories.js
jmgaya: Sure! Let me rollback this 👍

vmilan (approved):
LGTM, thanks!

PR #925, opened 2025-02-06
Fix issue while using useCallback

jmgaya:
## Type of change

- Fix

### DIFF
--- a/packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
+++ b/packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
@@ -432,22 +432,19 @@ function CategoryWidgetUI(props) {
     );
   };
 
-  const CategoryRow = useCallback(
-    ({ index, style }) => {
-      const d = animValues[index];
-      return (
-        <div style={style}>
-          <CategoryItem
-            data={d}
-            onCategoryClick={() =>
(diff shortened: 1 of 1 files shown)


packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
```diff
@@ -432,22 +432,19 @@ function CategoryWidgetUI(props) {
     );
   };
 
-  const CategoryRow = useCallback(
```
vmilan, line 435: Couldn't removing `useCallback` cause unnecessary re-renders in the virtualized list?

jmgaya: You're just rendering up to 9 elements, so it's not going to cause huge performance issues 👍

vmilan: ok, thanks! please consider adding someone from Builder to the review, as they probably have more context about this issue.

vmilan (approved):
UI LGTM

packages/react-ui/src/widgets/CategoryWidgetUI/CategoryWidgetUI.js
jmgaya: Sure! 👍

padawannn (approved):

PR #929, opened 2025-03-13
refactor(ui,widgets): Remove .defaultProps for React 18 compatibility [sc-469139]

donmccurdy:
# Description

Shortcut: https://app.shortcut.com/cartoteam/story/469139/bump-workspace-www-to-react-18

Removes use of `.defaultProps` throughout the codebase, for compatibility with React 18. Other than the obvious changes, this also requires using the pattern below ...

```diff
function MyComponent({
-  data = []
+  data = EMPTY_ARRAY
}) {
  ...
}
```

... because (unlike .defaultProps), JavaScript's default parameters are re-evaluated on each function invocation, which make change behavior (break caches, create infinite loops) in React hooks.

## Type of change

- Refactor

# Acceptance

n/a

# Basic checklist

- Good PR name
- Shortcut link
- Changelog entry
- Just one issue per PR
- GitHub labels
- Proper status & reviewers
- Tests
- Documentation

### DIFF
--- a/packages/react-ui/src/widgets/FeatureSelectionWidgetUI/FeatureSelectionWidgetUI.js
+++ b/packages/react-ui/src/widgets/FeatureSelectionWidgetUI/FeatureSelectionWidgetUI.js
@@ -15,6 +15,8 @@ const StylesWrapper = styled(Paper)(({ theme }) => ({
   padding: theme.spacing(0.5)
 }));
 
+const EMPTY_ARRAY = [];
+
 /**
  * Renders a <FeatureSelectionWidgetUI /> component
  *
@@ -44,7 +46,7 @@ const StylesWrapper = styled(Paper)(({ theme }) => ({
 function FeatureSelectionWidgetUI(props) {
   const {
     selectionModes,
-    editModes = [],
+    editModes = EMPTY_ARRAY,
     selectedMode,
     onSelectMode,
     enabled = false,
--- a/packages/react-ui/src/widgets/FeatureSelectionWidgetUI/FeatureSelectionWidgetUI.js
+++ b/packages/react-ui/src/widgets/FeatureSelectionWidgetUI/FeatureSelectionWidgetUI.js
@@ -47,7 +47,7 @@ function FeatureSelectionWidgetUI(props) {
     editModes = [],
     selectedMode,
     onSelectMode,
-    enabled,
+    enabled = false,
     onEnabledChange,
     geometry,
     onSelectGeometry,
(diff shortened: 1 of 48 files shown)


packages/react-ui/src/widgets/WrapperWidgetUI.js
```diff
@@ -131,27 +131,39 @@ const PaperMenu = styled(Menu)(({ theme }) => ({
...
+  footer
+}) {
   const wrapper = createRef();
 
   const [expandedInt, setExpandedInt] = useState(true);
   const externalExpanded =
-    typeof props.expanded === 'boolean' && typeof props.onExpandedChange === 'function';
+    typeof _expanded === 'boolean' && typeof onExpandedChange === 'function';
   const expanded =
-    props.expandable !== false ? (externalExpanded ? props.expanded : expandedInt) : true;
```
zbigg, line 141: nit: kinda "refactor" in the one that is really complex and fragile, without typescript and with chance for typo ...
for safety - i would leave props as they were

zbigg (approved):
good jo

packages/react-ui/src/widgets/FeatureSelectionWidgetUI/FeatureSelectionWidgetUI.js
```diff
@@ -47,7 +47,7 @@ function FeatureSelectionWidgetUI(props) {
     editModes = [],
```
jmgaya, line 47: Shouldn't this be `EMPTY_ARRAY` as per above changes?

packages/react-ui/src/widgets/comparative/ComparativeCategoryWidgetUI/CategoryItem.js
```diff
@@ -82,13 +82,13 @@ ComparativeCategoryTooltip.propTypes = {
 
 function CategoryItem({
   item,
-  animation,
-  animationOptions,
+  animation = true,
+  animationOptions = {},
```
jmgaya, line 86: Shouldn't this be `EMPTY_OBJECT`?

jmgaya (commented):
Apart from two minor issues that may affect per your great explanation, great job mate! 🚀

packages/react-ui/src/widgets/FeatureSelectionWidgetUI/FeatureSelectionWidgetUI.js
donmccurdy: Ah good catch, thank you!

